PR #73, opened 2013-12-04
`stbt.hold` to emulate holding down a button on the remote control

drothlis:
For background see the discussion on the mailing list:
https://groups.google.com/forum/#!topic/stb-tester/fvdWw8o_EjA

I'm not sure if this API for `stbt.hold` is a good idea, given that it
will be quite difficult to implement for the RedRat irNetBox unless
RedRat implement additional functionality. I suppose we could launch a
separate thread that sends lots of individual keypresses and hope that's
good enough (it will only work for infrared protocols where the initial
& repeat signals are the same). We might even have to do the same for
the LIRC implementation, given that LIRC's repeat sending doesn't work
inside a VM, at least not for me.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -309,6 +309,30 @@ press(key)
     `key` is a string. The allowed values depend on the control you're using:
     If that's lirc, then `key` is a key name from your lirc config file.
 
+hold(key)
+    Context manager that sends key-down/key-up to the system under test.
+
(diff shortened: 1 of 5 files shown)


drothlis:
Closing because we can't emulate a long-press infrared signal with LIRC.

PR #86, opened 2014-01-19
Rename stbt.UITestFailure => stbt.TestFailure; deprecate UITestError

drothlis:
The naming of these exceptions always bothered me. What does everyone
think of:

* Renaming `UITestFailure` to `TestFailure` (6cc0730)
* Deprecating `UITestError` (just use `RuntimeError` or whatever
  exception is appropriate) (d15cb36)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -317,16 +317,16 @@ The following functions are available:
 .. <start python docs>
 
 as_precondition(message)
-    Context manager that replaces TestFailures with TestErrors.
+    Context manager that replaces test failures with test errors.
 
     If you run your test scripts with stb-tester's batch runner, the reports it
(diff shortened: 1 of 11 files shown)


drothlis:
I'd like to merge this for the 0.22 release, or at least decide to abandon the idea.

drothlis:
Rebased onto master.

wmanley:
I agree that this should be merged, although I don't think it makes sense to rename `UITestFailure` if we're going to encourage people to use `AssertionError` instead.

drothlis:
This will affect people's classify script if they're grepping the logs for "UITestFailure".

drothlis:
Closing as this is essentially a cosmetic change and the amount of time spent on it already suggests that it isn't worth it.

PR #105, opened 2014-02-25
stbt.Display: added clock on each frame's left top corner

Drachenfels:
While recording a video with this commit there will be visible clock
on left top corner. Clock displays time when given frame was
processed. In addition when calling a function 'press' we will draw not
only an information that button was pressed but as well time when it was
pressed. This should provide accurate information that will let us find
relation between actions registered in logs and what happened on screen.

### DIFF
--- a/stbt/__init__.py
+++ b/stbt/__init__.py
@@ -1789,6 +1789,9 @@ class Display(object):
 
     def draw(self, obj, duration_secs):
         if type(obj) in (str, unicode):
+            obj = (
+                datetime.datetime.now().strftime("%H:%M:%S:%f")[:-4] +
+                ': ' + obj)
             self.text_annotations.append((obj, duration_secs, None))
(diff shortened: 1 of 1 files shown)


aiman:
Hi Dave. Now that stbt 0.21 is out, perhaps you can take a look at this again.

szemate:
Hi, sorry for pushing this again, but basically this patch is now the only reason why we're maintaining a fork of stb-tester. If you are worried that it might clash with https://github.com/wmanley/stb-tester/commit/e1b69a2bbf796d3c73e8258438af1250901dd7bf, couldn't it go behind a command-line switch?

drothlis:
Hi @szemate thanks for the reminder.

This needs rebasing onto master (it doesn't merge cleanly) and the test failures need to be fixed.

drothlis:
Closed as a duplicate of #255.

PR #107, opened 2014-02-27
extra/runner/report.py - use SQLite cache of file contents to speed up

asimihsan:
On file systems where both stat'ing files and getting their contents are
very slow (two or three orders of magnitude slower) this commit improves
performance by five-fold by using an SQLite cache and other changes.

This is a RFC. Although an old version of this code is currently successfully
running on a production system this pull request is of a recently rebased
but untested version.

I would like to break out the `FileCache` object into its own importable file.
This is because on a production system using stb-tester there are other
background processes that are interested in the contents of stb-tester runner
reports but are currently unable to use the existing caches to improve their
read performance. My plan was to break this out into a `FileCache` file and
add respective integration tests to the same file (i.e. tests that go beyond
simple mocking and use an actual directory of files, reads them into the
cache, deletes the actual files, checks the DB still has them, etc.). Thoughts?

I'm concerned that the way in which I've written `_get_path`, by returning
`None`, doesn't really make sense, however it did help me make the `read`
code more concise. Suggestions as to how to express the logic here are welcome.

### DIFF
--- a/stbt-batch.d/report.py
+++ b/stbt-batch.d/report.py
@@ -6,7 +6,9 @@
 
 """Generates reports from logs of stb-tester test runs created by 'run'."""
 
+import codecs
 import collections
+import contextlib
 from datetime import datetime
 import glob
 import itertools
@@ -14,6 +16,9 @@ import os
 from os.path import abspath, basename, dirname, isdir
 import re
 import sys
+import tempfile
+import shutil
+import sqlite3
 
 import jinja2
 escape = jinja2.Markup.escape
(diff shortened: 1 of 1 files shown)


drothlis:
Closing, because we want to take an alternate approach -- see <https://github.com/wmanley/stb-tester-service/pull/4>.

drothlis:
Closing, because we want to take an alternate approach -- see <https://github.com/wmanley/stb-tester-service/pull/4>.

PR #174, opened 2014-06-03
virtual-stb: Add virtual stb - useful for testing virtualized STBs

wmanley:
For instance to test an STB UI written for HbbTV or HTML5 you can use this with chrome.

This is not yet ready for merging or even detailed review.  Comments on approach would be appreciated though.

TODO:

- [x] Add readyness notification so it can be used as a part of a CI system without races.
- [x] Make it installable and give it a real `stbt virtual-stb` or similar command
- [ ] Move it out of extra/
- [ ] Improve documentation including using it in the context of CI or similar
- [ ] Write tests
- [x] Merge pull-request #173
- [ ] When using docker containers run [pulsevideo] inside the container and use that as the video source rather than attempting cross-container X11

[pulsevideo]: //github.com/wmanley/pulsevideo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,7 @@ tools += stbt-power
 tools += stbt-screenshot
 tools += stbt-templatematch
 tools += stbt-tv
+tools += stbt-virtual-stb
 
 # Generate version from 'git describe' when in git repository, and from
 # VERSION file included in the dist tarball otherwise.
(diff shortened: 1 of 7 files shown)


wmanley:
We're going to be opening a new pull-request with a more incremental approach (minus docker support) to this soon so I'm closing this PR.

PR #175, opened 2014-06-03
Add Dockerfile

wmanley:
[Docker] is a implementation of Linux containers which offer many of the
same benefits as virtual machines, but are much more lightweight.  They
allow you to package up a program and it's dependencies such that it will
run exactly the same on any machine.  They can also provide isolatoion such
that the host machine is isolated from changes made in the container.

This is potentially very useful to stb-tester.  stb-tester can be difficult
to deploy due to the dependencies.  Docker offers a way to easily provide a
standardised stb-tester environment.  It is also useful in a CI context as,
much like Travis, you could run arbritary user code without risking the
host system.

A Dockerfile contains the instructions on how to create a container.
Containers start out with a base image and build on top of that.  In this
case the stb-tester container is based ubuntu 14.04 (LTS).  Any container
can be used as the base for other containers.  Our users can base their
containers on the stb-tester container if they wish.

Docker also provides an [index].  This is a global repository of docker
images.  Docker containers in the index are easier to deploy[^1] and use as
a base for other images.  It also provides a service called "[trusted
builds]" where your image will be built and published for you, from github,
with a guarantee that the sources there are the source of the container.
One constraint is that the Dockerfile must be in the root of the repo and
present at clone time.  This means that we can't generate it as a part of
the build process.

[^1]: You can just type `docker run stb-tester:latest` and you're up and
      running.
[Docker]: http://docker.io/
[index]: https://index.docker.io/
[trusted builds]: http://blog.docker.com/2013/11/introducing-trusted-builds/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,30 @@
+FROM ubuntu:14.04
+MAINTAINER will@stb-tester.com
+
+ENV DEBIAN_FRONTEND noninteractive
+
+ADD extra/debian/control /tmp/runtime-deps
+RUN apt-get update && \
+    apt-get dist-upgrade -y && \
+    apt-get install -y $(awk '/^         [^$ ]/ {print $1}' /tmp/runtime-deps | sed 's/,//g') && \
+    apt-get clean && \
+    rm /tmp/runtime-deps
+
+ADD . /tmp/source
+RUN apt-get install -y git make python-docutils && \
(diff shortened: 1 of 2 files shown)


wmanley:
Decided that this was a bad idea.  If we want to provide a standard customisable environment for running tests we don't really want to be tied to docker.  Also from a deployment perspective it makes more sense for the Dockerfile to be in stb-tester-service instead as exactly what you put in the Dockerfile depends on your hardware setup among other things.  It's too much an ops decision to include it here.

Closing

PR #176, opened 2014-06-04
Visualize sleep

LewisHaley:
Just an idea I had that I thought might be useful.

Also there's some pylint disables but I can't finish running `make check` because
I'm not using gstreamer1 yet.


drothlis:
Closing in favour of the stb-tester-service web interface, which visualises the line being run.

Of course there's nothing stopping anybody from doing something like this in their own library functions -- that's what `stbt.draw_text` is there for. But I don't think this `sleep` function belongs in core stbt.

PR #191, opened 2014-06-26
stbt batch: New hooks for custom test report and summary generators

szemate:
`report.py` has certain limitations: generating static pages and not
using a database back-end makes it somewhat unsuitable for reporting
large volumes of test results; it doesn't allow sending stb-tester
results to an external reporting system.

To allow `stbt batch` users to overcome the above issues, add two new
user hooks, `batch.report` and `batch.summary` that may replace some or
all of `report.py` functionality with user programs.

The idea to replace `report.py` is the result of the following
considerations:
- Forking and building stb-tester with a customised `stbt batch` would
  involve the maintenance burden of continuously backporting changes
  from upstream. We'd prefer to use official releases.
- Implementing a custom test runner instead of using `stbt batch` is
  certainly viable, but replacing just the report generator saves the
  hassle of reimplementing a very similar test runner.
- Replacing `stbt batch report` would be ambiguous, because this is the
  script that and invokes the `classify` hook. Also, it generates
  `failure-reason`, which is actually very useful.

`report.py` has two different functionalities:
- `report.py <testrun-dir>` generates a test report of a single test
  run,
- `report.py index.html` generates a summary report of all test runs.
The former may be replaced with `batch.report` and the latter with
`batch.summary` user programs, using single-word hook names for
simplicity. Specifying `batch.report` or `batch.summary` with no
value skips test report or summary generation respectively.

A possible use case is generating dynamic report pages. The `report`
hook fills a database using the contents of files provided by `stbt
batch run`. The `summary` hook is skipped, or it may e.g. send e-mail
notifications if the ratio of failed tests exceeds a threshold.

See the additions to README.rst for further details.

### DIFF
--- a/stbt-batch.d/README.rst
+++ b/stbt-batch.d/README.rst
@@ -41,10 +41,10 @@ User hooks
 ----------
 
 The following hooks are available for custom log collection, failure
-classification, and failure recovery. Each hook is a variable in the stbt
-configuration file; if the variable is set to an executable program, ``run``
-will invoke that program at the appropriate time, with the current working
-directory set to the directory containing the test run logs:
+classification, failure recovery, and custom report generation. Each hook is a
+variable in the stbt configuration file; if the variable is set to an executable
+program, ``run`` will invoke that program at the appropriate time, with the
+current working directory set to the directory containing the test run logs:
 
 **batch.pre_run**
   Invoked immediately before the test is run, with the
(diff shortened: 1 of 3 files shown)


szemate:
> I have discussed is to have the built-in and the user's data-collection scripts (that is, stbt batch run and the post_run hook, respectively) write data in json format; store this in a Mongo DB (or any DB that allows us to store arbitrary json and run queries based on arbitrary fields within that json); and generate the report from the database.

This comment implies that the pull request is out of scope -- is it OK if I close it?

drothlis:
@szemate yes please.

drothlis:
@szemate sure. Closing.

PR #195, opened 2014-07-02
Classify "really uninteresting" failures differently in the batch report

LewisHaley:
By uninteresting failures, I mean the _really_ uninteresting things like sigterms, sigsegvs, etc
which have nothing to do with the system-under-test and therefore they shouldn't be coloured
and counted as part of the general "failures" as it skews the results.

### DIFF
--- a/stbt-batch.d/report.py
+++ b/stbt-batch.d/report.py
@@ -118,8 +118,10 @@ class Run(object):
             return "success"
         elif self.exit_status == 1:
             return "error"  # Red: Possible system-under-test failure
-        else:
+        elif self.exit_status == 2:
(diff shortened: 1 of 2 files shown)


wmanley:
I'm closing this pull request because I don't think it's going anywhere.  I think maybe the reason that this is a priority to you is that you've been seeing SIGSEGVs with your capture device, and it would be less of an issue if that were fixed.

I agree that we need better ways of classifying and displaying results but I don't think this is the best place to discuss it.

I do like the idea of the graph being coloured green.  It could certainly make some thing clearer.  Could you raise a separate pull-request for that with a screenshot?

PR #212, opened 2014-07-31
Add stbt.match_all returning a list of matches

wmanley:
Search for `image` in the source video stream returning a list of matches.

TODO:

* [x] Rebase on master since #285 
* [x] Test with real-world screenshots
* [x] Fix returning overlapping results (test_that_results_dont_overlap)
* [x] Reduce duplication
* [x] Rebase on master again (April 2016).
* [x] Update HTML logging to show the multiple passes
    * [x] Merge #352 & rebase.
* [x] Squash & tidy
* [x] Merge #354 & rebase.
* [x] Make `MatchResult.frame` read-only, don't copy it.
* Tests:
    * [x] Test regions
    * Test HTML logging -- at least that it doesn't raise any exceptions when enabled.
        * [x] When generator stops because first pass doesn't match
        * [x] When generator stops because second pass doesn't match
        * [x] With different confirm_methods
    * [x] Test (manually) what happens with the drawn regions for multiple matches. Answer: Nothing is drawn if I do `stbt.match_all("template.png")` because `_match_all` doesn't hold `Display.lock` so the frame is pushed out immediately, before `_match_all` calls `draw`. Even if I use `for frame, _ in stbt.frames(): list(stbt.match_all(..., frame=frame))` nothing is drawn because `_match_all` thinks that the frame wasn't captured from live.
    * [x] Test what happens if there's a match at every position.
* [x] More documentation
* [x] Re-implement caching
* [x] Update release notes


wmanley:
Rebased and added a test showing the outstanding issue.

drothlis:
Rebased onto master.

drothlis:
Found the cause of the infinite loop bug that returns matches at the same position: The current implementation is blanking out the previous positions in `roi_mask` (used to communicate regions of interest from one level of the pyramid algorithm to the next) instead of operating on the final heatmap from the last pyramid level. The current implementation is calling `cv2.matchTemplate` again for each new match! We pass `roi_mask` through `cv2.boundingRect`, so if there are other viable match positions surrounding the one we blanked out, it would get included again in the bounding rectangle.

drothlis:
Rebased onto master to get #357 (Fix virtual-stb intermittant test-failure on Travis).

drothlis:
@wmanley I still want to add some more tests but in the meantime this is ready for review.

_stbt/core.py
```diff
@@ -1847,6 +1921,10 @@ def _match_template(image, template, match_parameters, roi_mask, level, imglog):
     imglog.add("template", template, level)
     imglog.add("source_matchtemplate", matches_heatmap, level)
 
+    return matches_heatmap
+
+
+def _find_best_match_position(matches_heatmap, method, threshold, level):
```
wmanley: This is quite a bit clearer now that `_match_template` and `_find_best_match_position` are split up.

_stbt/core.py
```diff
@@ -1775,31 +1816,64 @@ def _find_match(image, template, match_parameters, imglog):
...
-            image_pyramid[level], template_pyramid[level], match_parameters,
+        heatmap = _match_template(
+            image_pyramid[level], template_pyramid[level], method,
             roi_mask, level, imglog)
 
+        # Relax the threshold slightly for scaled-down pyramid levels to
+        # compensate for scaling artifacts.
+        threshold = max(
+            0,
+            match_parameters.match_threshold - (0.2 if level > 0 else 0))
```
wmanley: This is new, right?  This was causing issues with `match_all` that we weren't seeing with `match`?

drothlis: It isn't new -- the old one is [here](https://github.com/stb-tester/stb-tester/pull/212/files#diff-b8db1b2ea3c8e5386ca27ac571ea7aa6L1788).

The comment explaining it is new.

_stbt/core.py
```diff
@@ -1775,31 +1816,64 @@ def _find_match(image, template, match_parameters, imglog):
...
+                      else 0)
+        cv2.rectangle(
+            heatmap,
+            # -1 because cv2.rectangle considers the bottom-right point to be
+            # *inside* the rectangle.
+            (region.x, region.y), (region.right - 1, region.bottom - 1),
+            mask_value, cv2.cv.CV_FILLED)
+
+        matched, best_match_position, certainty = _find_best_match_position(
+            heatmap, method, threshold, level)
```
wmanley: Overall this seems fairly straight-forward.  My understanding is:

1. We split up the heat map generation from the finding the top location so we can do the former once and the latter many times
2. Then `_find_match -> _find_candidate_matches` that implements the first-stage matching becomes an iterator. This is the most complicated bit because of the Pyramid meaning that we might have to return/yield early.
3. `_match -> _find_matches` then iterates and checks each location from `_find_match` against the second pass, after checking that the first pass passed
4. `match` just returns the first result from `_find_matches`
5. `match_all` generates all but the last one

_stbt/core.py
```diff
@@ -2005,9 +2002,7 @@ def _confirm_match(image, position, template, match_parameters, imglog):
         return True
 
     # Set Region Of Interest to the "best match" location
-    roi = image[
-        position.y:(position.y + template.shape[0]),
-        position.x:(position.x + template.shape[1])]
+    roi = image[region.y:region.bottom, region.x:region.right]
```
wmanley: This could be replaced with `_crop`?.  Probably not worth it though, it would only be slower.

wmanley:
I've reviewed up to "_stbt.core: Refactor: Use Region instead of Position".  So far so good.  I'll hold off on "_stbt: Implement debug image logging for match_all" until you've rebased on master (with the caching changes).  I need to build up some courage first ;).

I've got some ideas about how `ImageLogger` could be a bit easier and nicer to use but as it's not user facing I don't think it's important.

_stbt/logging.py
```diff
@@ -80,43 +81,59 @@ class ImageLogger(object):
 
     Create a new ImageLogger instance for each frame of video.
     """
-    _frame_number = 1
+    _frame_number = itertools.count(1)
+
+    def __init__(self, name, **kwargs):
+        self.outdir = None
+        if get_debug_level() <= 1:
```
wmanley: #354 introduced something like this too.

drothlis:
I've extracted the original refactoring + the later ImageLogger refactorings into #361, which is rebased on top of master. Next I'll proceed with rebasing the `match_all` implementation on top of #361.

drothlis:
Rebased onto #361.

_stbt/core.py
```diff
@@ -647,29 +671,27 @@ def match(self, image, frame=None, match_parameters=None,
...
+                for (matched, match_region, first_pass_matched,
+                     first_pass_certainty) in _find_matches(
+                        _crop(npframe, region), template.image,
+                        match_parameters, imglog):
+
+                    match_region = Region.from_extents(*match_region) \
+                                         .translate(region.x, region.y)
+                    result = MatchResult(
+                        get_frame_timestamp(frame), matched, match_region,
+                        first_pass_certainty, numpy.copy(npframe),
```
wmanley: We're taking a copy of the frame every time we go round this loop.  Perhaps it would be better to take a single copy at the beginning of the function, mark it read-only and return it for each `MatchResult`.  This would be a (minor) API break for `match()` but I think it could be worth it, particularly given some changes I want to make to how we map the `Gst.Samples` so we don't need to make copies before handing them to users.

_stbt/core.py
```diff
@@ -647,29 +671,27 @@ def match(self, image, frame=None, match_parameters=None,
...
+                        (template.name or template.image))
+                    if grabbed_from_live:
+                        self._display.draw(result, None)
+                    yield result
+
+            except GeneratorExit:
+                imglog.set(first_pass_matched=first_pass_matched, result=result)
+                _log_match_image_debug(imglog)
+                raise
```
wmanley: This except block won't be run if we exhaust `_find_matches`.

wmanley: Ah, but we get away with it because neither `match_all` nor `match` ever exhaust `_match_all`.

stbt/__init__.py
```diff
@@ -140,6 +141,31 @@ def match(image, frame=None, match_parameters=None, region=Region.ALL):
...
+
+    Examples:
+
+    .. code-block:: python
+
+        all_buttons = list(stbt.match_all("button.png"))
+
+        for match_result in stbt.match_all("button.png"):
+            # do something with match_result here
+            ...
```
wmanley: Perhaps this should be two seperate code-blocks to make it clear that these lines of code are unrelated.

_stbt/core.py
```diff
@@ -1770,11 +1772,15 @@ def _find_matches(image, template, match_parameters, imglog):
...
-            _find_candidate_matches(image, template, match_parameters, imglog):
+    for i, first_pass_matched, region, first_pass_certainty in \
+            _find_candidate_matches(image, template, match_parameters,
+                                    imglog):
         confirmed = (
             first_pass_matched and
-            _confirm_match(image, region, template, match_parameters, imglog))
+            _confirm_match(image, region, template, match_parameters,
+                           imwrite=lambda name, img: imglog.imwrite(
+                               "match%d-%s" % (i, name), img)))  # pylint:disable=cell-var-from-loop
```
wmanley: This would be a bit nicer with the logging proposal I made in #361.  Not important, just a data point.

tests/test_match.py
```diff
@@ -117,6 +119,59 @@ def test_that_match_all_obeys_region():
         (177, 75), (177, 119)]]
 
 
+# These ImageLogger tests don't verify the html & png debug output, just that
+# it doesn't raise any exceptions.
```
wmanley: Hmm.  Will putting a `try: except:` block around `_log_match_image_debug` (like we discussed) remove the value from these tests?  Maybe they should at-least check that `index.html` has been written.  You could even use a regex like `]*src=['"]([^'"]*)['"] [^>]*>` on index.html to check that all the images you've written are referenced and that all the referenced images were written.

wmanley:
Review complete for the time being.  We can pair on the caching stuff if you'd like.  It shouldn't take too long.

_stbt/core.py
drothlis: Too subtle, I agree. I've force-pushed a change to use `finally` instead.

Originally I had used `except GeneratorExit` to avoid running the `imglog.set` in that block when `first_pass_matched` and `result` hadn't been defined if `_find_matches` raised an exception. Instead I've moved the `imglog.set` before the `yield` (the next commit did this, anyway, to log _all_ results from the loop).

drothlis:
`test-stbt-auto-selftest.sh:test_auto_selftest_caching` failed once for me ("caching isn't fast") when I was running `make check` on my computer. Running it on its own repeatedly (with the command below) passed 120 times.

    i=0; while tests/run-tests.sh test_auto_selftest_caching; do echo $i; i=$((i+1)); done

I'll ignore it for now.

_stbt/core.py
drothlis: > We're taking a copy of the frame every time we go round this loop.

Fixed in 923cd77.

stbt/__init__.py
drothlis: > Perhaps this should be two seperate code-blocks to make it clear that these lines of code are unrelated.

Done.

drothlis:
Everything on the todo list is done. @wmanley please review the commits after "tests: Ignore OCR mis-read on Travis".

docs/release-notes.md
```diff
@@ -23,6 +23,10 @@ UNRELEASED
 
 ##### Breaking changes since 24
 
+* The `frame` parameter of the `MatchResult` returned from `stbt.match`,
+  `stbt.detect_match` and `stbt.wait_for_match` is now read-only. Use
+  `numpy.copy(match_result.frame)` if you need a writable copy.
```
wmanley: `match_result.frame.copy()` would avoid the need to import numpy.

wmanley:
+1

docs/release-notes.md
drothlis: I used `numpy.copy` because it has a different default (order='K' which means keep the same order) than `ndarray.copy` (order='C' which means row-major order). But order='C' is the default for new arrays, and all our frames are in that order anyway, so I've changed the release note to suggest `frame.copy()`.

drothlis:
Oops I clicked the merge button before I pushed my last 2 documentation changes. I've pushed them to master: 9dbe502 and dd083a1.

Issue #211, opened 2014-07-30
VidiU: Tests fail with GStreamer error when the set-top-box under test is rebooted and --save-video is in use


szemate:
**Description of the test**
- The set-top-box gets power-cycled.
- The test periodically attempts to enter the main menu. The test passes if it's possible to enter the main menu within 5 minutes from turning the power on.
- During boot-up the set top box displays two splash screens before getting back to live TV.
- The two splash screens have different resolutions, therefore the VidiU encoder restarts twice.

**Issues**

(1.) Running the test with `--save-video` and the following source pipeline:

```
source_pipeline = rtmpsrc location=rtmp://testaut-2-pc-1/live/testaut-2-stb-1\ live=1 ! decodebin ! videoscale ! video/x-raw,width=1280,height=720
```

results in the following GStremer error:

```
GStreamer error: negotiation problem.
gstvideoencoder.c(1368): gst_video_encoder_chain (): /GstPipeline:pipeline1/GstVP8Enc:vp8enc0:
encoder not initialized
```

Logs: https://github.com/szemate/stb-tester/blob/save-video-issue/logs/with-save-video-with-videoscale-without-videorate.log

(2.) Without `--save-video` the test is running happily.
Logs: https://github.com/szemate/stb-tester/blob/save-video-issue/logs/without-save-video-with-videoscale.log

(3.) Interestingly, with `--save-video` but without resizing the video:

```
source_pipeline = rtmpsrc location=rtmp://testaut-2-pc-1/live/testaut-2-stb-1\ live=1
```

there is no GStreamer error but the test fails with `NoVideo`.
Logs: https://github.com/szemate/stb-tester/blob/save-video-issue/logs/with-save-video-without-videoscale-without-videorate.log

(4.) I found a useful forum post[1] that recommended adding `videorate` to the source pipeline because the H.264 encoder doesn't like that the timestamps restart counting from zero when the source pipeline is rebuilt. Modifying the video saving pipeline as:

``` py
video_pipeline = (
    "t. ! queue leaky=downstream ! videorate ! videoconvert ! "
    "vp8enc cpu-used=6 min_quantizer=32 max_quantizer=32 ! "
    "webmmux ! filesink location=%s" % save_video)
```

eliminates the above errors, however, the saved video is useless as it only contains a few frames from before the box gets power-cycled.
Saved video: https://raw.githubusercontent.com/szemate/stb-tester/save-video-issue/logs/corrupted-video.webm
[...]

PR #214, opened 2014-08-28
Fixes for issues with restarting the device under test

szemate:
A number of fixes for issues that occur when the video stream is restarted or the video resulution changes during a set-top-box restart.

I've validated the fixes by running 2000 iterations of a test that reboots the set-top-box, but more thorough testing with various test scripts is still in progress.

The first three commits address three individual issues.

The last two commits are alternative solutions for a side-effect of "Stop sink pipeline while the source pipeline is restarting" and I'm not happy with either one; I would be glad if you could propose a better one.

--

The different issues addressed in this pull request:

1. **Drop the last sample on underrun**

    > `stbt._display.last_sample` stores the most recent GSt buffer until
    > `stbt.gst_samples` removes it or it's replaced by a newer buffer.
    > To put it other way, if the video stream stops then `gst_samples`
    > yields the very last sample of the video stream even though it's called
    > minutes later.

2. **"Not negotiated" error when restarting source pipeline after VidiU
   EOS, if using `--save-video`**

    This is issue #211.

3. **Timestamps restart from 0 after a source pipeline restart**

    This makes `timeout_secs` unpredictable, and it corrupts the saved video.

4. **Video isn't streamable**

    Though this has a separate fix in the pull request, hopefully it will be
    fixed by a proper solution to issue 2 above.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -143,6 +143,8 @@ Additional options to stbt run
 
 --save-video=<file>
   Record a video (in the HTML5-compatible WebM format) to the specified `file`.
+  Multiple video files are created in case of video loss from VidiU or HD PVR
+  video-capture devices, indexed as `file`, `file1`, `file2` etc.
 
 Additional options to stbt record
 ---------------------------------
--- a/README.rst
+++ b/README.rst
@@ -143,8 +143,6 @@ Additional options to stbt run
 
 --save-video=<file>
   Record a video (in the HTML5-compatible WebM format) to the specified `file`.
-  Multiple video files are created in case of video loss from VidiU or HD PVR
-  video-capture devices, indexed as `file`, `file1`, `file2` etc.
 
 Additional options to stbt record
 ---------------------------------
--- a/stbt/__init__.py
+++ b/stbt/__init__.py
@@ -1780,6 +1780,7 @@ class Display(object):
     def restart_source(self):
         warn("Attempting to recover from video loss: "
              "Stopping source pipeline and waiting 5s...")
+        self.sink_pipeline.set_state(Gst.State.NULL)
         self.source_pipeline.set_state(Gst.State.NULL)
         self.source_pipeline = None
         GObjectTimeout(5, self.start_source).start()
@@ -1791,6 +1792,7 @@ class Display(object):
         warn("Restarting source pipeline...")
         self.create_source_pipeline()
         self.source_pipeline.set_state(Gst.State.PLAYING)
+        self.sink_pipeline.set_state(Gst.State.PLAYING)
         warn("Restarted source pipeline")
         if self.restart_source_enabled:
             self.underrun_timeout.start()
--- a/stbt/__init__.py
+++ b/stbt/__init__.py
@@ -1547,17 +1547,17 @@ class Display(object):
             appsink])
         self.create_source_pipeline()
 
-        self.video_file_name = save_video
-        self.video_file_index = 0
-        if self.video_file_name:
-            if not self.video_file_name.endswith(".webm"):
-                self.video_file_name += ".webm"
-            debug("Saving video to '%s'" % self.video_file_name)
+        if save_video:
+            if not save_video.endswith(".webm"):
+                save_video += ".webm"
+            debug("Saving video to '%s'" % save_video)
+            if os.path.isfile(save_video):
+                os.unlink(save_video)
             video_pipeline = (
                 "t. ! queue leaky=downstream ! videoconvert ! "
                 "vp8enc cpu-used=6 min_quantizer=32 max_quantizer=32 ! "
-                "webmmux ! filesink name=_stbt_filesink location=%s"
-                % self.video_file_name)
+                "webmmux streamable=true ! filesink location=%s append=true"
+                % save_video)
         else:
             video_pipeline = ""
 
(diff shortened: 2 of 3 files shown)


szemate:
We've been using these fixes for a few months and unfortunately they proved to be quite unstable. However, we found that putting a video converter into the signal chain that always outputs video, even if the video source drops, eliminates the above issues.

I would close this pull request now, however, the first issue that https://github.com/szemate/stb-tester/commit/aa21e4ad3ac1bcf188300e1a74d943c9146b9c6f resolves is still valid and it affects every kind of capture device. Shall I open a new pull request for that fix?

szemate:
> the first issue that szemate@aa21e4a resolves is still valid and it affects every kind of capture device

It's also valid for https://github.com/szemate/stb-tester/commit/f979fd7c1cd9a0c4c0a85dd3ec90d98344871cf2 (Re-timestamp buffers after source pipeline restart).

drothlis:
> We've been using these fixes for a few months and unfortunately they proved to be quite unstable.

Thanks for the update. Now you understand our reluctance to touch the code related to the source pipeline restart. :-)

> However, we found that putting a video converter into the signal chain that always outputs video, even if the video source drops, eliminates the above issues.

Excellent news. Are you at liberty to share a link to the video converter?

> however, the first issue that szemate@aa21e4a resolves is still valid and it affects every kind of capture device. Shall I open a new pull request for that fix?

If I understood correctly, aa21e4a only affects HDPVRs and Blackmagics, and it isn't necessary with your video converter. (I'm just checking that aa21e4a is really the commit you meant?)

> It's also valid for szemate@f979fd7 (Re-timestamp buffers after source pipeline restart).

Is that true? f979fd7's commit message says:

> Note that this fix doesn't substitute the patch "Stop sink pipeline
> while the source pipeline is restarting". Even though timestamps are
> in the right order, without that patch restarting the source pipeline
> results in one of the GStreamer errors "streaming task paused, reason
> not-negotiated (-4)" or "negotiation problem ... encoder not
> initialized".

drothlis:
@szemate I didn't actually answer your question. The answer is yes, open a pull request (or an issue) for each separate problem and then we can discuss the approach there.

We're definitely interested in fixing:

* Not providing a very old frame when you call `get_frame` after a `sleep` while video wasn't rolling.
* Monotonic timestamps when the source pipeline restarts (#211).
* `--save-video` when the source pipeline restarts.

However due to time constraints we're unlikely to do significant testing on any video-capture hardware other than the [stb-tester ONE](http://stb-tester.com/hardware.html) so you'll need to convince us that you haven't introduced any regressions. :-)  The preferred way to achieve this is by adding to the [hardware test suite](https://github.com/drothlis/stb-tester/tree/master/tests/hardware).

wmanley:
In particular I certainly be more interested in patches which move some of the complexity in this area  *out* of stb-tester and into lower layers such as GStreamer.  We're collecting a whole bunch of hardware-specific code that I'm not convinced belongs in stb-tester or that we can justify maintaining in the long-term.

One example of a welcome simplification would be removal of our internal queue, replacing it with using appsrc directly.  This would require patching appsrc to add a call like "pull-with-timeout" or a property saying how many buffers are in its internal queue.

Another simplification would be to use GStreamer's `watchdog` element to inform pipeline restarts rather than our own underflow/timeout implementation.

As @drothlis says, this area has proven to be hairy and prone to regression so improvements to the hardware test suite would also be very valuable.

szemate:
> If I understood correctly, aa21e4a only affects HDPVRs and Blackmagics, and it isn't necessary with your video converter.

Sorry, it really needs more explanation. Neither of the patches are required if we use the video converter. I just find the bugs severe enough that others might benefit from the fixes (and we still have a few rigs that are not yet equipped with the video converter).

The two issues I still find relevant:
- The actual timeout of `wait_for_match` is unpredictable if the source pipeline restarts while `wait_for_match` is running. This issue affects everyone using `stbt` with the `--restart-source` option.
- Providing a very old frame when you call `get_frame` after a sleep while video wasn't rolling. This issue affects everyone not using a video source that outputs video stream all the time.

Both issues are difficult to identify and they lead to test failures that first seem "mysterious", that's why I would be inclined not to discard them.

> Is that true? f979fd7's commit message says:
> > Note that this fix doesn't substitute the patch "Stop sink pipeline while the source pipeline is restarting"

This note wants to say that the commit does not eliminate the GStreamer errors with VidiU that the patch "Stop sink pipeline while the source pipeline is restarting" was meant to resolve. This commit *does* resolve the issue that the timeout of `wait_for_match` is getting unpredictable if the source pipeline gets restarted.

> Are you at liberty to share a link to the video converter?

I will ask @aiman and post a follow-up.

> In particular I certainly be more interested in patches which move some of the complexity in this area out of stb-tester and into lower layers such as GStreamer. (...) a welcome simplification would be removal of our internal queue, replacing it with using appsrc directly.

It sounds like a major change. For now, I will open two separate issues for these bugs just to make them visible because I'm not sure that I'm the only one who encountered them and they were not easy to identify.

drothlis:
Thanks @szemate for your investigation so far, and for raising #246 and #247 with very detailed descriptions.

DrewIshy:
I've been using a HD IP encoder which when the STB reboots it changes resolution from HD to SD then back to HD.  The tests work fine when you are not saving a file, but as soon as you start to save a file the vp8enc / webmmux crashes, and kills the sink pipe. The test finish OK as the source pipe is not affected.  The issue is that the test PC locks up for 10 mins when the sink pipe dies. 

After lots reading and trying some of the patches here for loss of the source video, which never happens on our system as its only the sink pipe which dies.  I changed the way the file is encoded to use x264enc and mpegtsmux which does implement re negotiation on resolution changes.
This is a simple change in code.py to video_pipeline to change vp8enc cpu-used=6 min_quantizer=32 max_quantizer=32 ! webmmux to x264enc ! mpegtsmux.

NOTE: this does stop the playback from the web site but that is fine for what I am doing.  

Hope this helps someone else who has the same issues

PR #225, opened 2014-09-24
stbt batch report: List pass/fail/error counts per testcase

drothlis:


drothlis:
Closing because I didn't find it that useful after all, and I haven't had the time to investigate the problem with the subtotal calculations (I don't remember the details but I think I noticed an error in the calculations at some point).

If anyone would find this useful please feel free to finish the job and raise another pull request. :-)

I did cherry-pick d1ac94e onto master as 315deb7.

PR #227, opened 2014-10-06
stbt.draw_text: Fade out older text

drothlis:
`stbt.press` uses `draw_text` to draw the pressed key name on the output
video stream for 3 seconds, and I always found it very confusing when
there were several keypresses in quick succession -- is the latest
keypress at the top or at the bottom? Now it is very clear.

TODO:

* [x] Rebase now that #255 has been merged.
* [x] Fix [TypeError](#issuecomment-58545894).
* [x] Add release note
* [x] Squash

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -32,6 +32,10 @@ UNRELEASED.
   executing test.  The tools for replay have not yet been written.  For more
   information, including the format definition see `_stbt/state_watch.py`.
 
+* The text drawn by `stbt.draw_text` and `stbt.press` on the recorded video now
+  fades out over a few seconds. This makes it easier to distinguish the new
+  messages from the old messages.
+
 ##### Bugfixes and packaging fixes since 22
 
 ##### Developer-visible changes since 22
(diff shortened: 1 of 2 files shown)


drothlis:
This will need rebasing now that #255 has been merged (there are bound to be conflicts).

wmanley:
Rebased on master.

wmanley:
I do have some concerns about race-conditions within this function.  They're somewhat obscured by the line:

        texts = self.text_annotations

On line 1964.  We use `texts` in the function like it's a local variable, but actually it's an alias for `self.text_annotations` and is mutable shared state (between this function/thread and anyone who calls `Display.draw`.

wmanley:
I believe that I've fixed the `TypeError` with the locking changes I made.

drothlis:
I've pushed a fixup that adds a couple of improvements
to the visualisation:

1. The text doesn't fade out to illegibility leaving the black
   background box for the last second.
2. The text fades out at a consistent speed regardless of the
   duration_secs of the text, so that earlier text will always
   be greyer than newer texts.

drothlis:
> I believe that I've fixed the TypeError with the locking changes I made.

I don't have a regression test for the TypeError so I'll take your word for it. :-)
You can merge if you like, or wait until we've used it for a few days in real testing, whatever you prefer.
Your changes look great.

drothlis:
I've pushed another fixup commit with a release note.

drothlis:
Merged. Thanks @wmanley for fixing my incompetence at multithreading. :-)

PR #239, opened 2014-12-15
stbt batch run: Fix `UnicodeEncodeError` when printing Unicode strings

wmanley:
When stbt is outputting to a terminal it will choose the encoding of the
stdout and stderr streams based on the LOCALE in use.  When it doesn't
detect a TTY attached to stdout/stderr it will default to ASCII output.

`stbt run` outputs directly to the terminal.  `stbt batch run` redirects
the output of `stbt run` to `tee` and then to a log file.

This means that if you `print u"Alfonso Cuarón"` in a test script (or
more likely `print` the output of `stbt.ocr`) whether you get a
`UnicodeEncodeError` will depend on whether you are using `stbt run` or
`stbt batch run`.

Python 3 fixes this issue by falling back to encoding as utf-8 rather than
ASCII.

This is technically a backwards-incompatible change but I don't expect that
it will break any test-scripts in the wild.  In fact I think the opposite
will be true.  You could however make the argument that it would be better
to stick to Python 2, warts and all or make the jump to Python 3 proper
rather than creating this weird middle-ground.  This would be a stronger
argument if `stbt` were a normal Python library which people were using in
web-servers and text-processing where complete understanding and control
over encoding is necessary, but for stb test scripts I think the path of
least-surprise is setting the encoding as in this patch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ install:
  - sudo apt-get update
  - sudo apt-get install
         expect
+        expect-dev
         gir1.2-gstreamer-1.0
         gir1.2-gudev-1.0
         gstreamer1.0-libav
--- a/_stbt/logging.py
+++ b/_stbt/logging.py
@@ -1,7 +1,6 @@
 # coding: utf-8
 
 import argparse
-import codecs
 import os
 import sys
 from contextlib import contextmanager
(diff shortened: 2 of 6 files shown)


LewisHaley:
@wmanley - This passes on my Fedora machine and the Travis error looks like a dependency issue. I'd like to port this to our custom build but would like to verify that the error Travis is reporting is not more severe that it appears?

wmanley:
I believe the Travis issue is just a test-setup or dependency issue like you say.  The test passes on my machine and I think this commit should be safe to cherry-pick.

wmanley:
ok, the problem was that Travis sets `LC_ALL=en_US.UTF-8` which overrides `LANG` that I was setting in my test.  Issue fixed by `unset LC_ALL`.

Have rebased on master too.  I think this is ready to merge.

wmanley:
@LewisHaley: Thanks for the nudge.  This is now merged.

LewisHaley:
Thanks to you too.

PR #244, opened 2015-01-05
Change `draw_text` font style

szemate:
We found the sans-serif 'duplex' style of the Hershey font easier to read than the default serifed 'triplex' style. Extending the black background also improves legibility.


drothlis:
Hi @szemate,
Can you attach some screenshots showing the difference?
Thanks, Dave.

wmanley:
I'd prefer that this is not configurable.  I think it's better if we work out what the best option is and just choose it.

Can you post a screenshot of before and after so we can discuss more clearly.

szemate:
The original "triplex" style:
triplex-without-margin

"triplex" style with 2px margin; I think it's definitely easier to read:
triplex-with-2px-margin

"triplex" style with 3px margin; it might be even better:
triplex-with-3px-margin

"duplex" style with 2px margin; it's really only a matter of taste:
duplex-with-2px-margin

szemate:
It's possible that we found "duplex" better because it resembles more a font used in a typical terminal application. Nevertheless, I think the larger black rectangle definitely improves legibility.

drothlis:
@szemate if you make it non-configurable, using the "duplex" style with 2px margin (and update the selftests so they pass, of course) we'll be happy to merge this.

szemate:
> make it non-configurable, using the "duplex" style with 2px margin (and update the selftests so they pass, of course)

Changes made. It's still not the best with characters like lower case "g", "y" and "q" because the black background doesn't include the [descender line](http://en.wikipedia.org/wiki/Typeface_anatomy) (see screenshot) but it's not an issue if we're printing capitalised text only with `draw_text`.

gyv

drothlis:
Merged. Thanks!

PR #245, opened 2015-01-05
stbt.py: Modified as_precondition to include original traceback

will-bryant:
Within the as_precondition context manager, whenever a UITestFailure
is raised, log the original exception's traceback with stbt.debug.

This makes it easier to determine exactly where and why a precondition
has failed, as the PreconditionError that is raised within as_precondition
only shows the message of the original exception, not the traceback.

### DIFF
--- a/stbt/__init__.py
+++ b/stbt/__init__.py
@@ -21,6 +21,7 @@ import re
 import subprocess
 import threading
 import time
+import traceback
 import warnings
 from collections import deque, namedtuple
 from contextlib import contextmanager
@@ -1357,6 +1358,9 @@ def as_precondition(message):
     try:
         yield
     except UITestFailure as e:
+        debug("stbt.as_precondition caught a UITestFailure exception and will "
+              "re-raise it as PreconditionError.\nOriginal exception was:\n%s"
+              % traceback.format_exc(e))
         exc = PreconditionError(message, e)
         if hasattr(e, 'screenshot'):
             exc.screenshot = e.screenshot  # pylint: disable=W0201


drothlis:
Looks good to me (minor style nitpicking aside). If only we had python3k's [chained exceptions](https://www.python.org/dev/peps/pep-3134/)!

stbt/__init__.py
```diff
@@ -25,6 +25,7 @@
 from collections import deque, namedtuple
 from contextlib import contextmanager
 from distutils.version import LooseVersion
+import traceback
```
drothlis: Can we keep the PEP8 import order here (alphabetical, "import blahs" before "from blah import blahs").

stbt/__init__.py
```diff
@@ -1357,6 +1358,7 @@ def as_precondition(message):
     try:
         yield
     except UITestFailure as e:
+        debug("Original exception:\n%s" % traceback.format_exc(e))
```
drothlis: Making the logging more explicit would probably be helpful. Something like:

    stbt.as_precondition caught a UITestFailure exception and will re-raise it as PreconditionError. The original exception was:

drothlis:
My last nitpick: "stbt.py" (mentioned in the commit message) doesn't exist anymore... and "include original traceback" is a bit vague. How about:

    stbt.as_precondition: Log original traceback

Anyway thanks for this pull request, it's clearly a useful addition.

will-bryant:
I've made some updates based on your comments now.

drothlis:
Merged. Thank you very much!

PR #249, opened 2015-02-02
stbt-run: Allow running tests by test name

drothlis:
You can now run a test with:

     stbt run test.py::my_test_function_name

This mimics py.test's command line interface, except that `stbt run test.py` won't run all `test_*` functions in the file.

This brings `stbt` to parity with the stb-tester ᴏɴᴇ appliance.

TODO:

* [x] Decide whether we should scrap this and figure out how to get py.test to create `stbt batch`-style reports, or whether we need to support this in `stbt run` as an interim measure.
* [x] Selftest that `stbt batch` works with this style of test name.
* [x] Release notes.
* [x] Review release notes wording.
* [x] Update stbt record to create a script in this style.
* [x] Update `docs/getting-started.md` etc.
* ~~Update tab-completion script for `stbt run` and `stbt batch run`.~~

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -138,16 +138,19 @@ Edit the test script to:
 
     import stbt
 
-    stbt.press('gamut')
-    stbt.wait_for_match('0000-gamut-complete.png')
-    stbt.press('checkers-8')
-    stbt.wait_for_match('0001-checkers-8-complete.png')
-    stbt.press('snow')
-    stbt.wait_for_motion()
+    def test_that_the_video_pattern_changes():
+        stbt.press('gamut')
+        stbt.wait_for_match('0000-gamut-complete.png')
+        stbt.press('checkers-8')
+        stbt.wait_for_match('0001-checkers-8-complete.png')
+        stbt.press('snow')
+        stbt.wait_for_motion()
 
 The test script is written in the Python programming language. You can use any
 Python feature, including Python's standard libraries; you can also use
 third-party Python libraries if you have them installed on your system.
+
+Each test case is a Python function that starts with **test_**.
 
 **stbt.press** takes a string that must be understood by the control specified
 in the configuration file.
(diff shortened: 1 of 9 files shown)


wmanley:
If we're going to recommend this approach above one test-case per file we really need to update all our documentation (and `stbt record`) to reflect that.  What do you think?

wmanley:
Also, I'm not that happy with the release-notes wording.

drothlis:
> If we're going to recommend this approach above one test-case per file we really need to update all our documentation (and `stbt record`) to reflect that. What do you think?

Agreed. I've added those to the todo list.

drothlis:
I rebased onto master, squashed Will's fixups, added my own fixups to
the release notes (unsquashed for easier review), taught `stbt record`
to produce python files in the new format, and updated
`docs/getting-started.md`.

The `stbt record` change is a slight compatibility break to people's
manual workflows (if they want to keep the old `stbt run` command line
they'll have to edit the script generated by `stbt record` to remove the
function), but I can't imagine there will be any automated process
affected by this change.

drothlis:
I think the remaining todo item in this pull request (Decide whether we should scrap this and figure out how to use pytest instead) can be decided in favour of merging this, as it allows us to get people using this format (which is forward-compatible with pytest) now. @wmanley if you're happy with this I'd like to merge it (after squashing the fixups of course).

wmanley:
> I think the remaining todo item in this pull request (Decide whether we should scrap this and figure out how to use pytest instead) can be decided in favour of merging this, as it allows us to get people using this format (which is forward-compatible with pytest) now.

Agreed

> @wmanley if you're happy with this I'd like to merge it (after squashing the fixups of course).

+1

drothlis:
I'm unlikely to finish the bash tab-completion script today so I'm going to merge this now.

drothlis:
I'll do the tab-completion in #253.

Issue #111, opened 2014-03-13
Long teardown (20s) with VidiU and GStreamer 1.0


wmanley:
With GStreamer 1.0 there is a 20s delay during VidiU pipeline shutdown.  This is probably a GStreamer bug.  Investigation required.

wmanley:
As this is a bug in GStreamer rather than stb-tester itself I'm tempted to say that this issue (as well as #150) should be closed.  Thoughts?

drothlis:
Closing upstream GStreamer / librtmp bugs that we don't intend to fix since it doesn't affect the hardware we use.

PR #250, opened 2015-02-03
stbt run: Treat AssertionErrors as test failures, not test errors

drothlis:
This brings stb-tester in line with traditional testing frameworks like
nose and pytest. Along with the introduction of `match()` in 33e8e356
and (in a future commit) `wait_until()`, this is part of a drive to
provide a more composible API. You can say things like:

* `assert match("xyz.png")`
* `assert not match("xyz.png")`
* `assert match("a.png") and match("b.png") and not match("c.png")`
* `assert wait_until(lambda: not match("xyz.png"))`
* `assert match_text("Main menu")`
* etc.

There are drawbacks to using `assert` instead of `wait_for_match`:

* The exception message won't contain the reason why the match
  failed (unless you specify it as a second parameter to `assert`,
  which is tedious and we don't expect you to do it), and
* The exception won't have the offending video-frame attached (so the
  screenshot that `stbt batch run` saves alongside the failing test
  logs will be a few frames later than the frame that actually caused
  the test to fail).

We hope to solve both problems at some point in the future: the first
one by using pytest and its assertion introspection[1] (which rewrites
python modules on import to provide the additional information on
assertion failures), and the second one by extending pytest's
module-rewriting to attach the video frame to the AssertionError.

[1] http://pytest.org/latest/assert.html#advanced-assertion-introspection

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -211,9 +211,10 @@ EXIT STATUS
 
 Test scripts indicate **failure** (the system under test didn't behave as
 expected) by raising an instance of `stbt.UITestFailure` (or a subclass
-thereof). Any other exception is considered a test **error** (a logic error in
-the test script, an error in the system under test's environment, or an error
-in the test framework itself).
+thereof) or `AssertionError` (which is raised by Python's `assert` statement).
+Any other exception is considered a test **error** (a logic error in the test
+script, an error in the system under test's environment, or an error in the
+test framework itself).
 
 
 HARDWARE REQUIREMENTS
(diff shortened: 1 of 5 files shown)


drothlis:
This requires #86.

LewisHaley:
> The exception won't have the offending video-frame attached (so the screenshot that stbt batch run saves alongside the failing test logs will be a few frames later than the frame that actually caused the test to fail).

Worth noting that due to #111, rather and "a few frames", when using a VidiU this is more like 20-30 seconds.

wmanley:
I think being able to `assert` conditions as Failures has real value.

I'm a bit confused as to why this requires #86.  I couldn't see that from the diff?

> Worth noting that due to #111, rather and "a few frames", when using a VidiU this is more like 20-30 seconds.

Agreed that this delay isn't great, but hopefully we can solve these issues in time. In the mean-time this give you an even greater motivation to solve #111 ;).

drothlis:
> Worth noting that due to #111, rather and "a few frames", when using a VidiU this is more like 20-30 seconds.

Noted, but I don't think this should stop us from merging this pull request. VidiU users can refrain from using `assert` (or better, they can fix #111). :-)  I'll mention this in the documentation.

wmanley:
If it's possible to merge this first I think that would be better. I think this is ready to go whereas the consequences of #86 are a little harder to think through.

drothlis:
I've squashed, rebased onto master, fixed the merge issues, and force-pushed.

Note that I can't change this pull request's base branch from `rename-uitestfailure-deprecate-uitesterror` to `master` (because of https://github.com/isaacs/github/issues/18) so don't merge via the github web interface.

wmanley:
> so don't merge via the github web interface

Thanks for the heads-up. Merged manually.

PR #254, opened 2015-02-10
stbt run -h: Document that you can specify a python function to run

drothlis:
This was missing from #249.


drothlis:
I've improved the wording and also updated the readme/manpage.

PR #255, opened 2015-02-13
stbt.Display: Add clock on each frame's left top corner

szemate:
This is a replacement for https://github.com/drothlis/stb-tester/pull/105. It seems that @Drachenfels has already deleted the branch so I'm duplicating his pull request from my fork.


szemate:
I have addressed comments https://github.com/drothlis/stb-tester/pull/105#discussion_r20440874 and https://github.com/drothlis/stb-tester/pull/105#issuecomment-74246305.

szemate:
I've fixed the failing self-test too.

drothlis:
I'll merge this now but it would be good to have a self-test for the clock, because we're going to be refactoring this in order to merge #227.

For example use `videotestsrc pattern=black` and run OCR on the frame, check that it matches the pattern `\d\d:\d\d:\d\d:\d\d`, wait a second, run OCR again, check that it's approx. 1 second greater than the previous value.
The main concern would be is OCR sufficiently reliable so that the test isn't flaky -- you'd want to test it over a full 24 hour period to know that OCR can handle all the possible time strings.

drothlis:
It looks really nice, thanks!

clock

PR #259, opened 2015-02-18
_stbt.control: Support user-defined remote control types

szemate:
As all remote control types are defined in the private `_stbt.control`
module, it is not possible to use a custom control class without
modifying stb-tester.

Provide users of stb-tester a pluggable remote control type that allows
them to use their own remote control implementation from an external
module.

Usage:
```
stbt run --control user:mypackage.mymodule.MyRemoteControl test.py
```

Our use case for this feature: we'd like to use a custom factory for `IRNetBoxRemote`
that gets the set-top-box's manufacturer (and thus the config file to use) from an inventory
rather than the stbt config file.

As the public interface of a remote control is really simple - it
consists of a single `press(key)` function - I don't think it's
necessary to define an abstract base class that users should derive
their remote control class from.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -109,6 +109,12 @@ Global options
     `--source-pipeline=videotestsrc`. A script like `press("18")` will change
     videotestsrc's pattern property (see `gst-inspect videotestsrc`).
 
+  user:<module>[.<attributes>].<callable>
+    A user-defined remote control from an external Python module.
+    `callable` is either a remote control class or a factory function that
+    returns a remote control instance. The remote control instance must provide
+    a `press(key)` method; `key` is passed from `stbt.press`.
+
   x11:<display>
     Send keypresses to a given X display using the xtest extension. Can be used
     with GStreamer's ximagesrc for testing desktop applications and websites.
--- a/_stbt/control.py
+++ b/_stbt/control.py
@@ -33,6 +33,7 @@ def uri_to_remote(uri, display=None):
              :(?P<output>\d+)
              :(?P<config>[^:]+)''', IRNetBoxRemote),
         (r'x11:(?P<display>.+)?', _X11Remote),
+        (r'user:(?P<module>[^\.]+)\.(?P<attributes>.+)', new_user_remote),
     ]
     for regex, factory in remotes:
         m = re.match(regex, uri, re.VERBOSE | re.IGNORECASE)
@@ -199,6 +200,19 @@ def new_tcp_lirc_remote(control_name, hostname=None, port=None):
     debug("TCPLircRemote: Connected to %s:%d" % (hostname, port))
 
     return LircRemote(control_name, _connect)
+
+
+def new_user_remote(module, attributes):
+    """Use a remote control from an external module."""
+    return _get_last_attr(__import__(module), attributes.split("."))()
+
+
+def _get_last_attr(obj, names):
+    """
+    >>> _get_last_attr(os, ["path", "join", "func_code", "co_name"])
+    'join'
+    """
+    return _get_last_attr(getattr(obj, names[0]), names[1:]) if names else obj
 
 
 class IRNetBoxRemote(object):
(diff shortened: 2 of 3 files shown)


wmanley:
I have to admit I'm not a fan of this.  It expands the public API to cover the previously internal `Control` class interface, and so makes it difficult to make changes in this area in the future.  Plugin APIs make more sense with proprietary software where the users don't have the freedom to make changes to the software themselves.

To solve the inventory problem I would recommend conditionally setting the configuration before launching `stbt run`.

szemate:
> To solve the inventory problem I would recommend conditionally setting the configuration before launching stbt run.

That's what we are doing now, but it's not very convenient because we always have to run an additional command, even if we only like to launch `stbt control`. You're right that we can always fork stb-tester though.

wmanley:
So what do you key your inventory system off?  e.g. how do you know which configuration to use?

szemate:
> how do you know which configuration to use?

We resolve the OID from the set-top-box's MAC address.

wmanley:
And how do you get the MAC address of the set-top box from the test PC?

szemate:
> And how do you get the MAC address of the set-top box from the test PC?

We have a PC configured as a software switch that the set-top-boxes are connected to. It intercepts DHCP traffic and gets the IP address and MAC address of a set-top-box from a DHCP ACK message. We also use the MAC address as primary key in a database to store various parameters of the box. The switch PC runs a RESTful service that provides these pieces of information to the test executor. The purpose of all these is that we don't have to change any configuration parameter when we are swapping boxes in the test rig.

wmanley:
So you do a lookup in the database from your PC, but how does your PC know which MAC address to use?

szemate:
The switch PC stores MAC addresses of the connected boxes locally and provides them to my PC on request.

szemate:
It's a HTTP service that provides the current IP and MAC address of the boxes connected to each network interface of the software switch.

drothlis:
I'm guessing the PC is statically configured to say "I run tests on the set-top box at slot X".
When they go to run a test, the PC asks:

1. What is the MAC address of the set-top box at slot X? (lookup using network switch's rest api)
2. What is the IR configuration for the set-top box with that MAC address? (lookup in database)

It seems to me that this is a matter of configuration in general, rather than infrared configuration specifically, so maybe it would be better to solve the general problem.

szemate:
> It seems to me that this is a matter of configuration in general, rather than infrared configuration specifically

Our problem basically is that using the set-up described above involves two steps:
1. Set the stbt config to use the right remote configuration.
2. Launch `stbt run` or `stbt control` etc.
(3. Don't forget to do 1. next time, otherwise you might use the wrong config.)

Using a custom remote control factory function would save us step 1. and prevent 3.

wmanley:
I'm still not understanding, maybe we can make this more concrete:

Using your scheme I am logged into a PC with 3 set-top boxes connected to it, A, B and C.  I run 

    stbt control

which set-top box should I be controlling?

szemate:
> which set-top box should I be controlling?

The config file of `A` looks something like:
```
[global]
control = irnetbox:irnetbox-proxy-host:1:<set-top-box manufacturer>.conf
resource = http://switch-pc/stb-A
```
The config file of `B`:
```
[global]
control = irnetbox:irnetbox-proxy-host:2:<set-top-box manufacturer>.conf
resource = http://switch-pc/stb-B
```
First I export the config file of the set-top-box that I'd like to use. Then I'm sending a request to `global.resource` that returns the set-top-box manufacturer. Then I can use the manufacturer to set the right value of `global.control`.

wmanley:
Ok, so the steps you follow are:

1. `export STBT_CONFIG_FILE=A.conf`
2. Set the stbt config to use the right IR configuration.
3. Launch stbt run or stbt control etc.
4. (Don't forget to do 1. next time, otherwise you might use the wrong config.)

I would recommend creating a bash function which you can call like

    choose_stb A

which would download/create an appropriate config and `export STBT_CONFIG_FILE`, combining steps 1 and 2 and removing the requirement for 4.

szemate:
> I would recommend creating a bash function which you can call like `choose_stb A` which would download/create an appropriate config and `export STBT_CONFIG_FILE`

This is a good idea. I wanted to avoid modifying the config file all the time but it is fine to create a temporary file to export.

drothlis:
Especially now that we've merged your #238, your `choose_stb` function can do this quite elegantly.

PR #260, opened 2015-02-18
Write tracing information to STBT_TRACING_SOCKET

wmanley:
With this pull request many of the actions that stbt takes are written to a socket given by the `STBT_TRACING_SOCKET` environment variable.  This allows the current state to be communicated to the user through an interactive UI, making it easier to understand what stbt is doing at any one time.  We also write these tracing logs to disk.

At some point in the future we should implement a playback mode, synchronised with video, so you can see exactly what's going on.  The video synchronisation could be tricky however.

TODO
=====

- [x] Tidy up commit messages
- [x] Placate Travis
- [x] Rename `test_run.test` to `test_run.test_case` for consistency with https://github.com/wmanley/stb-tester-service/pull/4
- [x] Document the date as ISO8601 consistently
- [x] Squash

**More info from commit message:**
A test run is in a particular state.  This state includes what test is currently executing, what line is currently executing, the current results directory, etc.  This state needs to be communicated live to the UI and should be useful during test replay when investigating what has gone wrong.

Imagine a description of the state of stbt in a data structure like:

    {
        "test_run": {
            "current_line": {
                "file": "tests/my_file.py",
                "line": 123,
            },
            "test_case": {
                "name": "tests/my_file.py::test_that_this_rocks",
                "file": "tests/my_file.py",
                "function": "test_that_this_rocks,
                "line": 87
            }
        }
    }

Conceptually the data-structure lives in the test-pack.  It is ephemeral
and only exists while the test job is executing.  Changes to this data
structure are serialised and sent over a socket to the UI so the current
state of the system can be displayed.  The serialisation includes
timestamps such that the state of the system can be inspected and
replayed as part of the test result.

The change serialisation format is \\r\\n separated JSON dictionaries.
Here's an example entry pretty printed for clarity's sake (in reality it
would be on a single line):

    {
        "state_change": {
            "time": "2014-11-28T20:55:26.092343Z",
            "changes": {
                "test_job.current_line": {
                    "file": "tests/my_file.py",
                    "line": 654
                }
            }
        }
    }

This record indicates that the current line changed to line 654 of file
"tests/my_file.py" at 20:55:26.0992343 UTC on 2014-11-28.  Notes:

* The root object has a single key "state_change".  This allows extensions
  in the future for other types of messages to be sent over the same
  protocol.

* The state_change message has two keys:

    * "time" which is the time of the change as an ISO8601 formatted string e.g. `"2014-11-28T20:55:26.092343Z"`.

    * "changes" is a dictionary of changes made to the state of the
      system.  The keys identify the value that is changing, and the
      values are the new value of the respective key.  The key is dot
      seperated strings identifying the subtree of the heirarical data
      structure that should be replaced.  e.g:

          "test_run.current_line": {"cow": "moo"}

      means:

          date["test_run"]["current_line"] = {"cow": "moo"}

      The fact that multiple values may be replaced in one message allows
      atomic changes to happen to the hierarchy.


drothlis:
You've documented the change format quite well. I'd like to see more documentation of the data structure that describes the state of stbt.  Maybe I'll have a go at this, as it will help me understand it better.

Are we calling this a public api? (Even if it's private I'll need the documentation to refer to.)

wmanley:
> Are we calling this a public api? (Even if it's private I'll need the documentation to refer to.)

I suppose there are 3 aspects to compatibility here:

* Any files written in this format now should be readable at any point in the future (backwards compatibility). - I think this is a **must**, it should also be fairly easy.
* Fields that are present now should be present in the future so external tools can read traces created now and at any time in the future (forwards compatibility) - I think this is a **good to have**.  We shouldn't break this unless there is a real and demonstrable end-user advantage to doing so.
* The python API is fixed so users can use it from their test scripts (API compatibility) - I think this is a **no for now**.  It could be possibly be very useful in the future.  We'll have to think about API and namespacing issues among others and have a clearer set of use-cases.

drothlis:
Agreed. I was thinking of the structured data, not of someone calling the python function `_stbt.state_watch.StateSender.set` or whatever.

extra/debian/control
```diff
@@ -29,6 +29,7 @@ Build-Depends: curl,
                python-gobject,
                python-jinja2,
                python-lxml,
+               python-lzma,
```
drothlis: These new dependencies need to go into the Fedora package too.

wmanley: Fedora 21 and 20 at least already install `lzma` as a part of their standard python installation:

    $ docker run -it fedora:20 python
    Python 2.7.5 (default, Jun 25 2014, 10:19:55) 
    [GCC 4.8.2 20131212 (Red Hat 4.8.2-7)] on linux2
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import lzma
    >>> lzma.LZMAFile
    <type 'lzma.LZMAFile'>

drothlis:
I rebased this branch onto master.

_stbt/state_watch.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def test_state_changes():
+    f = StringIO()
+    sw = StateSender(f)
+    sw.set({"test_run.line_number": 23},
+           time=datetime.datetime(2014, 3, 4, 12, 45, 12))
+    assert f.getvalue() == (
+        '{"state_change": {"changes": {"test_run.line_number": 23}, '
+        '"time": "2014-03-04T12:45:12"}}\r\n')
```
drothlis: This doesn't match the StateSender docstring which says that the timestamp is in mongo format (`"time": {"$date": ...}`).

PR #261, opened 2015-02-20
New test for clock visualisation

szemate:
A self-test for https://github.com/drothlis/stb-tester/commit/f77a67ea37fd975514aa9745c51d13e986e1f41a, addressing https://github.com/drothlis/stb-tester/pull/255#issuecomment-74672577.

The best outcome I could get was 5 failures out of 40000 test runs. If you have any idea how to make it more reliable, please help. How about skipping the test if the OCR fails? If it's only 0.01% of all executions then it shouldn't be a risk.

### DIFF
--- a/tests/run-clock-test-forever.sh
+++ b/tests/run-clock-test-forever.sh
@@ -1,0 +1,13 @@
+trap 'echo -e "\nTotal: $i\nFailed: $f"; exit 2' sigint sigterm
+date
+i=0
+f=0
+while true; do
+    if timeout 10 ./run-tests.sh test_clock_visualisation >&2; then
+        echo -n .
+    else
+        echo -n F
+        f=$((f + 1))
+    fi
+    i=$((i + 1))
+done
--- a/tests/test-stbt-py.sh
+++ b/tests/test-stbt-py.sh
@@ -396,6 +396,43 @@ test_press_visualisation() {
         verify.py
 }
 
+test_clock_visualisation() {
+    cat > test.py <<-EOF &&
+	import time
+	time.sleep(60)
+	EOF
+    mkfifo fifo || fail "Initial test setup failed"
+
+    stbt run -v \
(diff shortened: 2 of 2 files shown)


drothlis:
> The best outcome I could get was 5 failures out of 40000 test runs.

That's 1 in 8,000 test runs. Stb-tester has <3,000 commits, including
all branches and pull requests (even the ones that weren't merged). So
we'd have to run `make check` 3 times on every commit ever to have it
fail. I think that's good enough. :-) If it becomes troublesome then we
can do something about it.

I do wonder though if `tesseract_user_patterns=["\d\d:\d\d:\d\d:\d\d"]`
would give better results.

Just fyi re. `run-clock-test-forever.sh`, if you do `run-tests.sh -i` it
will test the version of `stbt` on your `$PATH` instead of running `make
install` into a temporary directory each time -- this will save some
time and let you run more iterations.

drothlis:
Cherry-picked 4caaa6a onto master as e0dc79f3. Thanks @szemate.

drothlis:
Hmmph for some reason I decided to soak-test this *after* merging it.
It failed 6/205 times:

    ......................F..FF.............................................
    ...........F.F.........................F................................
    .............................................................^C
    Total: 205
    Failed: 6

1. Using `tesseract_user_patterns` didn't help (it just made the test fail 100% of the time).
2. Setting the videotestsrc background to black seems to help a bit (failed 4/738 times).
3. Using the above fixes 1 & 2 at the same time seems better still (failed 3/2969 times).

I pushed a fix (cf403855).

szemate:
> It failed 6/205 times

I was using Tesseract 3.02 instead of the latest one, do you think it could account for the difference?

drothlis:
> I was using Tesseract 3.02 instead of the latest one, do you think it could account for the difference?

Yes, we've seen quite a lot of differences between minor tesseract versions.

PR #262, opened 2015-02-21
stbt run: Add the script's directory to PYTHONPATH

wmanley:
when running a test using the `script.py::function` scheme.  This makes `stbt run` behave more like `python script.py` and makes it easier to import other files in the same directory.

The old-style one test-script per file approach already does this.


drothlis:
:+1:

PR #263, opened 2015-02-22
Upgrade tablequeryjs to 0.18

drothlis:
For bugfixes (https://github.com/asimihsan/tablequeryjs/issues/19 and
https://github.com/asimihsan/tablequeryjs/issues/21) and for
`set_filter` so we don't have to use undocumented internal tablequery
APIs.

This means I can revert #258 which was a bit of a hack.

### DIFF
--- a/stbt-batch.d/templates/index.html
+++ b/stbt-batch.d/templates/index.html
@@ -83,20 +83,6 @@
         {% endfor %}
       </tr>
       </thead>
-      <!-- hidden row with unambiguous types for tablequeryjs: -->
-      <thead>
-      <tr style="display:none">
-        <td>1900-01-01 00:00:00</td>
-        <td>I hope this will only be interpreted as text</td>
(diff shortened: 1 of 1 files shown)


drothlis:
This introduces a minor regression: I can no longer enter a search like:

    timestamp >= "2015-02-21 19:30"

It only works if I specify the full time with seconds:

    timestamp >= "2015-02-21 19:30:00"

This is because tablequeryjs 0.16 introduced a check against both the column contents and the search string -- they must match `^\d{4}-\d{2}-\d{2}(( |T)\d{2}:\d{2}:\d{2}(Z)?)?$`.

wmanley:
I suppose `^\d{4}-\d{2}-\d{2}(( |T)\d{2}:\d{2}(:\d{2})?(Z)?)?$` would be better.

drothlis:
I think this is worth merging, and the regression isn't significant.

wmanley:
Ok

PR #267, opened 2015-03-03
stbt batch: Fix test scripts messing up test list by reading from stdin

wmanley:
`stbt batch run` parses its arguments into a list of test cases (with `parse_test_args` which is then piped into a while loop which executes each test in order.  If one of the test cases were to read from stdin the list of tests would be consumed by the test case, not by the test-runner causing all subsequent tests to be skipped.

This commit fixes that bug by redirecting stdin to `/dev/null` for test runs.


drothlis:
:+1:

PR #269, opened 2015-03-10
stbt power: Refactor and Aviosys Power 8800 pro support

wmanley:
The aim with this pull-request is to add support for [Aviosys USB NET POWER 8800 Pro] USB controlled power outlets, but given that we want to be writing power support in Python I've first partially rewritten `stbt power` in Python.

We've been meaning to [move to Python] for the power controllers for a while as shell isn't really that good a fit for argument parsing, URI matching and making HTTP requests.  Also it will make it easier to access this functionality from the Python scripts themselves.

With this PR all of the real functionality is still in `stbt-power.sh` but the actual `stbt power` command itself is in Python which opens up a migration path to allow the existing supported power controllers to be
ported to Python.

Note: none of the currently supported power controllers have any tests at all and the command itself is barely documented, so it's difficult to be sure that we haven't broken anything.  As a result it will have to fall to the users of these devices to test them manually or provide automated tests.

[move to Python]: https://groups.google.com/d/msg/stb-tester/0wFNSIimHn0/317YMmc4CesJ
[Aviosys USB NET POWER 8800 Pro]: http://www.aviosys.com/8800.html

**TODO**

- [x] Update release notes
- [x] Use consistent terminology to refer to these devices


stbt-power
```diff
@@ -1,183 +1,53 @@
...
+from _stbt.power import uri_to_pdu
+
+
+def main(argv):
+    parser = argparse.ArgumentParser(
+        description="Send commands to a computer controllable power switch",
+        formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument(
+        "--power-outlet", metavar="URI",
+        default=get_config("global", "power", ""), help=dedent("""\
```
drothlis: This should be `power_outlet`, which is what the help message says and what the previous version used.

tests/test-stbt-power.sh
```diff
@@ -0,0 +1,39 @@
+# Run with ./run-tests.sh
+
+test_that_stbt_power_status_prints_ON_by_default() {
+    [ "$(stbt power status)" = "ON" ] || fail
+}
+
+test_that_stbt_power_status_on_and_off_fail_by_default() {
+    ! stbt power on || fail "Cannot change power if there is no PCU"
+    ! stbt power off || fail "Cannot change power if there is no PCU"
```
drothlis: Typo? "PCU" should be "PDU"? (On both lines.)

tests/test-stbt-power.sh
```diff
@@ -0,0 +1,39 @@
...
+}
+
+test_that_selected_power_controller_comes_from_command_line_then_config() {
+    echo 0 >power-status
+    echo 0 >power-status2
+    set_config global.power "file:$PWD/power-status"
```
drothlis: `global.power_outlet` as per above comment.

_stbt/stbt-power.sh
```diff
@@ -0,0 +1,194 @@
+#!/bin/bash
+
+# WARNING: DON'T ADD SUPPORT FOR NEW POWER DEVICES TO THIS SCRIPT.
+# WE WANT TO PORT IT TO PYTHON FIRST. SEE
+# https://groups.google.com/d/msg/stb-tester/0wFNSIimHn0/317YMmc4CesJ
```
drothlis: You could update this message to point the developer to the python script instead.

tests/test-stbt-power.sh
```diff
@@ -0,0 +1,39 @@
...
+
+    stbt power --power-outlet=file:$PWD/power-status2 on
+    [ "$(< power-status2)" = "1" ] || fail "Wrong power control outlet selected"
+
+    stbt power on
+    [ "$(< power-status)" = "1" ] || fail "Wrong power control outlet selected"
+}
+
+test_that_stbt_power_status_prints_status() {
+    set_config global.power "file:$PWD/power-status"
```
drothlis: `global.power_outlet`

_stbt/power.py
```diff
@@ -0,0 +1,188 @@
...
+    for regex, factory in remotes:
+        m = re.match(regex, uri, re.VERBOSE | re.IGNORECASE)
+        if m:
+            return factory(**m.groupdict())
+    raise ConfigurationError('Invalid PDU URI: "%s"' % uri)
+
+
+class _NoPdu(object):
+    def set(self, _power):
+        raise RuntimeError("No Computer Controlable Power Outlet Configured")
```
drothlis: Controllable (double L).

drothlis: Also, the sentence case isn't consistent with the rest of our error messages.

stbt-power
```diff
@@ -1,183 +1,53 @@
...
+             Address of the power device and the outlet on the device.
+             The format of <uri> is: (aten|ipp|pdu):<hostname>:<outlet>
+               aten|ipp|pdu  Model of the controllable power supply:
+                             * aten: ATEN (all models)
+                             * ipp: IP Power 9258
+                             * pdu: PDUeX KWX
```
drothlis: This help message needs to be updated with the new device you've added.

_stbt/power.py
```diff
@@ -0,0 +1,188 @@
...
+        self.device.write("p1=%i\n" % power)
+        self.device.readline()
+
+    def get(self):
+        self.device.write("readio\n")
+        self.device.readline()
+        return {
+            'IO:5\r\n': True,
+            'IO:0\r\n': False,
+        }[self.device.readline()]
```
drothlis: If this was written in a less condensed manner you'd get better error messages, i.e.

```
response = self.device.readline()
if response == "IO:5\r\n":
    return True
elif response == "IO:0\r\n":
    return False
else:
    raise RuntimeError("Unexpected response from Power 8800 device: %s" % line.strip())
```

drothlis:
There's a regression here: Previously you could run the `status` subcommand without specifying which socket (if you're using a multi-socket device), and it would print the status of each power socket, in a format that varied from device to device.

I'm leaning towards breaking that behaviour -- having this ported to python is worth it.

LewisHaley:
Considering the old API is `on|off|status`, what's your reasoning for `set|get`? The existing API is much more explicit - `power.turn_on()` vs. `power.set(on=True)`. I can't say I've heard someone say they were going to "set the power on" for any kind of electrical device.

FYI We have done this python wrapper for `stbt-power` ourselves, and here are a few notes:
- The only time we query the status of an outlet is when we are about to turn it on or turn it off. i.e. `power.turn_power_on` is actually `power.turn_power_on_if_not_already_on`.
- It's important to keep/make a distinction between the *mains power* and the *power of the SUT*. When someone says "turn the STB on" do they mean turn its mains power on or do they mean that it already has mains power and should be "booted".

Just something to think about.

wmanley:
> Considering the old API is on|off|status, what's your reasoning for set|get? The existing API is much more explicit - power.turn_on() vs. power.set(on=True). I can't say I've heard someone say they were going to "set the power on" for any kind of electrical device.

I agree that `power.set(on=True)` isn't a great API, but this PR was not intended to introduce any new APIs, just to be an internal re-factoring.  At the same time if we were to make an API public I do think a method taking a boolean value is appropriate because it makes it easier to pass "poweredness" properties into helper functions and manipulate it.  e.g. with this you can easily create constructs like 

    def flip_power():
        pdu.set(not pdu.get())

Either way I think it would be best to discuss this API once we have a PR proposing exposing it publically.

wmanley:
> There's a regression here: Previously you could run the status subcommand without specifying which socket (if you're using a multi-socket device), and it would print the status of each power socket, in a format that varied from device to device.
>
> I'm leaning towards breaking that behaviour -- having this ported to python is worth it.

Given that there are no tests, nor even documentation describing the old format I think as a practical matter it would be very difficult not to break this behaviour.  I would suggest merging this and if it affects someone then that's a good opportunity to more clearly define what the correct behaviour is.

drothlis:
> Given that there are no tests, nor even documentation describing the old format I think as a practical matter it would be very difficult not to break this behaviour. I would suggest merging this

Agreed.

wmanley:
>Typo? "PCU" should be "PDU"?

    +        raise RuntimeError("No Computer Controlable Power Outlet Configured")

I've been rather inconsistent with naming here.  I don't think PDU is appropriate as that stands for [Power Distribution Unit] which implies multiple power outputs for a single power input, but doesn't imply that they are computer controllable.  I made-up "PCU" meaning "Power Control Unit", which [some people seem to also use][1], but if we can avoid obscure [TLA]s while still being terse then all the better.  I've also written "Computer Controllable Power Outlet" in a few places.

I agree that we should choose one and be consistent but I'm not sure which.

[Power Distribution Unit]: http://en.wikipedia.org/wiki/Power_distribution_unit
[1]: https://www.google.co.uk/webhp?q=power%20control%20unit
[TLA]: http://en.wikipedia.org/wiki/Three-letter_acronym

drothlis:
> I agree that we should choose one and be consistent but I'm not sure which.

I wouldn't worry too much about the name not implying that it's computer controllable -- I think we can take that for granted since we're talking about a computer-controlled automated testing system. I think "Power control unit" is best for user-facing strings (documentation, error messages), and "pcu" for internal variable names etc. is fine. If the phrase "Power Control Unit" is in the commit message somewhere then if someone is losing sleep over what the variable name means they can look it up. :-)

In our command-line argument "pdu" means a specific model of a specific brand, which is a bit unfortunate but it's too late to change that.

wmanley:
> I think "Power control unit" is best for user-facing strings (documentation, error messages), and "pcu" for internal variable names etc. is fine.

I've standardised on "Power Outlet" and "Outlet" for short.

wmanley:
I consider this ready to merge now.

drothlis:
I pushed some very minor wording fixups. Other than that: :+1:

PR #271, opened 2015-03-12
stbt-batch: Fix inconsistency in help message

szemate:
The help message of `stbt batch` says that "For help on a specific
subcommand do `stbt batch <subcommand> --help`". However, running
`stbt batch run --help` exits with status code 1 saying
"Invalid option '--'. Use '-h' for help."

Advise users to use `stbt batch <subcommand> -h` to avoid this error.


drothlis:
Merged. Thanks!

PR #272, opened 2015-03-15
tv_driver.py: Add video server creation condition

saddopop:
If '--tv-driver' is 'assume' mode, there is no need to provide video server.
This commit adds video server creation condition. Video server is created
only if both 'manual' and 'fake' mode are given.

### DIFF
--- a/_stbt/tv_driver.py
+++ b/_stbt/tv_driver.py
@@ -161,16 +161,19 @@ def add_argparse_argument(argparser):
              default=get_config("camera", "tv_driver", "manual"))
 
 
-def create_from_args(args, video_generator):
-    desc = args.tv_driver
-    video_server = _HTTPVideoServer(
+def make_video_server(video_generator):
+    return _HTTPVideoServer(
         video_generator,
         video_format=get_config('camera', 'video_format'))
+
+
+def create_from_args(args, video_generator):
+    desc = args.tv_driver
     if desc == 'assume':
         return _AssumeTvDriver()
     elif desc.startswith('fake:'):
-        return _FakeTvDriver(desc[5:], video_server)
+        return _FakeTvDriver(desc[5:], make_video_server(video_generator))
     elif desc == 'manual':
-        return _ManualTvDriver(video_server)
+        return _ManualTvDriver(make_video_server(video_generator))
     else:
         raise RuntimeError("Unknown video driver requested: %s" % desc)


wmanley:
`--tv-driver=assume` only really exists for testing stb-tester itself, so unless there's a good reason for this I'd prefer to avoid the additional complexity. What's your use-case?

saddopop:
Recently, I have tried to use ```stbt-camera calibrate --tv-driver``` for my smart TV calibration because there was always scattered reflection between my webcam and TV screen. I think that ```--tv-driver=assume``` enables TV to be calibrated without the video server by lighttpd if calibration pattern is already shown on TV screen. I understood ```--tv-driver=assume``` was not just for testing stb-tester itself but use-case for real calibration with user-defined calibration pattern. And I also found the video server instance was only passed as a parameter when both ```manual``` and ```fake``` were given.
Finally,  I respect your opinion and understand your concern. And I am also afraid that I cannot still understand well about the feature. If you think it is not acceptable, I will agree with you. (-:

wmanley:
Thanks for the response.  Is there a particular reason that you don't want the http-server to be started?  Is it a concern about efficiency, or security, or something else?

_stbt/tv_driver.py
```diff
@@ -163,9 +163,10 @@ def add_argparse_argument(argparser):
 
 def create_from_args(args, video_generator):
     desc = args.tv_driver
-    video_server = _HTTPVideoServer(
-        video_generator,
-        video_format=get_config('camera', 'video_format'))
+    if desc.startswith('fake:') or desc == 'manual':
+        video_server = _HTTPVideoServer(
+            video_generator,
+            video_format=get_config('camera', 'video_format'))
```
wmanley: My biggest concern with the patch is that it adds an additional `if` statement that must be kept in-sync with the if statement below.  Perhaps a better approach would be to move the creation of `video_server` below into the existing `if`, `elif` ladder.

Perhaps we can use Python's closures to avoid duplication and additional conditional logic.  Something like:

    def make_video_server():
        return _HTTPVideoServer(
            video_generator,
            video_format=get_config('camera', 'video_format'))

and then call that when instantiating `_FakeTvDriver` or `_ManualTvDriver`.

saddopop:
Thank you for your kind feedback.
I just think that the http-server doesn't need to be started when instantiating `_AssumeTvDriver`. The `video_server` variable is used when instantiating `_FakeTvDriver` or `_ManualTvDriver`. As your comment, I will commit it again.

wmanley:
This fell through the cracks a little.  I've rebased on master and made a small change to the location of `make_video_server` and merged.  Thanks.

PR #277, opened 2015-04-29
stbt power: Implement ATEN PE6108G support in python

martynjarvis:
Implement [ATEN PE6108G](http://www.aten.com/products/Energy-Intelligence-Rack-PDU/Switched-PDU/Power-Distribution-Unit~PE6108.html) support in python and remove support from bash script. Add unit tests for the module.

These changes introduce the `pysnmp` and `mock` python modules as dependencies.

TODO:

- [x] Install pysnmp in the Fedora & Ubuntu packages (`extra/fedora/stb-tester.spec.in` and `extra/debian/control`, respectively). On Ubuntu the package seems to be called `python-pysnmp4`.
- [x] Move the unit tests to `tests/test_power.py` so that we don't have to add a dependency on pymock in production.
- [ ] Get travis CI passing (install pysnmp and pymock in `.travis.yml`).


drothlis:
Nice one.

You'll have to install pysnmp and pymock in `.travis.yml` to get the CI passing, and pysnmp in the Fedora & Ubuntu packages, and move the unit tests to a file under `tests/` so that we don't have to add a dependency on pymock in production. I've edited this pull request's description to add a TODO list.

_stbt/power.py
```diff
@@ -191,3 +195,130 @@ def write(self, data):
...
+        if isinstance(result, NoSuchObject):
+            raise RuntimeError("Invalid outlet {}".format(self.outlet))
+
+        if not isinstance(result, Integer):
+            raise RuntimeError("Unexpected result ({})".format(result))
+
+        return result
+
+
+class TestATEN(unittest.TestCase):
```
wmanley: Most of our other (Python) tests use nose.  Is there any reason you had to use unittest in this instance?  I tend to prefer nose as it can be less verbose.

_stbt/power.py
```diff
@@ -191,3 +195,130 @@ def write(self, data):
...
+        from pysnmp.entity.rfc3413.oneliner import cmdgen
+        cmdgen.UdpTransportTarget = mock.Mock()
+        time.sleep = mock.Mock()
+
+    def setUp(self):
+        oid_string = "1.3.6.1.4.1.21317.1.3.2.2.2.2.{0}.0"
+        self.outlet = 1
+        self.oid = oid_string.format(self.outlet + 1)
+
+        self.aten = _ATEN_PE6108G('fakehost.name', self.outlet)
```
wmanley: I'd prefer if you instantiated the object via `uri_to_power_outlet`.  That way your regex in the factory code gets a little testing too.

martynjarvis: Changed in https://github.com/martynjarvis/stb-tester/commit/cc2cb364fe3fb58f606f024866cd93fa2fad4eec

_stbt/power.py
martynjarvis: No major reason. I can change it to use nose if that is preferred.

drothlis:
The a857089 travis failure may be an unrelated intermittent test error. I've set Travis to re-run.

drothlis:
`test_that_stbt_batch_reports_results_directory` failed again. It seems unlikely to be related to this branch, but I need to investigate.

_stbt/power.py
drothlis: > I can change it to use nose if that is preferred. 

Yes please, if it isn't too much trouble. It would be nice to keep the test styles consistent.

drothlis:
I've raised #278 to investigate the `test_that_stbt_batch_reports_results_directory` failure.

_stbt/power.py
martynjarvis: > Yes please, if it isn't too much trouble. It would be nice to keep the test styles consistent.

Changed in https://github.com/stb-tester/stb-tester/commit/c5984140ac15a5ac68cb1d395151746e0047523a

extra/debian/control
```diff
@@ -33,6 +33,7 @@ Build-Depends: curl,
                python-lzma,
                python-nose,
                python-opencv,
+               python-pysnmp4,
```
drothlis: You also need to add `python-pysnmp4` to the `Depends` section of the `stb-tester` package.

(Also leave this line in the `Build-Depends`, so that we can run `make check` when building the deb.)

tests/test_power.py
```diff
@@ -0,0 +1,72 @@
...
+aten = uri_to_power_outlet('aten:mock.host.name:1')
+
+
+@mock.patch('pysnmp.entity.rfc3413.oneliner.cmdgen.CommandGenerator')
+def test_aten_get_on(mock_class):
+    mock_class.return_value.getCmd.return_value = mock_data(2)
+
+    result = aten.get()
+
+    eq_(result, True)
```
wmanley: Please use `assert results is True` here.  I find `assert` statements more readable, and the advantages of `eq_` will be negated once we move to py.test with it's [assert inspection](http://pytest.org/latest/assert.html#assert-with-the-assert-statement).

tests/test_power.py
```diff
@@ -0,0 +1,72 @@
...
+    """Match the format of the data returned from pysnmp"""
+    return (None, None, None, [(oid, Integer(int_value))])
+
+
+cmdgen.UdpTransportTarget = mock.Mock()
+time.sleep = mock.Mock()
+
+outlet = 1
+oid = "1.3.6.1.4.1.21317.1.3.2.2.2.2.{0}.0".format(outlet + 1)
+aten = uri_to_power_outlet('aten:mock.host.name:1')
```
wmanley: Please implement setup as a `@contextmanager`.  With this as it is the code will be run at import time rather than before each test and the same aten object could be shared between each of the tests.  This should allow you to tidy-up the mock patching too.

Something like:

    @contextmanager
    def mocked_aten():
        cmdgen.UdpTransportTarget = mock.Mock()
        time.sleep = mock.Mock()

        outlet = 1
        oid = "1.3.6.1.4.1.21317.1.3.2.2.2.2.{0}.0".format(outlet + 1)
        aten = uri_to_power_outlet('aten:mock.host.name:1')

        with mock.patch('pysnmp.entity.rfc3413.oneliner.cmdgen.CommandGenerator') as mock_class:
            yield (mock_class.return_value, aten)

Although I'd still be a little concerned about the side effects of `time.sleep = mock.Mock()`.  Should that also be `mock.patch('time.sleep')`, so `time.sleep` is not permanently mocked'?

wmanley:
I've fixed #278 which was causing the test failures.  Could you rebase and squash your commits appropriately and hopefully Travis will then be happy.

Thanks for your continued efforts.

extra/debian/control
```diff
@@ -67,6 +68,7 @@ Depends: ${shlibs:Depends},
          python-lxml,
          python-lzma,
          python-opencv,
+         python-pysnmp4,
          python-serial,
          snmp,
```
drothlis: I almost forgot: Can you please remove the dependency on `snmp` now that `stbt-power.sh` doesn't need it.

drothlis: (and `net-snmp-utils` in the fedora spec file too).

martynjarvis:
Rebased and squashed down to two commits. Travis is passing now, thanks.

wmanley:
Merged. Thanks for your perseverance.

PR #279, opened 2015-04-30
Add CONTRIBUTING.md guidelines

drothlis:
If this file exists in the root of the repo, GitHub displays a link to
it when you open a pull request:
https://github.com/blog/1184-contributing-guidelines

I've tried to keep it really short (700 words, and you only need to read
the second half of it if you're adding new dependencies).

What do you think? If you were thinking of contributing to stb-tester
would you appreciate this information, or is a list like this more
likely to put you off?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,109 @@
+Contributing to stb-tester
+==========================
+
+Our preferred workflow is via GitHub Pull Requests.
+
+Feel free to open a pull request if you want to start a discussion even if your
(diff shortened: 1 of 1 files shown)


wmanley:
It certainly reads well.  What's the effect that you're hoping for by introducing this document?

drothlis:
>  What's the effect that you're hoping for by introducing this document?

* Less back-and-forth during review.
* A place to point people at without re-typing it in each pull request.
* Documentation (for myself as much as anybody else) on how to test the Ubuntu & Fedora packaging.

PR #281, opened 2015-05-18
 stbt run: Fix UnicodeEncodeError from unicode exception messages 

drothlis:
**TODO:**

* [x] Save the original `sys.stderr` instead of relying on `sys.__stderr__`.
* [x] Add test with `stbt batch`.
* [x] Make it so there isn't a regression if you raise an exception with a utf-8 encoded byte-string.
* [x] More testcases, including `stbt run` behaviour when `print`ing byte-strings to a terminal.
* [x] Fix `print`ing non-ascii byte strings.

testcase | to UTF-8 capable terminal | to /dev/null | to file |
---|---|---|---
print byte-string with utf-8 | ✓ | ✓ | ✓ |
print unicode string | ✓ | ✓ | ✓ |
raise exception with byte-string with utf-8 | ✓ | ✓ | ✓ |
raise exception with unicode string | †  | ✓ | † |

† Doesn't break, but the last line of the traceback looks yucky:

```
FAIL: dave.py: AssertionError: öゲ
  File "./stbt-run", line 107, in <module>
    execfile(_filename)
  File "/home/drothlis/work/stb-tester.com/stb-tester-one/stb-tester/stb-tester/dave.py", line 10, in <module>
    assert False, u"öゲ"
AssertionError: \xf6\u30b2
```


drothlis:
I've addressed the remaining todo items, rebased onto master & force-pushed. I think this is ready to merge.

drothlis:
@LewisHaley said:

> I fixed this error on our fork, using https://pythonhosted.org/kitchen/api-text-converters.html#kitchen.text.converters.getwriter.

What are the benefits of using kitchen vs. the built-in `codecs`?

> I'd recommend adding a second test so that you cover both byte-string type containing unicode and unicode type containing unicode, because the handling is different.

Good point. I suspect the former will fail currently, as suggested by the "breaking change" in the release note.

> Unfortunately I can't push my patches easily because too much has diverged.

Can you post the diff as a comment, even if it doesn't apply cleanly to our master branch?

Thanks very much for the feedback!

LewisHaley:
> What are the benefits of using kitchen vs. the built-in codecs?

My understanding is that `kitchen` is solving the exact problem of "I want utf-8 stdout and stderr but changing the stderr streamwriter to utf-8 causes Unicode errors".

```
Author: Lewis Haley <lewis.haley@youview.com>
Date:   Wed Aug 26 09:48:54 2015 +0100

    stbt-run: fix Unicode in exceptions/tracebacks
    
    As per the warning in the Python docs on unicode [1], it's not a good idea
    to use a `StreamWriter` to overwrite sys.stderr because it will cause
    issues with writing tracebacks. It is recommended to use the `kitchen`
    module, which is what this commit implements. Note that the additional
    change to the string representation of the exceptions passed into
    `report_exception` (along with an additional selftest) was necessary to
    get the new test to pass and vice-versa (i.e. fixing `report_exception`
    in its own commit wouldn't work because of the sys.stderr `StreamWriter`).
    
    [1] https://pythonhosted.org/kitchen/unicode-frustrations.html#frustration-5-exceptions

diff --git a/stbt-run b/stbt-run
index d15d67f..a20404e 100755
--- a/stbt-run
+++ b/stbt-run
@@ -7,6 +7,7 @@ https://github.com/stb-tester/stb-tester/blob/master/LICENSE for details).
 """
 
 import argparse
+from kitchen.text.converters import getwriter
 import os
 import sys
 import traceback
@@ -47,7 +48,7 @@ def _setup_utf8_output():
     if sys.stdout.encoding is None:
         sys.stdout = codecs.getwriter('utf8')(sys.stdout)
     if sys.stderr.encoding is None:
-        sys.stderr = codecs.getwriter('utf8')(sys.stderr)
+        sys.stderr = getwriter('utf8')(sys.stderr)
 
 _setup_utf8_output()
 
@@ -61,10 +62,33 @@ def import_by_filename(filename_):
     return __import__(module_name)
 
 
+def _utf8_safe_exc_message(exception):
+    """Safely make a `str` repr of `exception`."""
+    try:
+        return str(exception)
+    except UnicodeEncodeError:
+        return unicode(exception).encode("utf-8")
+
+
+def _utf8_safe_print_exception(etype, value, tb):
+    """Print a traceback and exception message safely.
+
+    A thin wrapper around `traceback.print_exception` which encodes the
+    exception message contained in the traceback.
+    """
+    etype, exc, tb = sys.exc_info()
+    for line in traceback.format_exception(etype, exc, tb)[:-1]:
+        sys.stderr.write(line)
+    if isinstance(exc.message, unicode):
+        exc.message = exc.message.encode('utf-8')
+    for line in traceback.format_exception_only(etype, exc.message):
+        sys.stderr.write(line)
+
+
 def report_exception(script_name, exc, message):
     """Print the script's failing exception."""
     sys.stdout.write("FAIL: %s: %s: %s\n" % (
-        script_name, type(exc).__name__, message))
+        script_name, type(exc).__name__, message.decode('utf-8')))
 
 
 def report_failures(list_of_failures):
@@ -148,14 +172,14 @@ except Exception as e:  # pylint: disable=W0703
             registered_failures.remove(registered_failure)
             break
     report_failures(registered_failures)
-    error_message = str(e)
+    error_message = _utf8_safe_exc_message(e)
     if not error_message and isinstance(e, AssertionError):
         error_message = traceback.extract_tb(sys.exc_info()[2])[-1][3]
     report_exception(args.script, e, error_message)
     if hasattr(e, "screenshot") and e.screenshot is not None:
         stbt.save_frame(e.screenshot, "screenshot.png")
         sys.stderr.write("Saved screenshot to '%s'.\n" % ("screenshot.png"))
-    traceback.print_exc(file=sys.stderr)
+    _utf8_safe_print_exception(*sys.exc_info())
     if isinstance(e, stbt.UITestFailure):
         sys.exit(1)  # Failure
     else:
@@ -163,7 +187,8 @@ except Exception as e:  # pylint: disable=W0703
 else:
     exception, exit_code = report_failures(stbt.get_registered_failures())
     if exception is not None:
-        report_exception(args.script, exception, str(exception))
+        error_message = _utf8_safe_exc_message(exception)
+        report_exception(args.script, exception, error_message)
     sys.exit(exit_code)
 finally:
     stbt.trace.deactivate()
diff --git a/tests/test-stbt-run.sh b/tests/test-stbt-run.sh
index 5877c6a..3072ea0 100644
--- a/tests/test-stbt-run.sh
+++ b/tests/test-stbt-run.sh
@@ -446,3 +446,27 @@ test_that_trace_includes_the_right_functions() {
     grep -q "ExcludedClass.not_traced_method" trace.log &&
       fail "Method of excluded class should not be traced" ||:
 }
+
+test_that_unicode_in_exceptions_is_handled() {
+    # This fails if sys.stderr is directed to a "plain" `codecs.getwriter('utf8')`
+    # rather than using `kitchen.text.converters`.
+    cat > test.py <<-EOF
+	# coding: utf-8
+	raise Exception("Got: %s" % "ŬŋĩčœĐę")
+	EOF
+    stbt run -v test.py
+    cat log | grep -q -E '^FAIL: test.py: Exception: Got: ŬŋĩčœĐę$' ||
+        fail "Exception with unicode in byte-string message not reported"
+    cat log | grep -q -E '^Exception: Got: ŬŋĩčœĐę$' ||
+        fail "Exception with unicode in byte-string message not printed in traceback"
+
+    cat > test.py <<-EOF
+	# coding: utf-8
+	raise Exception("Got: %s" % u"ŬŋĩčœĐę")
+	EOF
+    stbt run -v test.py
+    cat log | grep -q -E '^FAIL: test.py: Exception: Got: ŬŋĩčœĐę$' ||
+        fail "Exception with unicode in byte-string message not reported"
+    cat log | grep -q -E '^Exception: Got: ŬŋĩčœĐę$' ||
+        fail "Exception with unicode in byte-string message not printed in traceback"
+}
```

drothlis:
After discussing with @wmanley we've decided not to merge this until we handle exceptions with byte-strings with utf-8 in them (i.e. removing the backwards incompatibility that this pull request currently introduces). We'll also check what the current behaviour of `print`ing byte-strings with utf-8 is by writing more tests.

stbt-run
```diff
@@ -106,7 +107,7 @@ try:
         sys.path.insert(0, os.path.dirname(_filename))
         execfile(_filename)
 except Exception as e:  # pylint: disable=W0703
-    error_message = str(e)
+    error_message = unicode(e)
     if not error_message and isinstance(e, AssertionError):
         error_message = traceback.extract_tb(sys.exc_info()[2])[-1][3]
     sys.stdout.write("FAIL: %s: %s: %s\n" % (
```
wmanley: Perhaps:

    try:
        error_message = unicode(e).encode('utf-8')
    except UnicodeDecodeError:
        error_message = str(e)

?

drothlis:
I've fixed the regression. Exception messages can now contain unicode or (for backward compatibility) non-ascii byte-strings.

I've also added a testcase to see what happens if you `print` non-ascii byte-strings. It turns out that we don't handle that either. But I think fixing it (if it's even possible) is outside the scope of this pull request; fixing the exception behaviour is valuable and is often triggered with exception messages originating from OCR output.

drothlis:
> I've also added a testcase to see what happens if you print non-ascii byte-strings. It turns out that we don't handle that either. But I think fixing it (if it's even possible) is outside the scope of this pull request.

@LewisHaley's suggestion of using [kitchen.getwriter](https://kitchen.readthedocs.org/en/latest/api-text-converters.html#print) does fix this. I'd still like to leave that to a separate pull request.

drothlis:
I've raised #325 for the changes to allow `print`ing non-ascii byte strings using `kitchen`. We can review it later, for now I just want to get this fix into v24.

wmanley:
I've pulled @drothlis/@LewisHaley's kitchen changes onto this PR after-all and done a bit of an audit myself.  @drothlis: Please review.

drothlis:
Current behaviour (as of bbd5c7d, with `kitchen.text.converters.getwriter` + `traceback.print_exc`):


testcase | to UTF-8 capable terminal | to /dev/null | to file |
---|---|---|---
print byte-string with utf-8 | ✓ | ✓ | ✓ |
print unicode string | ✓ | ✓ | ✓ |
raise exception with byte-string with utf-8 | ✓ | ✓ | ✓ |
raise exception with unicode string | †  | ✓ | † |

† Doesn't break, but the last line of the traceback looks yucky:

```
FAIL: dave.py: AssertionError: öゲ
  File "./stbt-run", line 107, in <module>
    execfile(_filename)
  File "/home/drothlis/work/stb-tester.com/stb-tester-one/stb-tester/stb-tester/dave.py", line 10, in <module>
    assert False, u"öゲ"
AssertionError: \xf6\u30b2
```

drothlis:
+1 to 6cd883e.

wmanley:
Hmm, travis is unhappy:


    test_that_stbt_run_can_print_exceptions_with_unicode_characters... FAIL
    Showing '/tmp/stb-tester.Mmj/log':
    --- /dev/fd/63	2016-01-12 17:21:21.134499583 +0000
    +++ /dev/fd/62	2016-01-12 17:21:21.134499583 +0000
    @@ -1,3 +1,4 @@
    +libdc1394 error: Failed to initialize libdc1394
     FAIL: test.py: AssertionError: ü
         execfile(_filename)
    error: 
    test_that_stbt_run_can_print_exceptions_with_encoded_utf8_string... FAIL
    Showing '/tmp/stb-tester.rSN/log':
    --- /dev/fd/63	2016-01-12 17:21:21.242499583 +0000
    +++ /dev/fd/62	2016-01-12 17:21:21.242499583 +0000
    @@ -1,3 +1,4 @@
    +libdc1394 error: Failed to initialize libdc1394
     FAIL: test.py: AssertionError: ü
         execfile(_filename)
    error: 

but it passes on my machine:

    test_that_stbt_run_can_print_exceptions_with_unicode_characters... OK
    test_that_stbt_run_can_print_exceptions_with_encoded_utf8_string... OK

I'll try with docker.

wmanley:
> I'll try with docker.

Yep, reproduced and fixed.

PR #286, opened 2015-06-11
pythonify stbt-batch

wmanley:
It turned out to be too difficult to reimplement this in one big go so I ended up taking a different approach.  I've split stbt batch run into:

* The scheduler (`stbt-batch.d/run` decides which tests to run and invokes the runner.
* The (private) runner (`stbt-batch.d/run-one`) runs the tests and all the associated pre_run, post_run, classify, recover, etc. scripts.

I've then rewritten the scheduler in Python, leaving the runner in bash.  The process heirarchy will now look like:

         python      ->         bash         -> python
    stbt-batch.d/run -> stbt-batch.d/run-one -> stbt-run

Potentially in the future we can look at pulling the functionality of `run-one` apart and into either the scheduler or `stbt-run` itself.

**TODO:**

- [x] Release note
- [x] Squash


stbt-run
```diff
@@ -25,6 +26,8 @@ parser.add_argument(
     '--save-trace', metavar='FILE', default=None,
     help='Write state to this file, as xz compressed newline-seperated JSON')
 parser.add_argument(
+    '--batch-mode', metavar="RESULTS_ROOT", help="Private: Used by stbt-batch")
```
drothlis: You can use `help=argparse.SUPPRESS`.

stbt-batch.d/run
```diff
@@ -1,315 +1,181 @@
...
+    [['test1.py']]
+    >>> parse_test_args(['test1.py', '--', 'test2.py'])
+    [['test1.py'], ['test2.py']]
+    >>> parse_test_args(['test1.py', '--', 'test2.py', '--'])
+    [['test1.py'], ['test2.py']]
+    >>> parse_test_args(['test1.py', 'test2.py'])
+    >>> parse_test_args(
+    ...     ['test1.py', 'arg1', 'arg2', '--', 'test2.py', 'arg', '--',
+    ...      'test3.py'])
+    [['test1.py', 'arg1', 'arg2'], ['test2.py', 'arg'], ['test3.py']]
```
drothlis: I suspect this doctest isn't being run:

* The output at line 160 should have a space in the filename.
* Line 161 is identical to line 169 but they have different outputs.

stbt-batch.d/run
```diff
@@ -1,315 +1,181 @@
...
-
-# Portable implementation of GNU "readlink -f" to support BSD/OSX.
-realpath() {
-  python -c 'import os, sys; print os.path.realpath(sys.argv[1])' "$1"
-}
 
-main "$@"
+def listsplit(l, v):
+    """
+    A bit like str.split, but for lists
```
drothlis: This could do with a simple doctest.

stbt-batch.d/run
```diff
@@ -1,315 +1,181 @@
...
+    args = parser.parse_args(argv[1:])
+
+    if args.tag is not None:
+        tag = '-' + args.tag
+    else:
+        tag = ""
+
+    os.environ['PYTHONUNBUFFERED'] = 'x'
+
+    term_count = [0]
```
drothlis: You define `term_count` here but you define `stop` below `on_term` (and they're both used in `on_term`).

stbt-batch.d/run
wmanley: Good catch, fixing...

stbt-batch.d/run
wmanley: Done.

stbt-batch.d/run
wmanley: Fixed.  I've removed `stop` from `on_term` because in future commits I will be simplifying the loop and that will let me remove `stop`.

stbt-batch
```diff
@@ -16,11 +16,14 @@
 usage() { grep '^#/' "$0" | cut -c4-; }
 
 [ $# -ge 1 ] || { usage >&2; exit 1; }
+cmd = shift
```
drothlis: This doesn't look like valid shell syntax for assignment. :-)

drothlis:
Don't do any force-pushing just yet, I've got a few fixup commits to push.

wmanley:
Too late :(

drothlis:
Doesn't matter, `git pull --rebase` is magic.

stbt-batch
```diff
@@ -16,11 +16,14 @@
...
 case "$1" in
```
drothlis: Also this will need to be `$cmd` now.

drothlis:
I've finished reviewing. This is very nice, thank you very much for tackling this -- I never should have written it in bash in the first place.

wmanley:
> stbt batch run: Remove workaround for bash arg-parsing

I think I'd prefer to keep the workaround.  The fact that we had to add it shows that something was managing to read from stdin. I'd be concerned that we might make hangs more likely in case a script waits for input on stdin somehow.

drothlis:
> I think I'd prefer to keep the workaround.

Agreed.

PR #287, opened 2015-06-15
stbt batch run: Add --shuffle option

wmanley:
The `--shuffle` option runs the given test cases in a random order.  This can be useful if you have structured your test pack as a large number of short targeted tests.  You can then use:

    stbt batch run --shuffle \
        epg.py::test1 epg.py::test2 menu.py::test1 menu.py::test2 ...

to attempt a random walk of different journeys though your stb UI.  This can be particularly effective at finding hard to trigger bugs and get more value out of the test-cases you have written.

Some tests may take much longer than other tests, which will then use up all your soaking time.  To work around that we measure how long each test takes the first time it is run, and use that as a weighting when choosing the next test to run attempting to equalise the time spent in each test
case.

This makes it reasonable to include both tests that take 10s and tests that take 10min in the same random soak.

This is built on-top of #286.

**TODO:**

- [x] Merge #286
- [x] Add a release note
- [x] Improve documentation?

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -22,6 +22,27 @@ UNRELEASED.
 
 ##### User-visible changes since 22
 
+* `stbt batch run` learned a new option: `--shuffle`. The `--shuffle` option
+  runs the given test cases in a random order.  This can be useful if you have
+  structured your test pack as a large number of short targeted tests.  You can
+  then use:
+
+        stbt batch run --shuffle \
+            epg.py::test1 epg.py::test2 menu.py::test1 menu.py::test2 ...
+
+  to attempt a random walk of different journeys though your set-top-box UI.
+  This can be particularly effective at finding hard to trigger bugs and get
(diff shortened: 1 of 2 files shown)


drothlis:
Very nice. Imagine doing this in bash -- hah!

I don't like the name of the flag -- "soaking" is controlled by the presence or absence of `--run-once`. It should be `--random-order` or just `--random` (I think I prefer the latter).

drothlis:
I really like the tests too (though I haven't tried running them myself).

wmanley:
> I don't like the name of the flag -- "soaking" is controlled by the presence or absence of `--run-once`. It should be `--random-order` or just `--random` (I think I prefer the latter).

Agreed.  Some other ideas naming wise:

    --shuffle
    --sort=random vs. --sort=in-order
    --order=equal-time

LewisHaley:
+1 for `--shuffle`. I think its nicely analogous to `random.shuffle`.

PR #291, opened 2015-06-23
stbt batch run: Fix and test `--shuffle` option

wmanley:
`stbt batch run --shuffle` was failing with

    TypeError: unhashable type: 'list'

This was not caught by the unit-tests because they don't use realistic test functions/arguments.

Before this commit each test case was a Python list which cannot be used as a key in a dictionary.  Now they are tuples so the dictionary of test-case to time-spent is now valid.


drothlis:
Yay for integration tests.

PR #292, opened 2015-07-09
Move man page out of README

drothlis:
**TODO:**

- [ ] Just after merging: update stbt man page link in wiki.
- [ ] Just after merging: update stb-tester.com/stbt.html redirect.


drothlis:
Rebased onto master, improved the README, and removed a bunch of stuff from the man page that was plain wrong or not appropriate for a man page (and it's covered in the wiki already).

PR #293, opened 2015-07-24
stbt lint: Check missing `assert` before `wait_until`

drothlis:
...plus some minor fixes.

Forgetting the `assert` before `wait_until` has bitten me several times already in testcases I've written.

Note that this breaks compatibility with pylint < 1.0 (see the first commit for details). I think it's worth it. On Travis we've been installing a newer pylint from PyPI since d4a450c.

TODO:

- [x] Also check `is_screen_black`, `match`, `match_text`, and `ocr`.
- [x] Also check that the argument to `wait_for_match` is a callable.

### DIFF
--- a/_stbt/pylint_plugin.py
+++ b/_stbt/pylint_plugin.py
@@ -8,7 +8,7 @@ Intended to be used by "stbt lint".
 
 Documentation on Abstract Syntax Tree traversal with python/pylint:
 
-* http://www.logilab.org/card/pylint_manual#writing-your-own-checker
+* http://docs.pylint.org/extend.html#writing-your-own-checker
 * http://hg.logilab.org/review/pylint/file/default/examples/custom.py
 * http://docs.python.org/2/library/ast.html
 
--- a/_stbt/pylint_plugin.py
+++ b/_stbt/pylint_plugin.py
@@ -57,7 +57,10 @@ class StbtChecker(BaseChecker):
         if re.search(r"\b(is_screen_black|match|match_text|ocr|wait_until)$",
                      node.func.as_string()):
             if type(node.parent) == Discard:
-                self.add_message('E7002', node=node, args=node.func.as_string())
+                for inferred in node.func.infer():
+                    if inferred.root().name in ('stbt', '_stbt.core'):
+                        self.add_message(
+                            'E7002', node=node, args=node.func.as_string())
 
         if re.search(r"\bwait_until", node.func.as_string()):
             if node.args:
--- a/_stbt/pylint_plugin.py
+++ b/_stbt/pylint_plugin.py
@@ -1,6 +1,8 @@
 """pylint plugin to do static analysis on stbt scripts
 
 * Identifies broken image links in parameters to `stbt.wait_for_match` etc.
+* Identifies calls to `wait_until` whose return value isn't used (probably
+  missing an `assert`).
 
 Intended to be used by "stbt lint".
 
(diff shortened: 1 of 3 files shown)


drothlis:
List of the packages we'd need to get added, sorted by source package. My
understanding of the [instructions] is that we need to specify the name of
the *binary* package, but we only need to specify one package from any given
source package.

[instructions]: https://github.com/travis-ci/apt-package-whitelist#package-approval-process

```
binary package                    source package in trusty
--------------------------------  ------------------------
python-flask                      flask
gstreamer1.0-libav                gst-libav1.0
gstreamer1.0-plugins-bad          gst-plugins-bad1.0
gstreamer1.0-plugins-base         gst-plugins-base1.0
gstreamer1.0-x                    gst-plugins-base1.0
libgstreamer-plugins-base1.0-dev  gst-plugins-base1.0 
gstreamer1.0-plugins-good         gst-plugins-good1.0
gir1.2-gstreamer-1.0              gstreamer1.0
gstreamer1.0-tools                gstreamer1.0
libgstreamer1.0-dev               gstreamer1.0 
lighttpd                          lighttpd
liborc-0.4-dev                    orc
pep8                              pep8
python-lzma                       python-lzma
gir1.2-gudev-1.0                  systemd
tesseract-ocr                     tesseract
tesseract-ocr-deu                 tesseract-deu
tesseract-ocr-eng                 tesseract-eng
v4l-utils                         v4l-utils
xserver-xorg-video-dummy          xserver-xorg-video-dummy
```

wmanley:
From `make check` output:

    test_that_stbt_lint_ignores_regular_expressions... FAIL
    Showing '/tmp/stb-tester.HQS/log':
    ************* Module test
    E:  2, 0: "re.match" return value not used (missing "assert"?) (stbt-unused-return-value)

wmanley:
> THIS BREAKS COMPATIBILITY WITH PYLINT < 1.0
> ...
> Note that Ubuntu 12.04 ships pylint 0.25. Travis-CI uses Ubuntu 12.04 but in our own travis configuration we install pylint 1.3 from PyPI.

Note that [14.04 ships with pylint 1.1](http://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=pylint&searchon=names) so this should be fine.

wmanley:
> stbt lint: Check missing `assert` before `wait_until` 

Elegant.  I like it.

_stbt/pylint_plugin.py
```diff
@@ -35,8 +29,19 @@ class StbtChecker(BaseChecker):
     msgs = {
         # Range 70xx reserved for custom checkers: www.logilab.org/ticket/68057
         'E7001': ('Image "%s" not found on disk',
+                  'stbt-missing-image',
                   'Used when the image path given to `stbt.wait_for_match` '
                   '(and similar functions) does not exist on disk.'),
+        'E7002': ('"%s" return value not used (missing "assert"?)',
+                  'stbt-unused-return-value',
+                  "When the return value from stbt's "
+                  "wait_until/match/is_screen_black isn't used in an 'if' "
```
wmanley: This list of functions doesn't match up with the ones used in the regex below (missing `match_text` and `ocr`), and I'm not sure of the value of listing every function that we check for here.  Perhaps '"%s" does not throw on failure and the return value is not used. Perhaps you meant "assert %s(...)"'.

wmanley:
> Calling a function *can* be valid, if that function returns another function. Unfortunately pylint's inference doesn't always succeed at figuring out what the function is. If we can't figure it out, we don't complain even if it looks like a function call. As the unit test shows, we don't actually complain about the above example. It will still catch some cases though, so it's still useful.

Hmm. Is there any way for us to annotate our functions at all to tell pylint properties of the return values?

> An alternative approach would be to complain if it's a function call and we fail to infer the return value.

We could then also warn if the function can only return `None`, or possibly always returns a Falsey or Truthy value.

docs/release-notes.md
```diff
@@ -14,6 +14,31 @@ necessary.
...
+
+* `stbt lint` will complain if you don't use the return value from
+  `is_screen_black`, `match`, `match_text`, `ocr`, or `wait_until`. When the
+  return value from `wait_until` isn't used in an `if` statement or assigned to
+  a variable, you've probably forgotten to use `assert`.
+
+* `stbt lint` will complain if its argument isn't callable (such as a function
+  or a lambda expression). This will catch mistakes like
+  `wait_until(is_screen_black())` when you meant to say
+  `wait_until(is_screen_black)`.
```
wmanley: Perhaps you could provide a different example than one that you know doesn't work?

wmanley: Never mind, I hadn't seen the later commit fbb94e0.

wmanley:
The `make check` failure is an intermittent failure related to the old version of ubuntu used on Travis.  Merging

PR #296, opened 2015-08-12
stbt batch run: Avoid directory name clashes

drothlis:
...if the previous test took <1s to run.

You're unlikely to ever see this issue in real usage, but it shows up
from time to time in stb-tester's selftests.


wmanley:
There's no test here, so if we were to re-write `run-one` in Python (a la #286) we would risk regressing.

OTOH I don't think that missing this is too much of a risk, so I'm merging.

PR #298, opened 2015-08-12
New API stbt.Region.replace

drothlis:
This is similar to `Region.extend`, but it takes absolute coordinates
within the image instead of adjusting by a relative number of pixels.

It just delegates to the private `namedtuple._replace` method (`Region`
is a subclass of `namedtuple`). We have found it quite useful in some
of the test cases we write, and this saves us from having to use a
private method.

TODO:

- [x] Release notes.


wmanley:
+1

PR #300, opened 2015-08-17
stbt.ocr and stbt.match_text: Cache the last result from tesseract

drothlis:
Cache the most recent OCR result so that you can call `match_text`
multiple times on the same frame, without invoking a tesseract process
each time. If you change any of the parameters (region, mode, lang, etc)
the cache is invalidated.

TODO:

- [ ] hash the source frame after cropping to the region.
- [ ] release notes.


_stbt/core.py
```diff
@@ -2160,6 +2163,26 @@ def _tesseract_version(output=None):
 
 def _tesseract(frame, region, mode, lang, _config,
                user_patterns=None, user_words=None):
+    """Cache the OCR result of the most recent frame so that you can call
+    `match_text` multiple times on the same frame, cheaply.
+    """
+    global _memoise_tesseract_output
+    h = unicode(hash(
+    (0, '_tesseract', _memoise_tesseract_version,
+     _image_hash(frame), region, mode, lang,
```
wmanley: I think you want to be hashing only after cropping.  This will reduce the amount of work that needs to be done in `_image_hash` and improve the chances of a cache hit.

wmanley: Also this will break if `frame` is a `Gst.Sample`.

drothlis: Good point. Specifically if you have lossless image capture and you're doing `wait_until(lambda: match_text(..., region=...))` where `region` is a UI element that rarely changes, and outside the region you have live TV.

wmanley: Perhaps move all the region handling out of `_tesseract_impl` and into `_tesseract`.

drothlis:
> Cache the most recent OCR result so that you can call match_text
> multiple times on the same frame

If this was the original motivation, such a case won't have any caching since #307, where we changed `match_text` to set tesseract's dictionary to contain the desired words; changing the dictionary changes tesseract's behaviour, so the dictionary (specifically the `user_words` parameter) is part of our cache key.

I'm tempted to close this without merging because it would only be useful in a much narrower set of cases.

drothlis:
I've rebased onto master and re-worked the self-test so that it exercises the desired region behaviour (which I haven't implemented yet, so the self-test fails).

drothlis:
Note also that this introduces some global state (`_memoise_tesseract_output`).

wmanley:
> I'm tempted to close this without merging because it would only be useful in a much narrower set of cases.

Good point.  So to achieve this the user would have to override this by passing `user_words` to each call to `match_text`.

There is still value in this though for accelerating the `wait_until(lambda: ocr() == "something")` use case.  I'm interested in caching for accelerating #343 anyway.

drothlis:
I've closed this, we can look at caching more comprehensively in the context of #343. For OCR specifically, if you want to call `match_text` multiple times against the same frame I feel that a more general solution would be to return a data-structure where you can get at individual lines, words, etc -- whatever granularity is right for your particular use case -- like tesseract's hOCR format but with a nicer Python API around it.

PR #301, opened 2015-08-25
stbt: Teach Region.replace() about width and height

wmanley:
This is useful when you have a function which returns a region of non-fixed size and you just want to take the top of it.  E.g.:

    title_region = find_info_region.replace(height=64)

**TODO:**

- [x] Tests
- ~~Update release notes~~ (I think the existing release note for Region.replace --which is new in the next release-- is sufficient).
- [x] Squash

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -334,7 +334,8 @@ class Region(namedtuple('Region', 'x y right bottom')):
         return Region.from_extents(
             self.x + x, self.y + y, self.right + right, self.bottom + bottom)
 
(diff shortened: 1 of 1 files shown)


drothlis:
I still think that `norm_coords(names, x, width. right, default_x, default_width, default_right)` would be clearer, more consistent & probably less code.

wmanley:
+1

wmanley:
CI failure is unrelated: Ubuntu 12.04 confusion about `Gst.Structure` vs. `Structure`.  Merging

PR #302, opened 2015-08-28
stbt-run: Save last used frame on failure as screenshot.png

wmanley:
stbt-run knows that a test has failed/errored if the test script raises an exception.  When `wait_for_match`, `wait_for_motion`, etc. fail they attach the frame that caused the failure to the exception that they raise.  stbt-run is expecting this so can save the exception for later inspection and debugging.

However: if users are using the more modern style of:

    assert wait_until(lambda: match(...))

then the exception is raised by the assert, which has no knowledge of the frame that caused the assert.  In this case `stbt batch run` will grab a screenshot of whatever happens to be on screen when the test finishes.  This screenshot often doesn't exhibit the issue that caused the test to fail/error in the first place.

This commit makes the assumption that the last frame that the test script actually used is the one that exhibited the issue and saves that instead.

**TODO:**

- [x] Tests
- [x] Release notes

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -927,7 +927,8 @@ def save_frame(image, filename):
     Takes an image obtained from `get_frame` or from the `screenshot`
     property of `MatchTimeout` or `MotionTimeout`.
     """
-    cv2.imwrite(filename, image)
+    with _numpy_from_sample(image, readonly=True) as imagebuf:
+        cv2.imwrite(filename, imagebuf)
 
 
 def wait_until(callable_, timeout_secs=10, interval_secs=0):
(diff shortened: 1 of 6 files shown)


drothlis:
I pushed a minor fixup to "test_that_invalid_control_doesnt_hang: Improve failure message" that fixes a missing quote in the assertion message.

drothlis:
Also pushed a fixup to the selftest.

drothlis:
Wouldn't it be worth removing the screenshot-saving code from `stbt batch run`?

drothlis:
Nice test btw.

drothlis:
For the record: See @LewisHaley's review comments [here](https://github.com/wmanley/stb-tester/commit/2c5646dd3cf7cc6632f7ae419fb09ed2ca5aefe8#commitcomment-12993332) (lost after a rebase -- yay github).

wmanley:
>Wouldn't it be worth removing the screenshot-saving code from stbt batch run?

I considered doing so. `stbt batch run` actually saves a screenshot whether it passed or fails to generate a jpg thumbnail so that approach would require more work than it at first seems. I prefer to defer until more of `stbt batch` is written in python.

drothlis:
Fair enough. I think you can merge this then (after you squash the fixups).

Issue #305, opened 2015-09-29
stb-tester fails with PyGObject 3.13


wmanley:
from @aiman on the stb-tester [mailing list](https://groups.google.com/d/msg/stb-tester/ESy5320lwH4/CyCZI6ygJxwJ):

Our team is currently running stb-tester 22 on Fedora. As Fedora 19 has reached EOL in January and Fedora 20 is scheduled to reach EOL in June we are looking to move to Fedora 21+ at the earliest opportunity. This is because EOL releases receive no security updates or bug fixes.

However we are having some trouble running stb-tester on Fedora 21, which uses gstreamer1-1.4.4 in the base image.

This is a minimal test case:

    stbt screenshot --source-pipeline=videotestsrc --sink_pipeline=fakesink

Running this command results in a series of tracebacks for approximately 30 seconds, followed by a core dump:

    Traceback (most recent call last):
      File "/usr/libexec/stbt/stbt/__init__.py", line 1920, in on_new_sample
        self.push_sample(sample)
      File "/usr/libexec/stbt/stbt/__init__.py", line 1984, in push_sample
        with _numpy_from_sample(sample) as img:
      File "/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
        return self.gen.next()
      File "/usr/libexec/stbt/stbt/__init__.py", line 1708, in _numpy_from_sample
        with map_gst_buffer(sample.get_buffer(), flags) as buf:
      File "/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
        return self.gen.next()
      File "/usr/libexec/stbt/_stbt/gst_hacks.py", line 47, in map_gst_buffer
        "Writable array requested but buffer is not writeable")
    ValueError: Writable array requested but buffer is not writeable

wmanley:
I have seen this on my Debian system.  It was due to pygobject being upgraded rather than GStreamer.  I tracked it down to this bug:
 
https://bugzilla.gnome.org/show_bug.cgi?id=736896
 
although I haven't yet had time to try the solution that Thibault proposed on that ticket.  It would be great if you could give it a go and report back regarding whether his patch works.

PR #306, opened 2015-09-29
Work around regression in PyGObject 3.13

wmanley:
Preventing stb-tester from working on Ubuntu >= 14.04 and Fedora >= 21.

`map_gst_sample` was failing with:

    Traceback(most recent call last):
      File"/usr/libexec/stbt/stbt/__init__.py", line 1920, in on_new_sample
        self.push_sample(sample)
      File"/usr/libexec/stbt/stbt/__init__.py", line 1984, in push_sample
        with _numpy_from_sample(sample)as img:
      File"/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
        returnself.gen.next()
      File"/usr/libexec/stbt/stbt/__init__.py", line 1708, in _numpy_from_sample
        with map_gst_buffer(sample.get_buffer(), flags) as buf:
      File"/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
        return self.gen.next()
      File"/usr/libexec/stbt/_stbt/gst_hacks.py", line 47, in map_gst_buffer
        "Writable array requested but buffer is not writeable")
    ValueError: Writable array requested but buffer is not writeable

PyGObject changed the behaviour of `Gst.Sample.get_buffer` to take a reference to the buffer it returns with `gst_mini_object_ref`.  to prevent use-after-free bugs if the `Gst.Buffer` is used after the `Gst.Sample` goes out of scope.  Unfortunately `gst_mini_object_ref` isn't just a reference-counting increment, it also marks the object as read-only. This makes it impossible to get a writable handle to the `Gst.Buffer` owned by the `Gst.Sample`.

In this commit we work around this by using the C API directly via ctypes as we had to for `map_gst_buffer()` (see f0c5b31).

See the upstream bug ticket: https://bugzilla.gnome.org/show_bug.cgi?id=736896

Fixes: #305

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -35,7 +35,7 @@ from gi.repository import GLib, GObject, Gst  # pylint: disable=E0611
 
 from _stbt import logging, utils
 from _stbt.config import ConfigurationError, get_config
-from _stbt.gst_hacks import gst_iterate, map_gst_buffer
+from _stbt.gst_hacks import gst_iterate, map_gst_sample
 from _stbt.logging import ddebug, debug, warn
 
 gi.require_version("Gst", "1.0")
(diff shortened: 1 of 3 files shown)


drothlis:
Nice! +1 from me.

PR #303, opened 2015-09-07
.travis.yml: Use Travis CI's new docker infrastructure

drothlis:
See http://blog.travis-ci.com/2015-08-19-using-docker-on-travis-ci/

This commit is to test whether we get Ubuntu 14.04 even though we don't actually use docker.

TODO:

- [x] Make CI pass

PR #308, opened 2015-10-06
Display.push_sample: write template name above MatchResult annotation

LewisHaley:
Debugging/Analysing a test failure from the video alone can be difficult
because the template matching regions are not annotated, and cross
referencing with the stderr log is all but necessary in order to identify
what was being searched for at the time.

This commit reduces the cross-referencing required by labelling the
match annotations with the template name being searched for. This helps
to quickly identify what activity is going on at any given time.

Note that, as with the match annotations themselves, this will not affect
`stbt.match` calls which pass in their own frame, unless that frame is
provided by iterating over `stbt._display.gst_samples()` (note also that
`stbt._display.gst_samples().next()` _does not_ work!)

---

I thought this might be of interest. I find it very useful.


drothlis:
Neat. Could you attach a screenshot showing what it looks like?

> Note that, as with the match annotations themselves, this will not affect `stbt.match`

I must say I've been using `match` pretty much exclusively instead of `wait_for_match`, and I do miss the visual feedback. We'll have to do something about that at some stage.

> unless that frame is provided by iterating over `stbt._display.gst_samples()` (note also that `stbt._display.gst_samples().next()` does not work!)

Interesting... I wonder why that is.

_stbt/core.py
```diff
@@ -1524,6 +1524,11 @@ def push_sample(self, sample):
                 color = (int(255 * max([1 - age, 0.5])),) * 3
                 _draw_text(img, x['text'], origin, color)
             for match_result in matches:
+                label = os.path.basename(match_result.image)
```
wmanley: I think this will fail if the image passed to `match` is a `numpy.array` rather than a filename.  `MatchResult`s are constructed as:

    result = MatchResult(..., image=(template.name or template.image))

Perhaps `_AnnotatedTemplate` should always have a non-`None` `.name` or similar?

_stbt/core.py
```diff
@@ -1524,6 +1524,11 @@ def push_sample(self, sample):
...
+                # Slightly above the match annotation
+                label_loc = (match_result.region.x, match_result.region.y - 10)
```
wmanley: What happens here if the match is already at the top of the screen?

wmanley:
This sounds like it could be very useful indeed.  It would be good to have some tests if possible.  Maybe take inspiration from the clock tests and the existing box drawing tests.

LewisHaley:
screenshot from 2015-10-07 11 12 17

_stbt/core.py
LewisHaley: I considered writing logic to check the frame boundaries and to draw the text on the bottom of the annotation if too near the top or aligned to the right of the annotation if too far to the right. I'd basically hacked this into my installed version because I needed help with debugging and didn't need edge-case handling (pun unintended), but I thought it might be useful to quickly put the patch together to push up (thus no tests either) to get feedback.

_stbt/core.py
LewisHaley: Is `friendly_name` the best thing?

LewisHaley:
Note that we've changed `_draw_text` to use a smaller font by default and to anti-alias to improve legibility. The performance impact of the anti-aliasing seemed negligible.

drothlis:
> I thought it might be useful to quickly put the patch together to push up (thus no tests either) to get feedback.

Definitely the right approach. :-)

We do like this and we'll merge once it has tests, plus the other stuff @wmanley said.

LewisHaley:
Fixup pushed (along with a fix for an unrelated failure I saw).

drothlis:
> stbt-batch.d/run: [fix] don't be influenced by $CDPATH

This really belongs in a separate pull request.

drothlis:
> FAIL: test.py: TypeError: 'lineType' is an invalid keyword argument for this function

Looks like CI failed because the version of OpenCV on Travis's Ubuntu 12.04 doesn't know about the lineType argument?

If you rebase this branch onto stb-tester master, you'll get the upgrade-travis-to-ubuntu-14.04 branch that I've just merged (#303), which would prove or disprove my theory.

drothlis:
It also looks like the new template images for the self-tests are missing.

drothlis:
> FAIL: test.py: TypeError: 'lineType' is an invalid keyword argument for this function

In OpenCV 2.4.0 they changed  the `cv2.putText` signature from `linetype` to `lineType` (i.e. they made it camel-case). How annoying.

Ubuntu 12.04 has OpenCV 2.3.1; Ubuntu 14.04 has OpenCV 2.4.8.

@wmanley what's our policy for supporting Ubuntu 12.04? Especially now that we're not testing it (since we merged #303 earlier today). One option would be not to build 12.04 packages for new stb-tester releases, so that people can still install v23 from the PPA?

wmanley:
> @wmanley what's our policy for supporting Ubuntu 12.04? Especially now that we're not testing it (since we merged #303 earlier today).

I believe the minimum version of Ubuntu is 14.04, we were only keeping (partial, broken) 12.04 support around for Travis.  Now that #303 has been merged I don't think there's any value in maintaining (the illusion of) 12.04 support, nor would it be practical.

It would probably be a good time to remove any [12.04 hacks](https://github.com/stb-tester/stb-tester/search?q="12.04"&type=Code) too.

> One option would be not to build 12.04 packages for new stb-tester releases, so that people can still install v23 from the PPA?

The lowest version of stb-tester that we've built packages for is 13.10: https://launchpad.net/~stb-tester/+archive/ubuntu/stb-tester

wmanley:
@LewisHaley: Could you rebase and force push?

LewisHaley:
pushed.

drothlis:
Merged. Thanks @LewisHaley!

drothlis:
> It would probably be a good time to remove any [12.04 hacks](https://github.com/stb-tester/stb-tester/search?q="12.04"&type=Code) too.

Done in #314.

PR #309, opened 2015-10-08
stbt-batch.d/run: [fix] don't be influenced by $CDPATH

LewisHaley:
I encountered a failure in
`test_that_stbt_batch_run_will_run_a_specific_function`. It became apparent
that the output directory called "results" was not being created and
instead a directory of my own called "results" which exists on my $CDPATH
was being used. This means that `subprocess.Popen`, is using the
`outputdir` in the environment dictionary by effectively doing
`cd env['outputdir']`.

This makes the behaviour unpredictable, and so this commit fixes it by
setting `outputdir` to be relative to the working dir unless it is already
an absolute path. This means that the path is not up for interpretation.

### DIFF
--- a/stbt-batch.d/run.py
+++ b/stbt-batch.d/run.py
@@ -101,7 +101,7 @@ def main(argv):
         subenv['tag'] = tag
         subenv['v'] = '-vv' if args.debug else '-v'
         subenv['verbose'] = str(args.verbose)
-        subenv['outputdir'] = args.output
+        subenv['outputdir'] = os.path.abspath(args.output)
         child = None
         try:
             child = subprocess.Popen(


stbt-batch.d/run.py
```diff
@@ -101,7 +101,8 @@ def on_term(_signo, _frame):
         subenv['tag'] = tag
         subenv['v'] = '-vv' if args.debug else '-v'
         subenv['verbose'] = str(args.verbose)
-        subenv['outputdir'] = args.output
+        subenv['outputdir'] = args.output if os.path.isabs(args.output) \
+            else os.path.join(os.path.curdir, args.output)
```
wmanley: Using `os.path.abspath(args.output)` directly here could be cleaner.

wmanley:
Huh, I'd never heard of `$CDPATH` before, thanks for the education :).

I wonder if there's a more general environment sanitization we should be doing here?  Or maybe we should be recommending using `stbt docker` (#304)?

drothlis:
This is why you should never export CDPATH.

That is, your .bashrc should say:

    CDPATH=:/path/to/a:/path/to/b

and not

    export CDPATH=:/path/to/a:/path/to/b

CDPATH is intended for interactive use. By exporting it, you're going to break a lot of unsuspecting bash scripts.

That said, I'll be happy to merge this (once you address @wmanley's abspath comment) because it's a bash implementation detail leaking out and it shouldn't cause `stbt batch run` to break.

This is yet more proof that I should've written `stbt batch run` in Python in the first place. :-)

LewisHaley:
change pushed.

drothlis:
Merged. Thanks @LewisHaley!

Issue #297, opened 2015-08-12
Librification


wmanley:
It would be good if stbt could be used as a well behaved library, rather than requiring `stbt run` to be used if you want to use the library.  This would aid integration with other testing frameworks such as py.test.

By well-behaved I mean:
1. Has no global state
2. Can be cleanly initialised and completely deinitialised
3. Configuration can be injected rather than is loaded from disk implicitly

TODO (probably incomplete):
- [x] Create a `DeviceUnderTest` class refactoring global variables into it #285 
- [x] Refactor `Display` to separate the sink-pipeline into a seperate class, allowing the sink-pipeline and the `DeviceUnderTest` objects to have independent lifetimes #420 
  - This would be needed for py.test support as I would think the `DeviceUnderTest` class would be a (potentially long-lived) pytest fixture, but the output video would probably be implemented using some kind of [runtest hook](https://pytest.org/latest/plugins.html#generic-runtest-hooks)
- [ ] Inject all configuration into the `DeviceUnderTest` class, removing any calls to `get_config` outside of a factory function? #333
- [x] Remove `stbt.init_run` and `stbt.teardown_run`.
  - `DeviceUnderTest` is already a `contextmanager`.
- [x] Create a `contextmanager` or similar to take care of most of the concerns of `stbt run` and refactor `stbt run` to use it.

allanlewis:
Thanks @wmanley  - this will be useful for tracking progress towards librarification (is that a word?) of `stb-tester` :+1:

drothlis:
@martynjarvis @LewisHaley @jcutbush @johnteslade @allanlewis have you guys done any work on the above?

LewisHaley:
None at all.

wmanley:
I'm not planning on doing any more work on this for the time being, closing.

drothlis:
It's pretty much done already, at least for the use case of being able to `import stbt` for linting, IDEs, etc.

PR #311, opened 2015-10-13
Mainloop handling refactoring

wmanley:
Some refactorings I made while working on dual camera support so we can have two `DeviceUnderTest` objects at the same time.

This should help towards librification (#297) too.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1641,6 +1641,7 @@ class Display(object):
                     debug("Timeout waiting for sink EOS")
             else:
                 debug("Sending EOS to sink pipeline failed")
+            self.sink_pipeline.set_state(Gst.State.NULL)
 
         self.mainloop.__exit__(None, None, None)
 
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1289,6 +1289,7 @@ class Display(object):
         self.start_timestamp = None
         self.underrun_timeout = None
         self.tearing_down = False
+        self.received_eos = threading.Event()
 
         self.annotations_lock = threading.Lock()
         self.text_annotations = []
@@ -1538,7 +1539,6 @@ class Display(object):
         err, dbg = message.parse_error()
         self.tell_user_thread(
             UITestError("%s: %s\n%s\n" % (err, err.message, dbg)))
-        _mainloop.quit()
 
     def on_warning(self, _bus, message):
         assert message.type == Gst.MessageType.WARNING
@@ -1554,7 +1554,7 @@ class Display(object):
 
     def on_eos_from_sink_pipeline(self, _bus, _message):
         debug("Got EOS")
-        _mainloop.quit()
+        self.received_eos.set()
 
     def on_underrun(self, _element):
         if self.underrun_timeout:
@@ -1606,6 +1606,7 @@ class Display(object):
     def startup(self):
         self.set_source_pipeline_playing()
         self.sink_pipeline.set_state(Gst.State.PLAYING)
+        self.received_eos.clear()
 
         self.mainloop_thread.start()
 
@@ -1623,6 +1624,9 @@ class Display(object):
         if not self.novideo:
             debug("teardown: Sending eos")
             self.appsrc.emit("end-of-stream")
+            if not self.received_eos.wait(10):
+                debug("Timeout waiting for sink EOS")
+            _mainloop.quit()
             self.mainloop_thread.join(10)
             debug("teardown: Exiting (GLib mainloop %s)" % (
                 "is still alive!" if self.mainloop_thread.isAlive() else "ok"))
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1635,14 +1635,14 @@ class Display(object):
                 debug("teardown: Source pipeline did not teardown gracefully")
             source.set_state(Gst.State.NULL)
             source = None
-        if not self.novideo:
-            debug("teardown: Sending eos")
-            if self.appsrc.emit("end-of-stream") == Gst.FlowReturn.OK:
-                if not self.received_eos.wait(10):
-                    debug("Timeout waiting for sink EOS")
-            else:
-                debug("Sending EOS to sink pipeline failed")
-            self.sink_pipeline.set_state(Gst.State.NULL)
+
+        debug("teardown: Sending eos")
+        if self.appsrc.emit("end-of-stream") == Gst.FlowReturn.OK:
+            if not self.received_eos.wait(10):
+                debug("Timeout waiting for sink EOS")
+        else:
+            debug("Sending EOS to sink pipeline failed")
+        self.sink_pipeline.set_state(Gst.State.NULL)
 
         self.mainloop.__exit__(None, None, None)
 
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1491,13 +1491,13 @@ class Display(object):
 
     def draw(self, obj, duration_secs):
         with self.annotations_lock:
-            if type(obj) in (str, unicode):
+            if isinstance(obj, str) or isinstance(obj, unicode):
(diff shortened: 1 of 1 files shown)


_stbt/core.py
```diff
@@ -1284,7 +1299,7 @@ def _get_frame_timestamp(frame):
 
 class Display(object):
     def __init__(self, user_source_pipeline, user_sink_pipeline,
-                 save_video,
+                 mainloop, save_video="",
```
drothlis: Why did you need to give `save_video` a default value?

_stbt/core.py
```diff
@@ -1531,12 +1545,13 @@ def push_sample(self, sample):
 
     def on_error(self, pipeline, _bus, message):
         assert message.type == Gst.MessageType.ERROR
-        Gst.debug_bin_to_dot_file_with_ts(
-            pipeline, Gst.DebugGraphDetails.ALL, "ERROR")
+        if pipeline is not None:
+            # If we receive an error after the pipeline has been torn down:
+            Gst.debug_bin_to_dot_file_with_ts(
+                pipeline, Gst.DebugGraphDetails.ALL, "ERROR")
```
drothlis: I'm a bit confused by the comment. It seems to suggest that you'll run the code inside the `if` block if we receive an error after the pipeline has been torn down.

In fact you've introduced the `if` so that `debug_bin_to_dot_file_with_ts` doesn't get run at all if we get an error after the pipeline has been torn down?

Seems clearer to just remove the comment altogether.

_stbt/core.py
```diff
@@ -1171,14 +1171,29 @@ def argparser():
...
 # ===========================================================================
 
-if hasattr(GLib.MainLoop, 'new'):
-    _mainloop = GLib.MainLoop.new(context=None, is_running=False)
-else:
-    # Ubuntu 12.04 (Travis) support: PyGObject <3.7.2 doesn't expose the "new"
-    # constructor we'd like to be using, so fall back to __init__.  This means
-    # Ctrl-C is broken on 12.04 and threading will behave differently on Travis
-    # than on our supported systems.
-    _mainloop = GLib.MainLoop()
```
drothlis: This is going to conflict with #314 which we merged earlier today.

_stbt/core.py
```diff
@@ -1531,12 +1545,13 @@ def push_sample(self, sample):
...
-        Gst.debug_bin_to_dot_file_with_ts(
-            pipeline, Gst.DebugGraphDetails.ALL, "ERROR")
+        if pipeline is not None:
+            # If we receive an error after the pipeline has been torn down:
+            Gst.debug_bin_to_dot_file_with_ts(
+                pipeline, Gst.DebugGraphDetails.ALL, "ERROR")
         err, dbg = message.parse_error()
         self.tell_user_thread(
             UITestError("%s: %s\n%s\n" % (err, err.message, dbg)))
-        _mainloop.quit()
```
drothlis: So now on GStreamer errors we try to keep going? Or teardown takes 10s longer?

drothlis: Hmmm, I suppose if we had previously received an error then when we emit end-of-stream during teardown we won't take the FlowReturn.OK branch and therefore teardown won't actually take 10s?

_stbt/core.py
```diff
@@ -1620,10 +1636,14 @@ def teardown(self):
...
-            self.appsrc.emit("end-of-stream")
-            self.mainloop_thread.join(10)
-            debug("teardown: Exiting (GLib mainloop %s)" % (
-                "is still alive!" if self.mainloop_thread.isAlive() else "ok"))
+            if self.appsrc.emit("end-of-stream") == Gst.FlowReturn.OK:
+                if not self.received_eos.wait(10):
+                    debug("Timeout waiting for sink EOS")
+            else:
+                debug("Sending EOS to sink pipeline failed")
+            self.sink_pipeline.set_state(Gst.State.NULL)
```
drothlis: Good catch with the set_state(NULL). I wonder if this will fix some very intermittent segfaults I've seen.

When we did a similar fix for the _source_ pipeline (5d258bf) it fixed some intermittent segfaults during teardown.

drothlis:
I've finished reviewing. Good stuff.

_stbt/core.py
wmanley: Not sure.  An error doesn't necessarily mean that data will stop flowing, but that's the convention.  I'm not quite sure what the `_mainloop.quit()` was achieving in here.  I guess it would avoid the possibility of a 10s wait at teardown.

I guess the sequence of events is:

1. Error occurs
2. User thread tries to pull a frame
3. Exception is raised in user thread
4. `self.teardown` is called

And then before this change:

5. Emit end of stream
6. Destroy pipeline

After this change:

5. Emit (and check) end of stream
6. Wait for EOS for up to 10s.
7. Set state NULL
8. Destroy pipeline.

I think the new implementation is probably an improvement.  We'd have to test with some real instances of ERRORs to be sure.

_stbt/core.py
```diff
@@ -1622,10 +1637,14 @@ def teardown(self):
             source = None
         if not self.novideo:
```
wmanley: Hmm.  We are only setting state to NULL if we haven't seen `NoVideo`.  This doesn't seem appropriate.

wmanley: Why `NoVideo` which is caused by the source pipeline should have an effect on the `sink_pipeline` I don't know

wmanley: This is why (from an excellent commit message by @drothlis) (aee8f04d97bc3b0b03894a588debbbb6fa45cc4d 2013-07-02):

> stbt.Display.teardown doesn't hang when No Video
> 
> When NoVideo was raised at the very beginning of the test, "teardown" completed without error but the EOS event was never received from the pipeline, so "on_eos" was never called, so the application didn't exit because the GLib main loop's thread was a non-daemon thread.[1]
> 
> [1] http://docs.python.org/2/library/threading.html#thread-objects
> 
> When NoVideo was raised halfway through a test (video was initially available but was then lost), "teardown" raised the following error (presumably because "self.source_bin" was "None" because we were halfway through a source restart) and the application also never exited:
> 
>     stbt-run: teardown: Sending eos
>     Traceback (most recent call last):
>       File "./stbt-run", line 50, in <module>
>         stbt.teardown_run()
>       File "/home/stb-tester/stb-tester/stbt.py", line 648, in teardown_run
>         _display.teardown()
>       File "/home/stb-tester/stb-tester/stbt.py", line 858, in teardown
>         gst.element_unlink_many(self.source_bin, self.sink_bin)
>     TypeError: argument must be a GstElement
> 
> This commit fixes both cases, by remembering if NoVideo was raised. We mark the GLib main loop's thread as a daemon thread, and (if there *is* video) we explicitly wait for the thread to complete (with a timeout just in case anything unforeseen goes wrong).

wmanley: So that was back when we only had a single pipeline and I don't think it applies any more, in particular since after this PR we are no longer depending on EOS to exit the mainloop.

_stbt/core.py
wmanley: Probably just a leftover from another change.

_stbt/core.py
wmanley: I think this is an improvement as even if the sink pipeline has had an error EOS may still work, and we can't know without trying.  And if there is a proper error then emitting the EOS event may fail so we should avoid the wait.  Either way we want to go to state NULL.

wmanley:
I've rebased on master and pushed an additional commit removing the `novideo` conditional. I think I'm happy enough for this to go in.  Receiving ERRORs from the source pipeline is rare anyway as it doesn't depend on video capture hardware so I doubt it will cause any regressions and it should cause some rare teardown problems to go away.

drothlis:
+1

PR #315, opened 2015-10-15
stbt.match: Fix error message for greyscale images

### LINKED ISSUE #310: Bad error message when matching non-BGR images
```python
def _match(image, template, match_parameters, template_name):
    if any(image.shape[x] < template.shape[x] for x in (0, 1)):
        raise ValueError("Source image must be larger than template image")
    if any(template.shape[x] < 1 for x in (0, 1)):
        raise ValueError("Template image must contain some data")
    if template.shape[2] != 3:
        raise ValueError("Template image must be 3 channel BGR")
```

The final `if` causes an `IndexError` with single-channel images because the image `shape` is `(y, x)` not the assumed `(y, x, c)`. The result is that the purposefully placed error message about the required image types does not get raised.

drothlis:
This was raising "IndexError: tuple index out of range" instead of the
desired "ValueError: Template image must be 3 channel BGR".

Fixes #310.

### DIFF
--- a/tests/test_match.py
+++ b/tests/test_match.py
@@ -1,6 +1,9 @@
+import cv2
 import numpy
+from nose.tools import raises
 
 import stbt
+from _stbt.core import _load_template
 
 
 def black(width=1280, height=720):
@@ -18,3 +21,9 @@ def test_that_matchresult_str_image_matches_template_passed_to_match():
 def test_that_matchresult_str_image_matches_template_passed_to_match_custom():
     assert "image=<Custom Image>" in str(
         stbt.match(black(30, 30), frame=black()))
+
+
+@raises(ValueError)
+def test_that_match_rejects_greyscale_template():
+    grey = cv2.cvtColor(_load_template("black.png").image, cv2.cv.CV_BGR2GRAY)
+    stbt.match(grey, frame=black())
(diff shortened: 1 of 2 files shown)


drothlis:
CI failed because Travis no longer configures `apt-get` to run non-interactively. I'm fixing that in #316.

drothlis:
Rebased onto master to get #316.

tests/test_match.py
```diff
@@ -18,3 +21,9 @@ def test_that_matchresult_str_image_matches_template_passed_to_match():
 def test_that_matchresult_str_image_matches_template_passed_to_match_custom():
     assert "image=<Custom Image>" in str(
         stbt.match(black(30, 30), frame=black()))
+
+
+@raises(ValueError)
+def test_that_match_rejects_greyscale_template():
+    grey = cv2.cvtColor(_load_template("black.png").image, cv2.cv.CV_BGR2GRAY)
```
wmanley: What is `_load_template` buying you here that `cv2.imread` isn't?  Path lookup?

drothlis: Yeah -- it saves `os.path.join(os.path.dirname(__FILE__), ...)`.

wmanley:
+1

PR #318, opened 2015-11-09
Add stbt virtual-stb

wmanley:
useful for testing against STB simulators

For instance to test an STB UI written for HbbTV or HTML5 you can use this with chrome.

This PR replaces #174 but is simpler and narrower in scope.  For one we don't worry about docker.  If you want to use docker you can combine this with #304, but you'll have to run the virtual-stb and the tests in the same container.

TODO:

- [x] Clean-up and fix races in the Xorg startup detection using [SIGUSR1](http://www.x.org/archive/X11R7.5/doc/man/man1/Xserver.1.html#toc10)
- [x] Add dependencies to Fedora & Ubuntu packages
- [x] Write tests
- [x] Write release note
- [x] Documentation
- ~~Add `--driver=x` or similar for forward compatibility so we can add `--driver=wayland` or `--driver=spice`?~~ No, we can default to X and worry about this later once these technologies exist.


drothlis:
> Add `--driver=x` or similar for forward compatibility so we can add `--driver=wayland` or `--driver=spice`?

Surely we could add that later if we need it -- as long as it defaults to `x` there wouldn't be any issue with compatibility?

drothlis:
We need to add the dependencies to the Fedora & Ubuntu packages (I've added a TODO item).

On Ubuntu the dependencies are at least:

* python-jinja2
* ratpoison
* xdotool
* xserver-xorg-input-void
* xserver-xorg-video-dummy

drothlis:
`virtual-stb` isn't listed in `stbt --help`.

drothlis:
Ditto for the manpage, although we could do with writing manpages for each sub-command (in a separate pull request).

wmanley:
@drothlis: Please review

Makefile
```diff
@@ -73,55 +63,81 @@ defaults.conf: stbt.conf .stbt-prefix
...
+    stbt-batch.d/templates/index.html \
+    stbt-batch.d/templates/testrun.html \
+    stbt-config \
+    stbt-control \
+    stbt-lint \
+    stbt-power \
+    stbt-screenshot \
+    stbt-templatematch \
+    stbt-tv \
+    stbt/__init__.py
```
drothlis: Would be worth sorting these.

Makefile
```diff
@@ -73,55 +63,81 @@ defaults.conf: stbt.conf .stbt-prefix
...
 	    $(DESTDIR)$(sysconfdir)/bash_completion.d/stbt
+	for filename in $(INSTALL_CORE_FILES); \
+	do \
+	    if [ -x "$$filename" ]; then \
+	        mode=0755; \
+	    else \
+	        mode=0644; \
+	    fi; \
+	    $(INSTALL) -m $$mode $$filename $(DESTDIR)$(libexecdir)/stbt/$$filename; \
+	done
```
drothlis: This (and the rule below) is quite verbose vertically, the mode calculation doesn't deserve quite so much prominence. How about:

```
for filename in $(INSTALL_CORE_FILES); do \
    [[ -x "$$filename" ]] && mode=0755 || mode=0644; \
    $(INSTALL) -m $$mode $$filename $(DESTDIR)$(libexecdir)/stbt/$$filename; \
done
```

extra/debian/control
```diff
@@ -59,7 +59,6 @@ Depends: ${shlibs:Depends},
          gstreamer1.0-plugins-base,
          gstreamer1.0-plugins-good,
          gstreamer1.0-tools,
-         gstreamer1.0-x,
```
drothlis: I don't agree with removing `xvimagesink` from the default stb-tester installation, considering how many people start by trying it out on their desktop. I see how it's unnecessary on a headless test rig, but we should explore other solutions to that (perhaps splitting out an `stb-tester-core` package with the minimal dependencies) in a different pull request.

README.rst
```diff
@@ -109,13 +109,26 @@ Global options
     `--source-pipeline=videotestsrc`. A script like `press("18")` will change
     videotestsrc's pattern property (see `gst-inspect videotestsrc`).
 
-  x11:<display>
+  x11:<display>,<key_mapping_file>
```
drothlis: Is this a comma because `display` can contain colons? Does display have to have the colon like `:0` or does it have to be without the colon like `0` or can it be either?

If `key_mapping_file` is optional it should read `[,<key_mapping_file>]`. Same goes for `display` I suppose.

README.rst
```diff
@@ -109,13 +109,26 @@ Global options
...
+
+        # This is a comment
+
+        KEY_FASTFORWARD   parenright
+        KEY_REWIND        parenleft
+
+    The column on the left is the key name you'll be using in your test-cases,
+    the column on the right is the X keysym that that key will be translated to.
+    For a full list of X keysyms see
+    http://www.cl.cam.ac.uk/~mgk25/ucs/keysyms.txt .
```
drothlis: Add:

> If a key_mapping_file isn't provided, stbt will translate KEY_A - KEY_Z, KEY_0 - KEY_9, KEY_SPACE, KEY_UP, KEY_DOWN, KEY_LEFT, KEY_RIGHT and KEY_OK to the appropriate X keysyms.

drothlis: Or maybe something like:

> If a key_mapping_file isn't provided, stbt provides some sensible mappings when there is an obvious match for our standard key names https://stb-tester.com/stb-tester-one/rev2015.1/getting-started#remote-control-key-names

docs/release-notes.md
```diff
@@ -30,6 +42,11 @@ UNRELEASED
 [standard key names]: https://stb-tester.com/stb-tester-one/rev2015.1/getting-started#remote-control-key-names
 [Roku key names]: https://sdkdocs.roku.com/display/sdkdoc/External+Control+Guide#ExternalControlGuide-3.4ValidKeys
 
+* The `x11` remote control now converts stb-tester's [standard key names] (like
+  "KEY_HOME") to the key names that `xdotool` expects.  You can also now
```
drothlis: > like ("KEY_HOME")

...but KEY_HOME isn't in the default mapping above.

_stbt/control.py
```diff
@@ -715,8 +715,10 @@ def test_x11_remote():
 
         # Can't be sure how long xterm will take to get ready:
         for _ in range(0, 20):
-            for keysym in ['t', 'o', 'u', 'c', 'h', 'space', 'g', 'o', 'o',
-                           'd', 'Return']:
+            for keysym in ['KEY_T', 'KEY_O', 'KEY_U', 'KEY_C', 'KEY_H',
+                           'KEY_SPACE',
+                           'g', 'o', 'o', 'd',
+                           'KEY_OK']:
```
drothlis: We could also do with a test for providing a custom mapping file.

drothlis: Nevermind, that's covered by the virtual-stb tests in the next commit.

extra/debian/control
wmanley: Good point about xvimagesink, it hadn't occurred to me.  We need to keep this considering our default sink is xvimagesink.

README.rst
wmanley: Yes, display will typically be of the form `:0`, so that's why a comma is used.  display is also optional so I'll update that too.

README.rst
wmanley: Sensible mappings are provided even if a key_mapping_file is provided, so I'll have to tweak that wording somewhat.

drothlis:
The chromium logs still go to the console, even with `--background` and without `-v`. Is that expected?

wmanley:
> The chromium logs still go to the console, even with --background and without -v. Is that expected?

Yes, I don't want to do output redirection in `virtual-stb run` for the same reason that the option is `--background` rather than `--daemonize`: it's not intended to be run like a daemon.  If someone wants to send these logs to a file they can use `>` redirection.

drothlis:
Pushed a small fixup to simplify the css styles in the vstb-example-html5.

wmanley:
Thanks for your detailed review.  I've squashed all the fixups and will merge once CI passes.

SatadruBhattacharjee:
Even It's showing the virtual-stb command in "stbt -h". on executing the stbt virtual-stb command , it's throwing _"stbt virtual-stb is not installed"._

My stbt version is **27-1-g2b61c3a**

drothlis:
@SatadruBhattacharjee See the [v25 release notes](https://github.com/stb-tester/stb-tester/blob/master/docs/release-notes.md#25) for installation instructions.

SatadruBhattacharjee:
Thanks a lot for your quick response @drothlis , working like a charm 👍

SatadruBhattacharjee:
Chromium is crashing very frequently with stbt virtual-stb command `stbt virtual-stb run --background chromium-browser --no-sandbox --app=file:///home/vagrant/stb-tester/tests/vstb-example-html5/virtual-stb/index.html
`
Chromium browser version is 55.0
Any clue ?

drothlis:
@SatadruBhattacharjee maybe take this to the mailing list: http://groups.google.com/group/stb-tester

PR #319, opened 2015-12-01
core: Support functools.partial objects in _callable_description

martynjarvis:
`functools.partial` creates `partial` objects that behave like a function.
However, unlike a function, they do not have a `__name__` attribute created
automatically[1]. This results in `_callable_description` raising an
`AttributeError` if `callable_` is a partial object.

This commit adds support for `functools.partial` objects in
`_callable_description`.

[1] https://docs.python.org/2/library/functools.html#partial-objects

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1003,8 +1003,17 @@ def _callable_description(callable_):
     >>> _callable_description(
     ...     lambda: stbt.press("OK"))
     '    lambda: stbt.press("OK"))\\n'
+    >>> _callable_description(functools.partial(int, base=2))
+    'int'
     """
-    d = callable_.__name__
+    try:
+        d = callable_.__name__
+    except AttributeError:
+        if isinstance(callable_, functools.partial):
+            # functools.partial wraps the original function
+            d = callable_.func.__name__
+        else:
+            raise
     if d == "<lambda>":
         try:
             d = inspect.getsource(callable_)


_stbt/core.py
```diff
@@ -1003,8 +1003,17 @@ def _callable_description(callable_):
...
+    >>> _callable_description(functools.partial(int, base=2))
+    'int'
     """
-    d = callable_.__name__
+    try:
+        d = callable_.__name__
+    except AttributeError:
+        if isinstance(callable_, functools.partial):
+            # functools.partial wraps the original function
+            d = callable_.func.__name__
```
allanlewis: Perhaps it would be worth including the baked-in arguments? Something like:
```py
name = callable_.func.__name__
args = ', '.join(str(arg) for arg in callable_.func.args)
kwargs = ', '.join('{}={!r}'.format(arg, value) for arg, value in callable_.func.keywords.iteritems())
d = '{}({}, {})'.format(name, args, kwargs)
```
(That doesn't handle the case where `args` or `kwargs` is empty, but something like that...)

drothlis:
Nice catch. We can always implement @allanlewis's suggestion in the future, but as it is this pull request is a valuable improvement over the current behaviour.

drothlis:
Hmmm I just realised that you can have a `partial` of a `partial`. @martynjarvis & @allanlewis would you mind doing a quick review of #320?

PR #320, opened 2015-12-01
Support recursive functools.partial objects in _callable_description

drothlis:
This extends the fix from #319  so that it also works with nested `functools.partial` objects.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1008,16 +1008,23 @@ def _callable_description(callable_):
     >>> _callable_description(functools.partial(functools.partial(int, base=2),
     ...                                         x='10'))
     'int'
+    >>> class T(object):
+    ...     def __call__(self): return True;
+    >>> _callable_description(T())
+    '<_stbt.core.T object at 0x...>'
     """
-    if isinstance(callable_, functools.partial):
+    if hasattr(callable_, "__name__"):
+        name = callable_.__name__
+        if name == "<lambda>":
+            try:
+                name = inspect.getsource(callable_)
+            except IOError:
+                pass
+        return name
+    elif isinstance(callable_, functools.partial):
         return _callable_description(callable_.func)
-    d = callable_.__name__
-    if d == "<lambda>":
-        try:
-            d = inspect.getsource(callable_)
-        except IOError:
-            pass
-    return d
+    else:
+        return repr(callable_)
 
 
 @contextmanager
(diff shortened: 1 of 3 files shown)


martynjarvis:
That makes sense :+1:

_stbt/core.py
```diff
@@ -1003,17 +1003,13 @@ def _callable_description(callable_):
...
-        d = callable_.__name__
-    except AttributeError:
-        if isinstance(callable_, functools.partial):
-            # functools.partial wraps the original function
-            d = callable_.func.__name__
-        else:
-            raise
+    if isinstance(callable_, functools.partial):
+        return _callable_description(callable_.func)
+    d = callable_.__name__
```
wmanley: I think we want to keep the try block too.  I just tried this:

```
In [1]: class abc(object):
   ...:     def __call__(self):
   ...:         return True
   ...:     

In [5]: func = abc()

In [6]: func()
Out[6]: True

In [7]: func.__name__
AttributeError                            Traceback (most recent call last)
<ipython-input-7-937dc720289a> in <module>()
----> 1 func.__name__

AttributeError: 'abc' object has no attribute '__name__'
```

I don't think we want this function to raise an `AttributeError` under any circumstances.  So perhaps:

    if isinstance(callable_, functools.partial):
        return _callable_description(callable_.func)
    try:
        d = callable_.__name__
    except AttributeError:
        d = repr(callable)

docs/release-notes.md
```diff
@@ -59,6 +59,9 @@ UNRELEASED
 * Fixed "stb-tester fails with PyGObject 3.13" #305.  This will allow
   stb-tester to work out of the box on Ubuntu >= 14.10 and Fedora >= 21.
 
+* `wait_until` can now take a `functools.partial` object. Thanks to Martyn
+  Jarvis for the patch.
```
wmanley: One nitpick, `wait_until` could already take partials, as long as the condition waited for would occur during the timeout.

drothlis:
I've addressed the above comments, squashed & rebased onto master.

wmanley:
+1 merging

_stbt/core.py
```diff
@@ -1005,21 +1005,26 @@ def _callable_description(callable_):
...
-        else:
-            raise
-    if d == "<lambda>":
-        try:
-            d = inspect.getsource(callable_)
-        except IOError:
-            pass
-    return d
+    if hasattr(callable_, "__name__"):
+        name = callable_.__name__
```
wmanley: Not important but I tend to prefer doing something and then checking if it succeeded to checking whether it will succeed before doing it.  It's not important in this instance but it can be good for avoiding race-conditions.

If you wanted to avoid `try...except` `getattr` could be used:

    name = getattr(callable_, __name__, None)

PR #321, opened 2015-12-07
stbt batch run: Store full test-pack sha in `git-commit-sha` artifact

drothlis:
In the results view you can search by test-pack sha, so this will allow
you to search by the complete sha.

`--abbrev=40` shows the full 40-character sha instead of abbreviating it
to the shortest unambiguous prefix (unambiguous at the time that `git
describe` was run; as more commits are added it might have become
ambiguous).

`--long` shows the sha even if the head exactly matches a tag.

I considered changing this to `git rev-parse HEAD` but then we'd lose
the "dirty" information.

### DIFF
--- a/stbt-batch.d/run-one
+++ b/stbt-batch.d/run-one
@@ -37,6 +37,7 @@ main() {
 
   test="$1" && shift &&
   testpath=$(realpath "$test") &&
+  test_script_dir=$(dirname "$testpath") &&
   mkdir -p "$outputdir" &&
   pushd "$outputdir" >/dev/null &&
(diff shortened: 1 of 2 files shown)


wmanley:
I think that we should leave git-commit alone and introduce a new file called `git-commit-sha` instead as both pieces of information could be useful.

drothlis:
+1 to @wmanley's changes (988871d).

drothlis:
Travis is having trouble according to https://www.traviscistatus.com so we'll need to give it a few hours.

PR #322, opened 2015-12-16
stbt.wait_for_match: Take `region` parameter

drothlis:
...much like `stbt.match` already does.

TODO:

- [x] Tests.
- [x] Release notes.


wmanley:
+1 Merging

PR #324, opened 2016-01-11
Debian packaging fixes

drothlis:
While trying to publish ppa packages for Ubuntu 15.04 and 15.10, I've run across an issue that caused `extra/debian/build-source-package.sh` to fail: Somewhere between Ubuntu 14.04 and 15.04, `dpkg` defaults to creating `.tar.xz` files instead of `.tar.gz` files, but our script expected the latter.

This pull request adds a script to build the packages in a docker container (Ubuntu 14.04) to avoid such incompatibilities. Similar to what we already have for building Fedora packages.

TODO:

- [x] Get the new packages actually published (dput says "Successfully uploaded packages" but I've yet to see them appear on launchpad).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 .stbt-prefix
 debian-packages
 defaults.conf
+extra/debian/changelog
 extra/fedora/stb-tester.spec
 extra/stb-tester*.debian.tar.xz
 extra/vm/.vagrant
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,8 +85,11 @@ Here are a few guidelines to keep in mind when submitting a pull request:
 
         * Debian:
 
-            * Build the package with `make deb`. (TODO: Docker script to build
-              the deb if your host system isn't Debian/Ubuntu.)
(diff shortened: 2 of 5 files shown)


drothlis:
A couple of tests failed on Travis because "Unable to print keymap because the terminal is too small". This has been fixed in v23-38-g10c25a0, but this branch is branched from v23 in order to publish ppa packages of v23 for new Ubuntu versions (vivid & wily).

wmanley:
So for the release you ran:

    ./ubuntu-shell.sh -c "make debsrc ubuntu_release='vivid wily'"

Which produced 2 .changes files, 2 .dsc files, 2 .orig.tar.gz, 2 debian.tar.gz files.

drothlis:
It turns out there is no point in publishing v23 for Ubuntu 15.04 or later, because of a regression in PyGObject 3.13 which we only fixed after v23 (in fb55265).

We'll have to make a new release (v24) which is long overdue anyway.

This branch is still necessary for building v24 packages for the newer Ubuntu releases, so I'll merge it.

PR #325, opened 2016-01-12
stbt run: Use `kitchen` StreamWriter for stdout & stderr

drothlis:
This allows users to `print` anything they want, even byte-strings with
utf-8 encoded data (or any other non-ascii data).

[kitchen] is a python package with no dependencies. It seems to have
been written by Fedora, judging by its github owner. Its license is
LGPL 2+.

Thanks to @LewisHaley for the suggestion.

[kitchen]: https://github.com/fedora-infra/kitchen


drothlis:
N.B. This is branched from #281 so ignore the first 3 commits.

drothlis:
I haven't tested this manually, just via the unit tests which use `unbuffer` to simulate a terminal, even though the test runner is redirecting output to a log file -- I want to do some manual testing to be 100% sure of the behaviour.

drothlis:
We've decided to fix this now in #281 after all, so I'll close this pull request.

PR #326, opened 2016-01-12
stbt batch run: Add --no-html-report option

drothlis:
Previously we generate an HTML report before each testrun (so that the
report shows a "running..." row for the current test) and again after
the testrun (to update that row with the test's result). This report
generation gets slower as you have more results because it scans all the
results in the output directory each time.

If you have your own reporting system (such as the stb-tester ONE's
database-backed results system), this report generation is superfluous
and only slows things down.

Apart from the "index.html" inside the testrun directory and the
"index.html" one directory above it (at the root of the output
directory), no other files are affected. That is, the result format on
disk hasn't changed.

### DIFF
--- a/stbt-batch.d/run-one
+++ b/stbt-batch.d/run-one
@@ -12,6 +12,7 @@
 #
 # Input environment variables:
 #
+# * $do_html_report ("true" or "false")
 # * $tag
 # * $v
 # * $verbose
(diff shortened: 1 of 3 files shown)


wmanley:
I've pushed a fixup commit which IMO simplifies the tests by removing argument parsing and a conditional.  Apart from that, +1.  @drothlis: please review my fixup.

drothlis:
+1 to your fixup.

wmanley:
Travis is having some difficulty, but one of the builds succeeded so I'll merge.

PR #327, opened 2016-01-12
stbt batch: Fix report generation when output directory name has spaces

drothlis:
The failure was at line 65 of `stbt-batch.d/report`:

61: for resultsdir in $(
62:   for rundir in "$@"; do dirname "$(abspath "$rundir")"; done | sort | uniq)
63: do
64:   (
65:     cd "$resultsdir" || die "Invalid directory '$resultsdir'"

If you used `stbt batch -o 'my results'` it would fail with:

> report: error: Invalid directory '/path/to/my'
> report: error: Invalid directory 'results'

`stbt batch run` would continue to run the tests, but no HTML report
would be generated.

### DIFF
--- a/stbt-batch.d/report
+++ b/stbt-batch.d/report
@@ -12,6 +12,8 @@
 #/ Generates a detailed report in <testrun directory>/index.html
 #/ and a summary report in <testrun directory>/../index.html.
 
+IFS=$'\n'
+
 usage() { grep '^#/' "$0" | cut -c4-; }
 
 main() {
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -69,6 +69,9 @@ UNRELEASED
   `functools.partial` object and `wait_until` timed out. Thanks to Martyn
   Jarvis for the patch.
 
+* `stbt batch run` wasn't generating HTML reports if you gave it an `--output`
+  directory name with spaces.
+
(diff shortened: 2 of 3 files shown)


stbt-batch.d/report
```diff
@@ -12,6 +12,8 @@
 #/ Generates a detailed report in <testrun directory>/index.html
 #/ and a summary report in <testrun directory>/../index.html.
 
+IFS=$'\n'
+
```
wmanley: Ok, I guess so.  I don't particularly like that this will change the behavior of the whole script, when it's only needed for a bit below.  I wouldn't say I'm a confident enough bash-scripter to say whether this is idiomatic or not.

An alternative to:

    for resultsdir in $(
      for rundir in "$@"; do dirname "$(abspath "$rundir")"; done | sort | uniq)
    do

could be:

    for rundir in "$@"; do dirname "$(abspath "$rundir")"; done | sort | uniq | while read resultsdir
    do

None of this is perfect though.  bash is such a pain for things if you want them to be right in all circumstances.

wmanley:
+1.  I trust your judgement on this more than my own.

drothlis:
Thanks for the review. I think it'll be OK, and all the tests pass. As far as being idiomatic, some people recommend using `IFS=$'\n'` in every new bash script you write.

PR #329, opened 2016-02-04
Add Roku remote control

drothlis:
- [x] Test manually with `stbt control` and a real Roku.
- [x] Unit tests.
- [x] Update man page.
- [x] Release note.


drothlis:
@wmanley please review.

_stbt/control.py
```diff
@@ -18,20 +20,21 @@
...
+        (r'lirc(:(?P<hostname>[^:/]+))?:(?P<port>\d+):(?P<control_name>.+)',
+         new_tcp_lirc_remote),
+        (r'lirc:(?P<lircd_socket>[^:]+)?:(?P<control_name>.+)',
+         new_local_lirc_remote),
+        (r'none', NullRemote),
+        (r'roku:(?P<hostname>[^:]+)', RokuHttpControl),
+        (r'samsung:(?P<hostname>[^:/]+)(:(?P<port>\d+))?',
+         _new_samsung_tcp_remote),
+        (r'test', lambda: VideoTestSrcControl(display)),
+        (r'vr:(?P<hostname>[^:/]+)(:(?P<port>\d+))?', VirtualRemote),
```
wmanley: > _stbt.control: Re-order remotes alphabetically

Yes, this is a bit tidier.  These are tried in order but I think the only dependency between two of them is for the lirc entry.

_stbt/control.py
```diff
@@ -242,6 +245,55 @@ def _connect(self):
...
+    """
+
+    # Map our recommended keynames (from linux input-event-codes.h) to the
+    # equivalent Roku keyname.
+    _KEYNAMES = {
+        "KEY_HOME": "Home",
+        "KEY_REWIND": "Rev",
+        "KEY_FASTFORWARD": "Fwd",
+        "KEY_PLAY": "Play",
+        "KEY_PAUSE": "Play",
```
wmanley: Perhaps `KEY_PLAYPAUSE` too

_stbt/control.py
```diff
@@ -242,6 +245,55 @@ def _connect(self):
...
+    }
+
+    def __init__(self, hostname):
+        self.hostname = hostname
+
+    def press(self, key):
+        if key in self._KEYNAMES:
+            roku_keyname = self._KEYNAMES[key]
+        elif len(key) == 1:
+            roku_keyname = "Lit_%s" % key
```
wmanley: I'm not so sure about this branch.  I think if someone wants to use the roku key names they can prepend `Lit_` if they like.  Otherwise we're combining two namespaces that roku have decided upon for whatever reason.

Also: if we did want to support this approach we should probably be translating `KEY_A`, etc. to 'Lit_a'

_stbt/control.py
```diff
@@ -586,6 +638,22 @@ def test_that_local_lirc_socket_is_correctly_defaulted():
...
+    with responses.RequestsMock() as mock:
+        # This raises if the URL was not accessed.
+        mock.add(mock.POST, 'http://192.168.1.3:8060/keypress/Home')
+        control.press("KEY_HOME")
+    with responses.RequestsMock() as mock:
+        mock.add(mock.POST, 'http://192.168.1.3:8060/keypress/Home')
+        control.press("Home")
+    with responses.RequestsMock() as mock:
+        mock.add(mock.POST, 'http://192.168.1.3:8060/keypress/Lit_r')
+        control.press("r")
```
wmanley: Out of interest, what does the roku actually respond with?

_stbt/control.py
```diff
@@ -9,6 +9,8 @@
 import time
 from contextlib import contextmanager
 
+import requests
```
wmanley: I think this probably deserves to be imported in press (or init).  That way it becomes a optional dependency that is only required if you use the roku remote control.

_stbt/control.py
drothlis:
```
$ curl -i -d '' http://192.168.0.17:8060/keypress/Home
HTTP/1.1 200 OK
Server: Roku UPnP/1.0 MiniUPnPd/1.4
Content-Length: 0
```

_stbt/control.py
drothlis: Indeed, I made sure to leave the two `lirc` entries in the same order relative to each other.

_stbt/control.py
drothlis: And:

```
$ curl -i -d '' http://192.168.0.17:8060/keypress/Homeopathy
HTTP/1.1 400 Bad Request
Server: Roku UPnP/1.0 MiniUPnPd/1.4
Content-Length: 0
```

_stbt/control.py
drothlis: Done.

_stbt/control.py
drothlis: Removed the `a` => `Lit_a` translation.

_stbt/control.py
drothlis: Done.

drothlis:
I've addresses all your comments. I've also added a unit test for a bad key name (400 error from the Roku). Thanks for the review.

PR #330, opened 2016-02-11
Fix for Popen Error

Pooze:
When running on OS X this is needed otherwise it throws:
```
OSError: [Errno 2] No such file or directory
```


wmanley:
Whoops, sorry, closed by accident.

What is the command you were running to get the `OSError`?  Can you show the full output including the Traceback too?

Pooze:
Think I might have been too quick, trying to reproduce it. However, even if the `OSError` disappeared it doesn't seems to work. Don't know if it is related but Chrome won't open properly. But I'll get back when I can reproduce it.

Pooze:
Unable to reproduce it.

PR #331, opened 2016-02-11
core: Fix path in error message if template image not found

drothlis:
...in calls to `match`, `wait_for_match`, etc.

Before, this would show an absolute path from wherever the tests were
run, for example if you had the code `match("images/abc.png")` and you
ran it with `stbt batch -o /results` it would say "No such template
file: /results/2015-02-11_17.29.03/images/abc.png". Now it says "No such
template file: images/abc.png".


wmanley:
Nice incremental usability improvement, merging.

PR #332, opened 2016-02-24
API: Introduce `FrameObject` base class

wmanley:
This is a helper class to make implementing [Frame Objects] easier.

[Frame Objects]: https://stb-tester.com/blog/2015/09/08/meet-the-frame-object-pattern

A Frame Object is a class that extracts information from a frame of video,
typically by calling `stbt.ocr()` or `stbt.match()`. All the rest of
your testing code then uses these objects. A Frame Object translates from
the vocabulary of low-level image processing functions and regions (like
`stbt.ocr(region=stbt.Region(213, 23, 200, 36)))` to the vocabulary of
high-level features and user-facing concepts (like `programme_title`).

For more information about frame objects see
https://stb-tester.com/blog/2015/09/08/meet-the-frame-object-pattern

This base class is provided to make creating well-behaved frame objects
easier.  It defines:

*  `__init__` - optionally taking a frame in the constructor
* `__nonzero__` - based on the property is_visible which derived classes must
  define.  This means the class will only be considered `True` if it's
  visible, and so the other properties are valid.  This makes it easy to use
  with `wait_until`.
* `__repr__` - including all the user-defined properties.  This makes using
  the object in doctests convenient
* `__hash__` and `__cmp__` - If all the properties match between two instances
  of a FrameObject then they are considered equal, even if the underlying
  frame is different.  This can be useful for detecting changes or waiting
  for a frame to stop changing before interrogating it.

This ticks three of the boxes in the `Frame Objects checklist`(
https://stb-tester.com/blog/2015/09/08/meet-the-frame-object-pattern)
allowing frame objects to be defined fairly minimally:

    class Guide(FrameObject):
        @property
        def is_visible(self):
            return bool(stbt.match_text("Guide", frame=self._frame))

        @property
        def current_channel(self):
            return stbt.ocr(frame=self._frame)

        @property
        def current_time(self):
            m = stbt.match('clock.png', frame=self._frame)
            return stbt.ocr(
                region=m.region.extend(x=20, right=100), frame=self._frame)

TODO:

- [x] Tests
- [x] Pull test example into API documentation.
- [x] Check generated documentation
- [x] Come up with a story for testing, even if it's not implemented in this PR
    - Allow a class member `EXAMPLE_SCREENSHOTS = "guide-*.png"`
- [x] I'm not a big fan of inheriting to reuse, though I think alternative approaches (like using a class decorator) may be just less readable?  But it might be clearer that this does something a little odd, particularly if we add caching to properties.
    - Yep, just have to come to terms with this.  It's implemented as a metaclass which makes it look like normal inheritance.
- [x] Add caching
- [x] Release note
- [x] Squash documentation commits into single commit
- [x] Reinstate ordering tests in `test_frame_object.py`

Future work

- Automatic selftests
- A pylint check that checks that you're passing `frame=` into all the stbt functions you're calling and not calling `sleep`


wmanley:
In favour of a `FrameObject` base class:

* If the user wants to pass additional stuff into the constructor it's easy to do so
* Overriding things in the base class is understood, as is chaining up the the base class implementation

Against a `FrameObject` base class:

* The base class must introspect the derived class to be useful

In favour of a `FrameObject` class decorator:

* It's a more natural way to modify a class (e.g. adding caching to properties)
* We can validate that the user hasn't messed up at parse time rather than waiting for it to be instantiated

Against a `FrameObject` base class

* Defining an `__init__` that takes a `self._frame` is less straight forward if the user has added additional parameters to `__init__`
* I don't know if pylint will understand `self._frame` this way

_stbt/core.py
```diff
@@ -1191,6 +1191,52 @@ def argparser():
...
+    def __nonzero__(self):
+        return bool(getattr(self, self.__attrs[0]))
+
+    def _astuple(self):
+        if self:
+            return tuple(getattr(self, x) for x in self.__attrs)
+        else:
+            return (False,)
+
+    def __cmp__(self, other):
```
drothlis: I don't know that an ordering of these instances makes sense... Shouldn't we define `__eq__` and `__ne__` instead?

_stbt/core.py
```diff
@@ -1191,6 +1191,52 @@ def argparser():
...
+        # pylint: disable=W0212
```
drothlis: I new code we should start using pylint's short descriptions: `pylint: disable=protected-access`.

_stbt/core.py
```diff
@@ -1191,6 +1191,52 @@ def argparser():
...
+        if isinstance(other, self.__class__):
+            return cmp(self._astuple(), other._astuple())
+        else:
+            return NotImplemented
+
+    def __hash__(self):
+        return hash(self._astuple())
```
drothlis: Why do we want these to be hashable? Why would you use them in a set, or as keys in a dict? If you had 2 different frames (but showing the same UI elements, at least those exposed by properties of the FrameObject) they'd be treated as the same key.

I understand why we want them to be `__eq__`, and given that "Hashable objects which compare equal must have the same hash value", then I suppose this can't hurt. Just wondering if you had a particular use-case in mind.

drothlis:
I like it!

It seems clear to me from the list of pros & cons that a base class is the way to go. Especially for sub-classes that need to provide their own `__init__`.

> I'd like to add caching in the future. Will this be possible with this design?

I think we should add memoization of properties in this PR -- it wouldn't be hard to do and it's very useful. I'm not sure if you meant memoization, or more long-term caching (across process lifetime, to speed up self-tests).

_stbt/core.py
wmanley: Being able to add these to a set was useful in a previous project.  I think I was iterating through a list of things and wanted to know when I'd come across one I'd seen before or something like that.

wmanley:
> It seems clear to me from the list of pros & cons that a base class is the way to go. Especially for sub-classes that need to provide their own `__init__`.

I've added an additional option to the pros and cons, using a metaclass.  I think it's at least worth prototyping.

_stbt/core.py
wmanley: I'm thinking of these like a lazily evaluated `namedtuple`, so defining `__cmp__` seems natural to me.  Although I think this `__cmp__` could be improved so as to avoid querying all the properties at once invalidating lazyness.

wmanley:
Thinking about this I think the mechanism can be decomposed into two separable parts:

* Creating a `self._frame` member, using `stbt.get_frame()` if none is provided.  This is just 
* Turning something with properties into something `namedtuple`-like (most suited to the decorator approach)

wmanley:
Plus there's one slightly anomalous bit: If `is_visible` is `False`ish the object almost behave like `None`.  In this mode `__repr__`, `__eq__`, `__hash__`, etc. will behave like the other properties don't exist at all, but you can still call them and they may return valid values.

This could be a little surprising.

Maybe all the properties should return None if not `self.is_visible()`?

* This would also be surprising, as you've defined as you may have defined a property that never returns `None`.

Maybe we should be defining our own `@stbt.property`

* Less surprising
* Another thing to remember and get wrong

wmanley:
Ok, I've made it a metaclass and added caching and tests.

@drothlis: Please review

_stbt/core.py
```diff
@@ -1191,6 +1191,98 @@ def argparser():
...
+            for x in self.__class__.__attrs:
+                yield x, getattr(self, x)
+        else:
+            yield "is_visible", False
+
+    def __nonzero__(self):
+        return bool(self.is_visible)
+
+    def __cmp__(self, other):
+        # pylint: disable=W0212
```
drothlis: Could you make this `pylint: disable=protected-access`?

stbt/__init__.py
```diff
@@ -431,6 +431,58 @@ def is_screen_black(frame=None, mask=None, threshold=None):
...
+      the object in doctests convenient
+    * __hash__ and __cmp__ - If all the properties match between two instances
+      of a FrameObject then they are considered equal, even if the underlying
+      frame is different.  This can be useful for detecting changes or waiting
+      for a frame to stop changing before interrogating it.
+
+    This ticks three of the boxes in the `Frame Objects checklist`(
+    https://stb-tester.com/blog/2015/09/08/meet-the-frame-object-pattern)
+    allowing frame objects to be defined fairly minimally:
+
```
drothlis: Ideally this documentation would be more self-contained and wouldn't need to refer to the blogpost twice.

It's a shame it doesn't include the documentation in `test_frame_object.py` as this documentation here is what shows up in our python api docs at <https://stb-tester.com/manual/python-api>. Not sure what we could do about that, define the entire `Dialog` example frame object here in the docstring?

_stbt/core.py
```diff
@@ -1191,6 +1191,98 @@ def argparser():
...
+                    f = _noneify_property_fn(f)
+                dct[k] = property(f)
+
+        return super(_FrameObjectMeta, cls).__new__(cls, name, parents, dct)
+
+    def __init__(cls, name, parents, dct):
+        property_names = sorted([
+            p for p in dir(cls)
+            if isinstance(getattr(cls, p), property)])
+        assert 'is_visible' in property_names
```
drothlis: Could we have a nicer error message for this?

wmanley: This is a genuine assertion.  `is_visible` is defined in the base class so this should really never happen.

stbt/__init__.py
wmanley: > define the entire Dialog example frame object here in the docstring?

Done

wmanley:
Ok, I've tidied up the documentation.  I'm still not 100% happy with it but it's better than before.  This is a little bit difficult to document as there are many different ways to approach it:

1. How does the `FrameObject` base class behave and what does it do
2. What are Frame Objects and why should you be using them?
3. How to get started using the `FrameObject` base class as quickly as possible

I'll have to update the blog post once this is merged too.

drothlis:
@wmanley: I've taken the the documentation from your `frame-object-docs-refactor` branch and reworked it slightly. The main difference (which I've kept in a separate commit for now for ease of review) is that it defines the Frame Object semantics in the text before the example. I tried to keep it short by moving any motivations/justifications to the example. Let me know if you're happy with it, and I'll squash my changes together.

drothlis:
P.S. The documentation is really looking quite good, nice work.

drothlis:
Very nice indeed.  It's a lot easier to understand the way you've done it.  I also particularly like the addition of:

> If you define any methods that change the state of the device-under-test, they should return a new Frame Object instead of modifying self.

That's an important point that was missing otherwise.

wmanley:
Very nice indeed.  It's a lot easier to understand the way you've done it.  I also particularly like the addition of:

> If you define any methods that change the state of the device-under-test, they should return a new Frame Object instead of modifying self.

That's an important point that was missing otherwise.

PR #334, opened 2016-02-25
stbt-run: Don't install a trace handler if it's not necessary

LewisHaley:
We're attempting to forward-port our own patches onto a new version of STBT, and in doing so found a clash between our trace and stbt's build in tracer (https://github.com/stb-tester/stb-tester/commit/fa60ea989c9d87e82af04eaaab76cfe6409d194b). The other commits are tidy-ups.

I wasn't able to get `make check` to pass fully because of failures with the virtual-stb tests: 

> /usr/libexec/Xorg.wrap: Only console users are allowed to run the X server

### DIFF
--- a/_stbt/state_watch.py
+++ b/_stbt/state_watch.py
@@ -7,7 +7,6 @@ from cStringIO import StringIO
 
 from _stbt import utils
 
-
 _TRACER = None
 
 
--- a/_stbt/state_watch.py
+++ b/_stbt/state_watch.py
@@ -5,6 +5,11 @@ import socket
 import sys
 from cStringIO import StringIO
 
+from _stbt import utils
+
+
+_TRACER = None
+
 
 class StateSender(object):
     """
@@ -209,6 +214,36 @@ class StateReceiver(object):
             except StandardError as e:
                 sys.stderr.write(
                     "Error processing state change: %s" % str(e))
+
+
+def activate(test_script, trace_file=None):
+    """Activate the state-watch tracer."""
+    global _TRACER
+    _TRACER = new_state_sender(trace_file)  # pylint: disable=W0212
+
+    absfilepath, funcname, function = utils.parse_test_uri(test_script)
+
+    def tracefunc(frame_, event, _):
+        if event == "line" and frame_.f_code.co_filename == absfilepath:
+            _TRACER.log_current_line(frame_.f_code.co_filename, frame_.f_lineno)
+        return tracefunc
+
+    sys.settrace(tracefunc)
+    _TRACER.log_test_starting(
+        test_script, os.path.relpath(absfilepath), funcname,
+        function.func_code.co_firstlineno if function is not None else 1)
+
+
+def deactivate():
+    """Deactivate the state-watch tracer.
+
+
+    If the tracer has not been `activate`d, this function is a no-op.
+    """
+    if _TRACER is not None:
+        sys.settrace(None)
+        _TRACER.log_test_ended()
+        _TRACER.close()
 
 
 def test_statereceiver():
--- a/stbt-run
+++ b/stbt-run
@@ -14,6 +14,7 @@ import traceback
 import _stbt.core
 import stbt
 from _stbt.state_watch import new_state_sender
+from _stbt.utils import parse_test_uri
 
 parser = _stbt.core.argparser()
 parser.prog = 'stbt run'
@@ -72,14 +73,6 @@ def _print_exc_utf8(file_=None):
         traceback._some_str = _old_some_str
 
 
-def import_by_filename(filename_):
-    module_dir, module_file = os.path.split(filename_)
-    module_name, module_ext = os.path.splitext(module_file)
-    if module_ext != '.py':
-        raise ImportError("Invalid module filename '%s'" % filename_)
-    sys.path = [os.path.abspath(module_dir)] + sys.path
-    return __import__(module_name)
-
 _tracer = None
 try:
     # pylint: disable=W0611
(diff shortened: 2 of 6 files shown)


wmanley:
I like this in theory, but I'd like to see some changes to the implementation.  I'll make some suggestions as line comments.

_stbt/state_watch.py
```diff
@@ -211,6 +215,36 @@ def write(self, data):
...
+def deactivate():
+    """Deactivate the state-watch tracer.
+
+
+    If the tracer has not been `activate`d, this function is a no-op.
+    """
+    if _TRACER is not None:
+        sys.settrace(None)
+        _TRACER.log_test_ended()
+        _TRACER.close()
```
wmanley: I think this would be better as a `@contextmanager` member of `StateSender`.  That way you can share state between `activate` and `deactivate` without needing a `_TRACER` global in this file.  In `stbt-run` you could have:

    with _tracer.trace_test_execution(args.script, _filename, function):
        # The rest of the stuff

and the `trace_test_execution` can decide for itself whether to install a trace handler.

stbt-run
```diff
@@ -22,7 +25,7 @@ parser.add_argument(
     '--save-video', help='Record video to the specified file', metavar='FILE',
     default=stbt.get_config('run', 'save_video'))
 parser.add_argument(
-    '--save-trace', metavar='FILE', default=None,
+    '--save-trace', metavar='FILE', nargs='?', const=TO_SOCKET,
```
wmanley: I don't think we need this 'TO_SOCKET' sentinel.  `trace_test_execution` can decide whether to install the trace handler based on whether the `STBT_TRACING_SOCKET` environment variable is set.  In general I'm not a fan of using values that would otherwise be valid as sentinels.  You may think that it's unlikely that someone would want to log to a file called `TO_SOCKET` but the presence of this special value can be the cause of confusing failures, particularly when `stbt-run` is being scripted.

_stbt/state_watch.py
wmanley: Note: you may find actually using a `with` statement in `stbt-run` to be inconvenient, but you can still use the [context manager protocol](https://www.python.org/dev/peps/pep-0343/) directly.

wmanley:
> We're attempting to forward-port our own patches onto a new version of STBT

Do you have those patches pushed to a repo somewhere?  I'd be interested in seeing them.

_stbt/state_watch.py
LewisHaley: Seems like a fair bit of extra work if it's simply to work around having a global variable? It's not like you would/could instantiate multiple tracers because of the `sys.settrace` limitation, and ultimately it's not user-facing

And isn't
```python
state_watch.deactivate()
```
much nicer to read (and easier to understand) than
```python
if _tracer is not None:
    _tracer.__exit__(None, None, None)
```

stbt-run
LewisHaley: I agree that having to have a sentinel isn't great, but if the tracing is conditional on if `STBT_TRACING_SOCKET` is set then
`STBT_TRACING_SOCKET=x stbt run --save-trace -- test.py` and
`STBT_TRACING_SOCKET=x stbt run test.py`
will both result in the tracing being activated, which is then particularly troublesome if we or someone else were to push our own tracing functionality upstream, resulting in:
`STBT_TRACING_SOCKET=x stbt run --custom-trace=trace.log test.py`
where some secret, internal decision has to be be made about which trace takes priority.

Perhaps a better solution is to have separate, mutually exclusive command line arguments, one which writes to a file and one which writes to a socket.

_stbt/state_watch.py
wmanley: > Seems like a fair bit of extra work if it's simply to work around having a global variable?

I doesn't seem too difficult to me.  Either way I've been putting in a bunch of effort to remove global state (#297 #333) so I don't want to merge something that introduces more.

> It's not like you would/could instantiate multiple tracers because of the sys.settrace limitation, and ultimately it's not user-facing

[`sys.settrace` docs](https://docs.python.org/2/library/sys.html#sys.settrace) say:

> The function is thread-specific; for a debugger to support multiple threads, it must be registered using settrace() for each thread being debugged.

And we could use `sys.gettrace` to chain the trace functions together, not that I'm suggesting that this would be necessary for this pull request.

> much nicer to read (and easier to understand) than
> 
>     if _tracer is not None:
>         _tracer.__exit__(None, None, None)
￼
Note: Under my proposed scheme `_tracer` in `stbt-run` would never be `None`, it just might not install a trace handler.

stbt-run
wmanley: > we or someone else were to push our own tracing functionality upstream

I can't make technical judgements based upon the contents of patches that I've never seen.

drothlis:
For the record, here's how it's intended to work:

* If `STBT_TRACING_SOCKET` is set, write tracing information to the
  specified socket. This is intended for live monitoring of a test as
  it's running, from a GUI such as the stb-tester ONE's GUI.
* If `--save-trace` is specified, write tracing information to the
  specified file. This is intended for later off-line playback of the
  test-run, for example a GUI that draws `match` rectangles over the
  video as it's playing back so that `stbt run` doesn't have to
  hard-code the rectangles into the recorded video stream. In future
  (once the playback tools are in place) this will save CPU and result
  in a better quality video recording, because it won't have to be
  re-encoded (for example you might be able to use the H.264 you got
  from your video source directly).

You can specify either one, or none, or both -- neither takes precedence
over the other. This decision is encapsulated in `state_watch.py`;
`state_watch.new_state_sender` will return a dummy `StateSender` if
neither `STBT_TRACING_SOCKET` nor `--save-trace` are specified. That
way, `stbt-run` doesn't need to have an `if` statement surrounding every
logging statement.

drothlis:
@LewisHaley would you be interested in porting your tracing to use stb-tester's mechanism? What does it log that stb-tester doesn't currently? I understand that our mechanism isn't comprehensively documented so we'll be happy to assist if you have any questions.

> I can't make technical judgements based upon the contents of patches that I've never seen.

In case that came across as grumpy: It wasn't meant that way; we do appreciate your contributions, including any minor cleanups. Just be aware that as it stands this pull request is asking us to spend time & effort (in code review) for no benefit to the project, only to your private fork -- and time is in short supply.

Having said that, we'll be happy to merge this anyway, because it *is* cleaner than what's already there, and we don't want to actively hinder you either. But you'll need to address @wmanley's comments -- in particular, we won't entertain changing the current command-line API (that is, `STBT_TRACING_SOCKET` and `--save-trace`) nor adding new global state anywhere.

drothlis:
> I wasn't able to get make check to pass fully because of failures with the virtual-stb tests:
>
> > /usr/libexec/Xorg.wrap: Only console users are allowed to run the X server

I don't have a Fedora host to test this with, but do you have xorg-x11-drv-dummy installed?

drothlis:
@LewisHaley your problem will be fixed by #509 so I'll close this. It would still be nice to know what your own tracing does / what your tracing requirements are, at least at a high level.

PR #337, opened 2016-02-28
tests: Use our own copy of webminspector.py 

drothlis:
...to fix `test_that_video_index_is_written_on_eos`.

The upstream project deleted webminspector.py from the repo in December 2015, so our test hasn't run in 2 months (it skips if it can't find webminspector.py).


drothlis:
`test_that_video_index_is_written_on_eos` ran & passed on Travis.

wmanley:
+1.  It's a shame to bundle external projects in here, but this certainly seems better than the alternatives.

wmanley:
One alternative would be to make our own git clone of webminspector on stb-tester github and clone from there instead

wmanley:
But I think this is fine.  It's ~1300 lines of code and self-contained.

PR #338, opened 2016-03-01
Makefile: Add `make shell`

wmanley:
This is intended as a kind of virtualenv-lite.  It installs the
dev-dependencies that often change in incompatible ways into a
subdirectory pinned at the versions we use in CI.

Unlike a virtualenv we rely on the system packages for our runtime
dependencies because they are more stable and we want to be compatible
with these packages on as many systems as practical.


drothlis:
Closing. These days we'd use venv with `--system-site-packages`.

PR #341, opened 2016-03-03
stbt virtual-stb: Fix relative path given to --x-keymap argument

wmanley:
When launching stbt run from a different directory to which `stbt virtual-stb` was launched.


wmanley:
This also fails early if the keymap file doesn't exist rather than failing when `stbt run` is run.

drothlis:
:+1:

PR #342, opened 2016-03-04
More camera calibrate improvements

wmanley:
* Mostly readability improvements and documentation
* Improved separation between the calibration process and the element that will be performing the correction by defining a dictionary containing the parameters to the correction function.  These are mathematically well defined.  For definitions see [Camera Calibration and 3D Reconstruction
](http://docs.opencv.org/2.4/modules/calib3d/doc/camera_calibration_and_3d_reconstruction.html).  This will be useful for a separate implementation in OpenGL.
* Fixes to the calibration ASCII art


drothlis:
Rebased onto master.

The camera tests (run locally, because Travis doesn't run them) all pass except test_that_stbt_camera_calibrate_corrects_for_geometric_distortion and test_that_validation_passes_on_pristine_input -- but both of those also fail on v25.

drothlis:
CI failure is unrelated (the unexplained, intermittent GObject error we've seen before: `AttributeError: type object 'GstMapFlags' has no attribute 'READ'`).

drothlis:
@wmanley says that the test failure is due to font rendering discrepancies in the test infrastructure. I'm going to merge this.

PR #343, opened 2016-03-09
Add tool stbt auto-selftests

wmanley:
An implementation of the technique described in [Improve black-box testing agility: automatic self-regression tests](https://stb-tester.com/blog/2015/09/24/automatic-self-regression-tests).

You put screenshots of your UI into selftest/screenshots in your test pack and run

    stbt auto-selftests generate

and it will produce a bunch of python doctests in selftest/auto-selftests describing how each of your `FrameObject`s interpret each of these screenshots.

Then make some changes to your FrameObjects and rerun

    stbt auto-selftests generate

and look at the diff to see how they've changed.

By default it will try every `FrameObject` against every screenshot printing a result to the doctests only if the resulting `FrameObject` is `True`.  Otherwise the result is omitted from the doctests.  This behaviour can be more finely controlled by setting `AUTO_SELFTEST_EXPRESSIONS`, `AUTO_SELFTEST_SCREENSHOTS` or `AUTO_SELFTEST_EXCLUDES` members of your `FrameObject`s.

**TODO:**

- [x] Merge #332 and rebase
- [x] Update `Makefile` and `stbt.sh.in` to install it
- [x] Add more tests
- [x] Test operating on a whole test-pack
- [x] Don't rely on test_frame_object.py for testing `stbt auto-selftests`, create some specific tests instead and leave it unchanged.
- [x] Add `stbt auto-selftests validate` tool
- [x] Reference documentation
- [x] Remove `AUTO_SELFTEST_EXCLUDES` in favour of a way of specifying which images it should opportunistically test against. This would default to `'*.png'`
- [x] Test `stbt auto-selftest validate` failing to validate
- [x] Add release note
- [x] Consider functions that throw exceptions "Truthy"
- [x] Handle unicode & utf8-byte-strings
- [x] Put it behind `stbt --with-experimental` flag?
- [x] Bash completion
- [x] Fix failing unicode in `AUTO_SELFTEST_EXPRESSIONS`
- [x] Change exit status of `stbt selftest validate` when there are no `FrameObject`s or screenshots
- [x] Port (private) auto-regressiontests to this system and test.
- [x] Document `AUTO_*_*`
- [x] Don't leave behind a bunch of `_bare.pyc` files
- [x] Remove empty directories
- [x] Don't generate tests for classes that have just been `import`ed from other files
- ~~`--screenshots-root`, if given, should be relative to current directory, not test-pack root.~~
- [x] Remove `--screenshots-root`

**For a future PR:**

- Caching of expensive stbt operations to make it feasible to run with very many images
- Better story for calling other properties from `is_visible`. Currently it goes into an infinite loop. Options:
    - Automatically disable the "return None if not visible" behaviour for properties if we detect that `is_visible` (for the *same object*) is on the call stack.
    - Detect and raise a more useful exception early.
    - Don't detect it at runtime, write a `stbt lint` checker instead.
- Add `stbt auto-selftests retro-regenerate` tool? - very useful, but maybe too advanced and too dangerous?
- Some sort of pretty-printing
- More general documentation covering best-practices and how it all fits together
- A `stbt check` command to run all the checks that we recognise.


drothlis:
Some thoughts on pretty-printing the doctest output:

* If we're ever going to do pretty-printing we need to add it now, or when you upgrade stb-tester the diff to your auto-regression-tests would be so noisy that it hides actual impactful changes.
* Python's stdlib `pprint` isn't designed to be extended to user-defined classes.
* iPython's [pretty](https://ipython.org/ipython-doc/stable/api/generated/IPython.lib.pretty.html) is intended to replace the old `pprint`.
    * Your class implements `_repr_pretty_(self, p, cycle)` where `p` is a pretty-printer object that gets passed in.
    * We could implement that protocol in `FrameObject` even if we didn't want to add a dependency on iPython.
* A simpler option would be to have the doctest generator implement its own pretty printer, accessing `FrameObject` internals.
    * If the `__repr__` output is > 80 chars, split it to 1 property per line. Do allow each property to span more than 80chars.

drothlis:
`AUTO_SELFTEST_EXCLUDES` should instead be a whitelist. If you want to exclude some screenshots (because you have so many that it takes too long to run the auto-regression-tests) then the common use-case is: For my "Guide" frame object, test all the "guide-*.png" screenshots, plus this other one and this other one as regression tests for some old false positive that I've now fixed. I don't see how blacklisting all the non-Guide areas of my UI, explicitly, would be useful.

wmanley:
> A simpler option would be to have the doctest generator implement its own pretty printer, accessing FrameObject internals.

Another alternative I'm leaning towards is to post-process the `__repr__` output itself to add whitespace.  This would have the advantage of working for any `repr`.  It wouldn't have to be too clever and it could bail out if it gets confused.

wmanley:
One aspect I'm umming and ahhing about is how we should classify doctests that resulted in an exception.  Currently they are considered to be `False`y so are excluded if not part of `AUTO_SELFTEST_SCREENSHOTS`.  Perhaps exceptions are sufficiently interesting that they should be included.

wmanley:
> `AUTO_SELFTEST_EXCLUDES` should instead be a whitelist

Good idea, I've replaced this with `AUTO_SELFTEST_TRY_SCREENSHOTS` which is a whitelist for screenshots that will be included if they match.

wmanley:
I've been thinking about it.  I'm not too concerned with getting the pretty-printing right first time.  We can change it in the future as long as we only make white-space changes.  Users can just use `git diff --ignore-all-space`.

wmanley:
@drothlis: Please review

_stbt/core.py
```diff
@@ -2196,6 +2196,13 @@ def _log_image_descriptions(
         ))
 
 
+def _iter_frames(depth=1):
```
drothlis: Call it `_iter_stack` to avoid overloading the term "frame" (of video)? And for similarity with the previous implementation -- the caller diff would be:

    - for caller in inspect.stack()[2:]
    + for caller in _iter_stack(depth=2):

I wasn't sure about the name `depth` -- it suggests that you iterate up to a certain depth, rather than starting from a certain depth -- but I see that it's the terminology used by [`sys._getframe`](https://docs.python.org/2/library/sys.html#sys._getframe).

_stbt/core.py
```diff
@@ -2196,6 +2196,13 @@ def _log_image_descriptions(
...
+    frame = inspect.currentframe(depth + 1)
```
drothlis: `inspect.currentframe` is CPython only. I suppose that's OK as we're pretty much wedded to CPython 2.7 at this point.

Makefile
```diff
@@ -159,7 +159,8 @@ check-nosetests: tests/ocr/menu.png
 	              -e tests/test2.py \
 	              -e tests/test_functions.py \
 	              -e tests/vstb-example-html5/ \
-	              -e tests/webminspector/) \
+	              -e tests/webminspector/ \
+	              -e tests/auto-selftest-examples/bare.py) \
```
drothlis: This was sorted alphabetically...

stbt_auto_selftest.py
```diff
@@ -4,12 +4,92 @@
...
+
+            oldstdout = sys.stdout
+            io = StringIO.StringIO()
+            result = None
+            try:
+                sys.stdout = io
+                try:
+                    result = eval(cmd, module.__dict__)  # pylint: disable=W0123
+                except SyntaxError:
+                    exec cmd in module.__dict__  # pylint: disable=W0122
```
drothlis: Might want to warn or fail if somebody used `# remove-if-false` with an expression?

(If the `exec` path is taken here, `result` will be `None`. So if you used `# remove-if-false` anything printed by the selftest-expression will be removed from the outfile.)

stbt_auto_selftest.py
```diff
@@ -4,12 +4,92 @@
...
+
+                if result is not None:
+                    io.write("%s\n" % repr(result))
+                truthy = bool(result)
+            except Exception:  # pylint: disable=W0703
+                io.seek(0)
+                io.truncate()
+                traceback.print_exc(0, io)
```
drothlis: Probably should use `kitchen` here for the same reason we do in `stbt run`.

stbt_auto_selftest.py
```diff
@@ -4,12 +4,92 @@
...
+                truthy = False
+            finally:
+                sys.stdout = oldstdout
+
+                if not remove_if_false or truthy:
+                    outfile.write(("%s" % line).encode('utf-8'))
```
drothlis: Probably should use `kitchen` here for the same reason we do in `stbt run`.

drothlis:
I've reviewed up to "Add `update_doctests`" (I'll continue tomorrow).

Notes so far:

* `# remove-if-false` handling for statements (as opposed to expressions)
* Unicode handling
* We might want to hide `stbt auto-selftest` behind `--with-experimental` until we implement caching and settle on the output formatting.

I've pushed some fixups including one that will fail travis until we fix the unicode handling.

drothlis:
@wmanley I was going to push the attached patch but I see you're working on this right now.

[0001-stbt-auto-selftest-Handle-unicode-or-utf-8-bytes.patch.txt](https://github.com/stb-tester/stb-tester/files/178284/0001-stbt-auto-selftest-Handle-unicode-or-utf-8-bytes.patch.txt)

drothlis:
@wmanley let me know when you're ready for me to continue reviewing.

wmanley:
Please continue.  I didn't want to use kitchen because I want to match the output of doctest exactly and so want more control than that.

I've rewritten `update_doctests` to be more like the [real doctest](https://github.com/python/cpython/blob/2.7/Lib/doctest.py#L1309-L1322).

drothlis:
Squashed & force-pushed.

stbt_auto_selftest.py
```diff
@@ -4,12 +4,133 @@
...
+    with NamedTemporaryFile() as outfile:
+        update_doctests(_find_file('tests/auto-selftest-examples/bare.py'),
+                        outfile.name)
+        actual = outfile.read()
+        with open(_find_file('tests/auto-selftest-examples/expected.py')) as f:
+            expected = f.read()
+        if actual != expected:
+            subprocess.call(
+                ['diff', '-u', _find_file('tests/auto-selftest-examples/'
+                                          'expected.py'), outfile.name])
```
drothlis: This diff output is so much more helpful. :-)

drothlis:
I've pushed some commits trying for a more consistent naming (of the files in our own `tests/` directory) as I found it a bit confusing.

Also pushed some work on the `stbt auto-selftest --help` output.

drothlis:
I'm pretty happy with the `--help` output but I do think I'll do some further work to document `AUTO_SELFTEST_SCREENSHOTS` etc in the help text rather than leaving it to example code (like the change I did for the FrameObject documentation).

drothlis:
> I've pushed some commits trying for a more consistent naming (of the files in our own tests/ directory)

I've moved the example test-pack to `tests/auto-selftest-example-test-pack`. Perhaps it should just be `tests/example-test-pack` and it can hold other things as we codify more of our test-pack best practices into stbt.

tests/auto-selftest-example-test-pack/tests/example.py
```diff
@@ -0,0 +1,64 @@
...
+        return False
+
+
+# If we want an item in a module to be tested we just need to add a
+# AUTO_SELFTEST_EXPRESSIONS member to it.  In fact this is exactly why
+# stbt auto-selftests knows to test FrameObjects: the FrameObject base class
+# defines this member for you.
+#
+# Here's an example of testing a function instead of a class:
+def not_a_frame_object(name, _):
```
drothlis: I'd like this function example to be a bit more realistic, seeing we link to it as documentation from the output of `stbt auto-selftest --help`. For example it could be `is_dialog_visible`. Maybe it could print a debug message only if it does see a dialog.

tests/test-stbt-auto-selftest.sh
```diff
@@ -0,0 +1,59 @@
...
+{
+    cd_example_testpack &&
+    stbt auto-selftest validate
+}
+
+test_that_no_auto_selftests_fail_stbt_auto_selftest_validate()
+{
+    cd_example_testpack &&
+    rm -r selftest/auto_selftest &&
+    ! stbt auto-selftest validate
```
drothlis: Is this really the behaviour we want? If we add a `stbt check` that automatically runs pylint + `stbt auto-selftest validate` on every commit, we wouldn't want it to fail for test-packs that haven't defined any FrameObjects or selftest-screenshots.

wmanley:
@drothlis: I like all of your changes, thank you.

tests/test-stbt-auto-selftest.sh
wmanley: I suppose we could say:

* If there are no `FrameObject`s then having no tests shouldn't fail
* If there are no screenshots then having no tests shouldn't fail
* If there are `FrameObject`s and screenshots then having no tests **should** fail

I guess we should test each of these cases.

drothlis:
Any one of these causes `stbt auto-selftest generate` to fail with a UnicodeDecodeError:

```
identity.AUTO_SELFTEST_EXPRESSIONS = [
    '"Spinal Tap"',
    'u"Spinal Tap"',
    '"Spın̈al Tap"',
    'u"Spın̈al Tap"',
    'print "Spinal Tap"',
    'print u"Spinal Tap"',
    'print "Spın̈al Tap"',
    'print u"Spın̈al Tap"',
    'raise Exception("Spinal Tap")',
    'raise Exception(u"Spinal Tap")',
    'raise Exception("Spın̈al Tap")',
    'raise Exception(u"Spın̈al Tap")',
]
```

wmanley:
> * reStructuredText hyperlink targets don't look nice as plain text.

Agreed.  I guess we can solve the problem of how to document this outside `--help` output seperately.

wmanley:
> * Document `AUTO_*_*`

I took this off the TODO list as it's already documented in `example.py` which I think acts as a suitable reference for the time being.  I've tried writing something more formal but it always seems to come out less clear.

wmanley:
I've made a fairly significant re-factoring separating the module inspection logic from the test-generation logic.  This improves error handling and robustness.

I've run it on our private test-packs and fixed any issues I've seen from there.  It's super slow without caching (52 min to run) but it now works at least.

All the TODO items are ticked off now.  Please merge if happy.

stbt_auto_selftest.py
```diff
@@ -135,6 +139,9 @@ def generate_into_tmpdir(screenshots_root):
         for module_filename in _recursive_glob('*.py'):
             if module_filename.startswith('selftest'):
                 continue
+            if not is_valid_python_identifier(
+                    os.path.basename(module_filename)[:-3]):
+                continue
```
drothlis: Shouldn't we be catching exceptions from `inspect_module` instead?

tests/auto-selftest-example-test-pack/tests/empty_dir/subdir/example_with_no_tests.py
```diff
@@ -0,0 +1,2 @@
+# There should be no selftest file generated for this file because there are
+# no tests to be generated, but also the directory it's in should be deleted.
```
drothlis: Might also be worth having a test in a subdir that *does* have selftests generated for it?

stbt_auto_selftest.py
wmanley: This doesn't cause a problem for `inspect_module`.  The problem happens later when we try to fill in the doctest that we've created.  It would begin with `import bad-module-name` which is a `SyntaxError`.  Since I know it's going to fail I prefer to skip early to save importing it at all.

drothlis:
I've pushed a couple of small fixups. Looks good to me otherwise.

I've added "Better story for calling other properties from `is_visible`" (currently it enters an infinite loop) to the todo list for future PRs.

stbt_auto_selftest.py
```diff
@@ -77,7 +80,6 @@ def main(argv):
     subparsers.add_parser('validate', help='Run (and check) the auto-selftests')
 
     cmdline_args = parser.parse_args(argv[1:])
-    screenshots_root = cmdline_args.screenshots_root or "selftest/screenshots"
```
wmanley: This isn't quite right.  I think that this should have been:

    if cmdline_args.screenshots_root is None:
        screenshots_root = os.path.join(root, "selftest/screenshots")
    else:
        screenshots_root = os.path.abspath(cmdline_args.screenshots_root)

As it is currently the `--screenshots-root` will always be taken relative to the test-pack root if it's a relative path.

drothlis: Isn't that the desired behaviour?

wmanley: I don't think so.  I use git as a reference point.  If you are in a subdirectory of your git repo and do `git add .` you expect the subdirectory to be added, not the whole repo.

drothlis: So, to be clear, my commit is a refactoring that doesn't change the existing behaviour, but now you think the behaviour should be different? I'll add an item to the todo list.

drothlis: What if you run `stbt auto-selftests` outside the test-pack directory (e.g. in your CI's temporary directory, or in your stbt results directory)? Is the `--screenshots-root` still relative to the working directory? I suppose that some people might want to keep their screenshots out of the git repo?

Note that you won't be able to change the location (within your test-pack) and still give a relative path to it. I suppose that's ok, we don't need to support that.

Either way, we should make sure that giving an absolute path always does the right thing.

wmanley: > So, to be clear, my commit is a refactoring that doesn't change the existing behaviour, but now you think the behaviour should be different?

Yes, that's right.  Originally I'd added `--screenshots-root` for testing but didn't end up using it.  I'll remove it instead.

wmanley:
> Yes, that's right. Originally I'd added --screenshots-root for testing but didn't end up using it. I'll remove it instead.

Done and squashed

tests/auto-selftest-example-test-pack/tests/empty_dir/subdir/example_with_no_tests.py
wmanley: Done

wmanley:
Ok, I think this is ready for real now.  I'll merge it if it passes CI.

Issue #345, opened 2016-03-31
match_text doesn't produce debug output


rinaldomerlo:
SUMMARY
There is no debug output from match_text, even in verbose mode. This is inconsistent with other stbt functions like match.

STEPS TO REPRODUCE
  assert stbt.match_text(“Some Text”)

EXPECTED RESULTS
  stbt-run: No match found: 'Some Text'
  stbt-run: No match found: 'Some Text'
or
  stbt-run: Match found: 'Some Text' Region(x=532, y=446, width=127, height=31)

ACTUAL RESULTS
  Nothing, other than the assert if it fails.

rinaldomerlo:
slight amendment to the steps to reproduce above..

STEPS TO REPRODUCE
  stbt run --verbose assert stbt.match_text(“Some Text”)

PR #346, opened 2016-03-31
Update core.py for #345

rinaldomerlo:
Proposed patch for issue:

 match_text doesn't produce debug output #345

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -875,7 +875,7 @@ class DeviceUnderTest(object):
         xml, region = _tesseract(frame, region, mode, lang, _config,
                                  user_words=text.split())
         if xml == '':
-            return TextMatchResult(ts, False, None, frame, text)
+            result = TextMatchResult(ts, False, None, frame, text)
         hocr = lxml.etree.fromstring(xml)
         p = _hocr_find_phrase(hocr, text.split())
         if p:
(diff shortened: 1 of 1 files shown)


drothlis:
Hi @rinaldomerlo, thanks for the patch!

You've missed one case [here](https://github.com/stb-tester/stb-tester/pull/346/files#diff-b8db1b2ea3c8e5386ca27ac571ea7aa6L878) where there's a return earlier in the function.

Otherwise it looks good to me! The travis failure is unrelated, it's a bug in our self-tests. 

Could you also amend the commit message so that the first line has a brief summary of the change? For example: `core.py: Add debug output to match_text`

rinaldomerlo:
Hi,

Hi @drothlis,

I didn't know what that earlier return did, so I left alone. Shall I just return a no match found for that? Or something neutral.

Rinaldo.

drothlis:
Presumably that case is when tesseract doesn't find any text at all, so you'd want to print example the same message as you do for the "No match found" case.

drothlis:
I just noticed that you're logging the search text instead of the complete TextMatchResult. I would prefer logging the whole TextMatchResult, in the same way that `match` logs the whole MatchResult. Something like this:

    result = TextMatchResult(ts, True, box, frame, text)
    debug("Match found: %s" % result)
    return result

and similarly for the other `return`s.

It might make sense to refactor the function so that there's a single return at the end, like [match] does.

[match]: https://github.com/stb-tester/stb-tester/blob/v24/_stbt/core.py#L630-L663

drothlis:
Superseded by #348.

PR #348, opened 2016-04-04
Add debug output to match_text for issue #345

rinaldomerlo:
core.py: Add debug output to match_text

refactored function to be more like match() as requested.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -892,7 +892,7 @@ class DeviceUnderTest(object):
             result = TextMatchResult(ts, True, box, frame, text)
         else:
             result = TextMatchResult(ts, False, None, frame, text)
-            
+
         if result.match:
             debug("Match found: %s" % str(result))
         else:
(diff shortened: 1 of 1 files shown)


_stbt/core.py
```diff
@@ -875,7 +875,7 @@ def match_text(self, text, frame=None, region=Region.ALL,
         xml, region = _tesseract(frame, region, mode, lang, _config,
                                  user_words=text.split())
         if xml == '':
-            return TextMatchResult(ts, False, None, frame, text)
+            result = TextMatchResult(ts, False, None, frame, text)
```
drothlis: This isn't *just* a refactoring -- it is also changing the behaviour. Now it will proceed to run the code below, and `lxml.etree.fromstring('')` will raise an XMLSyntaxError. (This is caught by one of our unit tests -- if you run `make check` you should see it fail.)

drothlis:
Hi @rinaldomerlo 

Sorry for the delay, I hadn't noticed your new commits. For some reason github doesn't send me email notifications for new commits, only for new comments.

I've squashed your commits into a single commit (466cd62) and added my own wording changes (b35f6d3) in #351.

Thanks for your patience!

Closing - superseded by #351.

PR #352, opened 2016-05-03
Refactor image debug logging to use ImageLogger class

drothlis:
Instead of writing debug images to disk directly, `match`, `detect_motion` and `is_screen_black` register the debug images with an instance of `ImageLogger`. Helper functions like `_find_match` are injected the `ImageLogger` instance (this is the observer pattern).

This is better because previously `_log_image` (which wrote images to disk) and `_log_image_descriptions` (which wrote an html file) communicated with each other via the filesystem -- `_log_image_descriptions` would list the files on disk to determine how many levels of pyramid-matching had happened. This would create incorrect html output if a previous invocation of `stbt run` had created 3 levels of debug screenshots in `stbt-debug/detect_match/00001/` but the current invocation had only created 1 level (because no match was found).

It's also less magical -- previously we incremented the frame counter when someone logged an image called "source.png". Now we increment the frame counter when you create a new instance of `ImageLogger`.

My motivation in doing this work is that I want to add additional logging for https://github.com/stb-tester/stb-tester/pull/212 so I want to simplify the code first.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -41,8 +41,6 @@ from _stbt.logging import ddebug, debug, warn
 gi.require_version("Gst", "1.0")
 from gi.repository import GLib, GObject, Gst  # isort:skip pylint: disable=E0611
 
-if getattr(gi, "version_info", (0, 0, 0)) < (3, 12, 0):
-    GObject.threads_init()
 Gst.init(None)
 
 warnings.filterwarnings(
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -13,7 +13,6 @@ from __future__ import absolute_import
 import argparse
 import datetime
 import functools
-import glob
 import inspect
 import os
 import Queue
@@ -687,13 +686,10 @@ class DeviceUnderTest(object):
             mask_image = _load_mask(mask)
 
         previous_frame_gray = None
-        log = functools.partial(
-            _log_image, directory="stbt-debug/detect_motion")
 
         for sample in self._display.gst_samples(timeout_secs):
             with numpy_from_sample(sample, readonly=True) as frame:
                 frame_gray = cv2.cvtColor(frame, cv2.COLOR_BGR2GRAY)
-            log(frame_gray, "source")
 
             if previous_frame_gray is None:
                 if (mask_image is not None and
@@ -705,14 +701,17 @@ class DeviceUnderTest(object):
                 previous_frame_gray = frame_gray
                 continue
 
+            imglog = logging.ImageLogger("detect_motion")
+            imglog.add("source", frame_gray)
+
             absdiff = cv2.absdiff(frame_gray, previous_frame_gray)
             previous_frame_gray = frame_gray
-            log(absdiff, "absdiff")
+            imglog.add("absdiff", absdiff)
 
             if mask_image is not None:
                 absdiff = cv2.bitwise_and(absdiff, mask_image)
-                log(mask_image, "mask")
-                log(absdiff, "absdiff_masked")
+                imglog.add("mask", mask_image)
+                imglog.add("absdiff_masked", absdiff)
 
             _, thresholded = cv2.threshold(
                 absdiff, int((1 - noise_threshold) * 255), 255,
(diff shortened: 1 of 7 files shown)


_stbt/gst_utils.py
```diff
@@ -1,13 +1,118 @@
...
+    s = Gst.Sample.new(Gst.Buffer.new_wrapped(
+        "row 1 4 px  row 2 4 px  row 3 4 px  "),
+        Gst.Caps.from_string("video/x-raw,format=BGR,width=4,height=3"),
+        None, None)
+    with numpy_from_sample(s, readonly=True) as a:
+        assert a.shape == (3, 4, 3)
+
+
+def gst_sample_make_writable(sample):
+    if sample.get_buffer().mini_object.is_writable():
```
wmanley: Huh, this will always make a copy because of https://bugzilla.gnome.org/show_bug.cgi?id=736896 .  I realise this isn't new in your commit.

_stbt/logging.py
```diff
@@ -69,6 +75,46 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+        self.frame_number = ImageLogger._frame_number
+        ImageLogger._frame_number += 1
+
+    def add(self, name, image):
+        if get_debug_level() <= 1:
+            return
+        if name in self.images:
+            raise ValueError("Image for name '%s' already logged" % name)
+        with numpy_from_sample(image, readonly=True) as img:
+            self.images[name] = img.copy()
```
wmanley: Why not write this image out immediately?

_stbt/gst_utils.py
drothlis: > Huh, this will always make a copy because of https://bugzilla.gnome.org/show_bug.cgi?id=736896 .

But only in python-gi >= 3.13 (Ubuntu > 14.04).

_stbt/core.py
```diff
@@ -2055,156 +1952,147 @@ def _confirm_match(image, position, template, match_parameters):
...
+        self._levels = set()
 
-def _log_image_descriptions(
-        template_name, matched, position,
-        first_pass_matched, first_pass_certainty, match_parameters):
-    """Create html file that describes the debug images."""
+    def add(self, name, image, pyramid_level=None):  # pylint:disable=arguments-differ
+        if pyramid_level is not None:
+            name = "level%d-%s" % (pyramid_level, name)
+            self._levels.add(pyramid_level)
```
wmanley: I can't see where this `pyramid_level` argument is used?

drothlis: It's used in:

```
    def write(self, ...):
        ... template.render(
            ... levels=list(reversed(sorted(self._levels)))
```

_stbt/core.py
```diff
@@ -2055,156 +1952,147 @@ def _confirm_match(image, position, template, match_parameters):
...
-        utils.mkdir_p(d)
-    except OSError:
-        warn("Failed to create directory '%s'; won't save debug images." % d)
-        return
-    with _numpy_from_sample(image, readonly=True) as img:
-        if img.dtype == numpy.float32:
-            img = cv2.convertScaleAbs(img, alpha=255)
-        cv2.imwrite(os.path.join(d, name) + ".png", img)
-
+class MatchImageLogger(logging.ImageLogger):
```
wmanley: Inheritance doesn't seem appropriate here.  `add` is overridden and `write_images` isn't used by `_match`.

_stbt/logging.py
drothlis: Because it will allow me to pull additional debugging out of `_match` and into `MatchImageLogger`. Specifically this block of code:

    if logging.get_debug_level() > 1:
        source_with_roi = image.copy()
        _draw_match(source_with_roi, region, first_pass_matched, thickness=1)
        imglog.add("source_with_roi", source_with_roi)

I had forgotten to do that, so I'll push a fixup shortly.

wmanley:
I agree that this is a functional improvement, avoiding implicit communication via the filesystem is definitely a good idea.  I'm less sure of the inheritance and OO side of things.  Clearly there's a concept of logging context here which is very valuable and is reasonably provided by an object. Hmm...

wmanley:
I wonder if we could more usefully separate the logging from the report generation.  I've been thinking about these ideas WRT live display of what stb-tester is upto.  Perhaps we could log with the Python logger: https://docs.python.org/2/howto/logging-cookbook.html#adding-contextual-information-to-your-logging-output and separately have a Handler that will spit out the HTML.

drothlis:
> Perhaps we could log with the Python logger

I think that's out of scope for this pull request as I really want to get on with fixing `match_all` (#212). I did leave the function for generating the `match` debug as a function, instead of a subclass of `ImageLogger` (which is what I'd done in an earlier version of this pull request) inspired by your suggestion to make it more dict-like.

Merging.

PR #353, opened 2016-05-03
WIP: stbt.wait_for_motion, stbt.detect_motion: Add region parameter

drothlis:
Passing a Region is more convenient than making a mask image for the simple cases where a single rectangular region is sufficient.

TODO:
- [x] Merge #352 as I don't want conflicts with the logging (and I might want to draw the region on the debug images?)
- [ ] Implement it
- [ ] API documentation in `stbt/__init__.py`
- [ ] Tests
- [ ] Also for `is_screen_black`?

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -675,7 +675,12 @@ class DeviceUnderTest(object):
             self._display.draw(result, None)
             yield result
 
-    def detect_motion(self, timeout_secs=10, noise_threshold=None, mask=None):
+    def detect_motion(self, timeout_secs=10, noise_threshold=None, mask=None,
+                      region=Region.ALL):
+
+        if mask is not None and region != Region.ALL:
+            raise ValueError("You can't specify both 'mask' and 'region'")
+
         if noise_threshold is None:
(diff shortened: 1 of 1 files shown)


drothlis:
I think a `region` parameter could be confusing. It should be the region to *include* (for consistency with `stbt.match`, and because the use-case that prompted this pull request is to check a Picture-in-Picture for motion) but a more common use-case is to *exclude* some regions (where the loading spinner appears, or where an info overlay can appear/disappear).

A more composable solution would be to change the existing `mask` parameter to accept a numpy array (currently it only accepts filenames) and provide some helper functions to create a mask from a `stbt.Region`. Perhaps `mask_in(region)` and `mask_out(region)`.

drothlis:
Closing. I've raised #480 instead to accept numpy images as masks. We can construct the mask using a stand-alone function as per my previous comment.

drothlis:
We've decided to implement this after all in #484, because it's more performant which may be important if you're doing lots of `wait_for_motion`s in parallel threads.

PR #354, opened 2016-05-06
stbt auto-selftest: Add caching of image processing result for speedup

wmanley:
This makes running `stbt auto-selftest generate` practical on large test-packs with many selftest screenshots.  On a private repo with 196 selftest-screenshots and 73 expressions (14308 doctest items to evaluate) the results are rather promising:

* 1:01:13 without caching, median expression time 120ms
* 37:50 with a cold cache, median expression time 38ms (1.6x speed up)
* 2:20 total with a hot cache.  Median time per expression: 3.4ms (26x speed up)

This is still a little slower than I'd like, but certainly makes `stbt auto-selftest` usable on large test-packs where it just wasn't fast enough before.  Further improvements can come in future PRs.  I'll need to profile first.  Possible speedup: Avoid the need to load the selftest images from disk (up to 17*3=51s saving)

In this instance the cache file weighs in at 8.2MB with 22843 entries (361B/entry).

We use the xxhash library for fast (non-crypto) image hashing.  I evaluated sha1 and python's `hash()` builtin and they were too slow taking >3ms for a 720p image.  xxhash does the same in 240us.

We use lmdb as the on-disk cache database because it's small, fast and has good Python bindings.

TODO:

* [x] Measure performance with caching disabled and with a cold cache
* [x] Some tests
* [x] Check Fedora package building
* [x] Bundle py-lmdb as it's not available in Ubuntu 14.04
* [x] Decide how to distribute xxhash and pylmdb?

    * ~~Include in this repo?~~
    * **Submodules**
    * ~~Download as part of build process?~~
* [x] Think about hashing "serialisation".  It's currently rather ugly and not very robust.  Perhaps use json instead with a custom `JSONEncoder` that knows about `numpy.ndarray` and `Gst.Sample`?
* [x] Mention versions of submodules in commit messages


wmanley:
I've squashed and broken some of the miscellaneous changes out into #355.

wmanley:
@drothlis: Please review

wmanley:
> @drothlis: Please review

Sorry not this one.  I meant #355

wmanley:
All TODO items done, @drothlis: Please review

_stbt/core.py
```diff
@@ -36,7 +36,7 @@
 from _stbt.config import ConfigurationError, get_config
 from _stbt.gst_hacks import gst_iterate
 from _stbt.gst_utils import (get_frame_timestamp, gst_sample_make_writable,
-                             numpy_from_sample)
+                             numpy_from_sample, sample_shape)
```
drothlis: This belongs in the previous commit.

_stbt/core.py
```diff
@@ -1741,7 +1742,7 @@ def _match(image, template, match_parameters, imglog):
     region = Region(position.x, position.y,
                     template.shape[1], template.shape[0])
 
-    return matched, region, first_pass_certainty
+    return matched, list(region), first_pass_certainty
```
drothlis: I suppose it's too inconvenient to teach the json encoder how to serialize/deserialize `Region` objects... anyway if we expose the caching mechanism to user-defined image-processing functions then they wouldn't be able to extend the json encoder/decoder either... so I suppose it makes sense to stick to the default objects support by  `json.JSONEncoder`.

Makefile
```diff
@@ -222,8 +237,10 @@ stb-tester-$(VERSION).tar.gz: $(DIST)
 	    exit 1; }
 	# Separate tar and gzip so we can pass "-n" for more deterministic
 	# tarball generation
+	SUBMODULE_DIST=$$( \
+	    git submodule foreach -q 'git ls-files | sed s,^,$$path/,') && \
```
drothlis: Where is this `$path` defined?

drothlis: I see, it's set by `git submodule foreach`.

vendor/xxHash
```diff
@@ -0,0 +1 @@
+Subproject commit fcabf31f4e0812416a6759b6fb8716c165566470
```
drothlis: It would be worth mentioning in the commit message that this is tag v0.5.1 (?).

Makefile
```diff
@@ -190,7 +192,14 @@ check-bashcompletion:
...
 
-SUBMODULE_FILES =
+XXHASH_SOURCES = \
+    vendor/xxHash/xxhash.c \
+    vendor/xxHash/xxhash.h
+
+_stbt/libxxhash.so : $(XXHASH_SOURCES)
+	$(CC) -shared -fPIC -O3 -o $@ $(XXHASH_SOURCES)
+
+SUBMODULE_FILES = $(XXHASH_SOURCES)
```
drothlis: Nice & simple, I like it.

drothlis:
The commit message of the last commit needs to be updated with your timings instead of "XXX".

_stbt/imgproc_cache.py
```diff
@@ -0,0 +1,211 @@
...
+
+@contextmanager
+def cache(filename=None):
+    if os.environ.get('STBT_DISABLE_CACHING'):
+        yield
+        return
+
+    global _cache
+
+    sys.path.append(os.path.dirname(__file__))
```
drothlis: Is this `sys.path.append` necessary? We have `from _stbt.gst_utils import ...` above, so if that works why wouldn't this work?

_stbt/imgproc_cache.py
```diff
@@ -0,0 +1,211 @@
...
+    import _stbt.lmdb as lmdb
+    del sys.path[-1]
+
+    if filename is None:
+        cache_home = os.environ.get('XDG_CACHE_HOME') \
+            or '%s/.cache' % os.environ['HOME']
+        mkdir_p(cache_home + "/stbt")
+        filename = cache_home + "/stbt/cache.lmdb"
+    with lmdb.open(filename) as db:  # pylint: disable=no-member
+        assert _cache is None
```
drothlis: This assert could do with an error message, like "Cache cannot be used concurrently"?

_stbt/core.py
wmanley: Fixed

vendor/xxHash
wmanley: Done

wmanley:
> The commit message of the last commit needs to be updated with your timings instead of "XXX".

Done

_stbt/imgproc_cache.py
wmanley: Because the lmdb module itself calls `import lmdb` which fails without this.

_stbt/imgproc_cache.py
wmanley: This is a genuine assertion, and the check for use is racy so the throwing behaviour shouldn't be relied upon.

wmanley: I'm happy to add a message.  "concurrently" isn't quite the right term though as this will happen if you do:

    with cache():
        with cache():
            ...

drothlis: Oh I see. Nevermind then.

drothlis:
This is missing a "version" field in the cache key, so that we can invalidate the cache when/if we change the implementation of `match`. (Or perhaps we can implement that later when/if it happens? We would certainly need it by the time we made `imgproc_cache.memoize` a public API.)

drothlis:
The tesseract caching should take the tesseract version into account.

drothlis:
Perhaps `imgproc_cache.memoize` could take arbitrary keyword arguments that are added to the cache key.

drothlis:
I've finished my review. Looks great apart from the above comments.

wmanley:
> Perhaps `imgproc_cache.memoize` could take arbitrary keyword arguments that are added to the cache key.

Yes, that's a very nice design.  I'm implementing that now.

drothlis:
Consider adding this to the last commit's message, or to the imgproc_cache.py docstring, or both:

```
To enable caching, decorate the cachable function with
`imgproc_cache.memoize` and call the function within the scope of a
`with imgproc_cache.cache():` context manager. For now this is a private
API but we intend to make it public at some point so that you can add
caching to any custom image-processing functions in your own
FrameObjects.
```

_stbt/imgproc_cache.py
```diff
@@ -0,0 +1,211 @@
...
+def _cache_hash(value):
+    from _stbt.xxhash import Xxhash64
+    h = Xxhash64()
+
+    class HashWriter(object):
+        def write(self, data):
+            h.update(data)
+            return len(data)
+
+    json.dump(value, HashWriter(), cls=_ArgsEncoder)
```
wmanley: And another issue: we need to have `sort_keys=True` here to make it deterministic.

wmanley:
> Consider adding this to the last commit's message, or to the imgproc_cache.py docstring, or both:

I've added a bunch of documentation to `memoize`.

wmanley:
>> Perhaps imgproc_cache.memoize could take arbitrary keyword arguments that are added to the cache key.
>
> Yes, that's a very nice design. I'm implementing that now.

Done

wmanley:
Came across another issue.  From the [lmdb documentation](https://lmdb.readthedocs.io/en/release/#lmdb.Environment):

> **map_size:**
> Maximum size database may grow to; used to size the memory mapping. If database grows larger than map_size, an exception will be raised and the user must close and reopen Environment. On 64-bit there is no penalty for making this huge (say 1TB). Must be <2GB on 32-bit.
> 
> > **Note**
> >
> > The default map size is set low to encourage a crash, so users can figure out a good value before learning about this option too late.

I was getting `lmdb.MapFullError`s thrown.  I've bumped the map size up to 1GB and made it so `MapFullError` is non-fatal.

wmanley:
Squashed, not just waiting for Travis before merging.

PR #355, opened 2016-05-09
Auto selftest improvements

wmanley:
3 Miscellaneous improvements split out of #354

### DIFF
--- a/stbt_auto_selftest.py
+++ b/stbt_auto_selftest.py
@@ -147,6 +147,21 @@ def init_worker():
     signal.signal(signal.SIGINT, signal.SIG_IGN)
 
 
+def iterate_with_progress(sequence, width=20, stream=sys.stderr):
+    ANSI_ERASE_LINE = '\033[K'
+    stream.write('\n')
+    total = len(sequence)
+    for n, v in enumerate(sequence):
+        if n == total:
+            break
+        progress = (n * width) // total
+        stream.write(
+            ANSI_ERASE_LINE + '[%s] %8d / %d - Processing %s\r' % (
+                '#' * progress + ' ' * (width - progress), n, total, str(v)))
+        yield v
+    stream.write('\n')
+
+
 def generate_into_tmpdir():
     start_time = time.time()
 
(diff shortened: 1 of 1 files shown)


wmanley:
@drothlis: Please review

wmanley:
CI failed because of virtual-stb tests.  I'll have to have a look at them  It could just be that we need a bigger timeout on how long the vstb takes to start.

stbt_auto_selftest.py
```diff
@@ -147,6 +147,20 @@ def init_worker():
     signal.signal(signal.SIGINT, signal.SIG_IGN)
 
 
+def iterate_with_progress(iterator, width=20, stream=sys.stderr):
```
drothlis: Nitpick on the naming: an iterator, technically, only supports `next()`, not `len()`. Python's terminology for this seems to be "sequence": https://docs.python.org/2/library/functions.html#len

stbt_auto_selftest.py
```diff
@@ -147,6 +147,20 @@ def init_worker():
...
+    stream.write('\n')
+    total = len(iterator)
+    for n, v in enumerate(iterator):
+        if n == total:
+            break
+        progress = (n * width) // total
+        stream.write(
+            '\033[K[%s] %8d / %d - Processing %s\r' % (
```
drothlis: What does `\033[K` do?

stbt_auto_selftest.py
```diff
@@ -147,6 +147,20 @@ def init_worker():
...
+                '#' * progress + ' ' * (width - progress), n, total, str(v)))
+        yield v
+    stream.write('\r' + ' ' * (total + 28) + '\r')
```
drothlis: Is this supposed to erase the progressbar after we're done? If so:

* The `28` won't overwrite the `str(v)` at the end of the line.
* Shouldn't it be `' ' * width` rather than `' ' * total`?
* Perhaps it would be easier to leave the progressbar up and write a '\n' at the end.

stbt_auto_selftest.py
wmanley: Fixed.

stbt_auto_selftest.py
wmanley: It's [ANSI] Erase Line.  I've made this clearer in my fixup.

[ANSI]: https://en.wikipedia.org/wiki/ANSI_escape_code

stbt_auto_selftest.py
wmanley: > Perhaps it would be easier to leave the progressbar up and write a '\n' at the end.
￼
Yes, and probably clearer too.  I've done this.

drothlis:
+1.

wmanley:
Squashed.  I'll merge once CI is happy.

PR #357, opened 2016-05-10
Fix virtual-stb intermittant test-failure on Travis

wmanley:
test_that_virtual_stb_configures_stb_tester_for_testing_virtual_stbs fails
intermittently on Travis because sometimes chrome takes longer than 10s to
start-up.  This causes the test to fail with:

> MatchTimeout: Didn't find match for '.../stb-tester-350px.png' within 10
> seconds

This commit should fix that issue.

TODO:

- [ ] Re-run the test a few times to see if it is actually fixed.

### DIFF
--- a/tests/vstb-example-html5/tests/rotate.py
+++ b/tests/vstb-example-html5/tests/rotate.py
@@ -3,7 +3,7 @@ from stbt import press, wait_for_match
 
 
 def wait_for_vstb_startup():
-    wait_for_match('stb-tester-350px.png')
+    wait_for_match('stb-tester-350px.png', timeout_secs=20)
 
 
 def test_that_image_is_rotated_by_arrows():


wmanley:
1 success, restarting

wmanley:
2 successes, restarting

wmanley:
The last test run failed, but not because of virtual-stb test:

    test_stbt_batch_run_sigterm_twice... FAIL
    Showing '/tmp/stb-tester.feK/log':
    tests/test.py ... 
    Received interrupt; waiting for current test to complete.
    Received interrupt; exiting.
    FAILED
    /home/travis/build/stb-tester/stb-tester/tests/test-install/libexec/stbt/stbt-batch.d/run-one: line 95: 1462879347 -  : syntax error: operand expected (error token is "-  ")
    diff: latest/failure-reason: No such file or directory
    error: Bad failure-reason

I've not seen this one before, rerunning

wmanley:
And another pass.  That's 4 passes of `test_that_virtual_stb_configures_stb_tester_for_testing_virtual_stbs`.  Seems to be fixed.

PR #358, opened 2016-05-10
Rename "stbt templatematch" command to "stbt match"

drothlis:
For consistency with the Python API "stbt.match". I've been trying to
remove the wording "templatematch" from our public APIs as it is the
name of the OpenCV function we use in the first pass of our matching
algorithm; as such it is an implementation detail. The argument names to
`stbt.match`, `stbt.wait_for_match` etc are "image" (not "template") and
in prose we call this a "reference image" rather than a "template".

For backwards compatibility we keep "stbt templatematch" as an alias.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,11 +95,11 @@ INSTALL_CORE_FILES = \
     stbt-config \
     stbt-control \
     stbt-lint \
+    stbt-match \
     stbt-power \
     stbt-record \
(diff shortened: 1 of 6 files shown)


drothlis:
Just a small tidy-up that I'd like to see in v25.

wmanley:
+1 seems reasonable.

PR #361, opened 2016-05-11
ImageLogger refactorings in preparation for implementing `match_all`

drothlis:
Extracted from #212 and rebased on top of master to resolve conflicts from #354.

### DIFF
--- a/_stbt/logging.py
+++ b/_stbt/logging.py
@@ -82,7 +82,7 @@ class ImageLogger(object):
     """
     _frame_number = 1
 
-    def __init__(self, name):
+    def __init__(self, name, **kwargs):
         self.enabled = get_debug_level() > 1
         if not self.enabled:
             return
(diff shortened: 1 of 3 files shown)


_stbt/logging.py
```diff
@@ -114,12 +114,17 @@ def set(self, **kwargs):
...
             self.data[k] = v
 
-    def imwrite(self, name, image, pyramid_level=None):
+    def append(self, **kwargs):
+        if not self.enabled:
+            return
+        for k, v in kwargs.iteritems():
+            if k not in self.data:
+                self.data[k] = []
+            self.data[k].append(v)
```
wmanley: Not important, but I've had a few ideas how this could be different.

Code sketch:

```

class ImageLogger(object):
    def __init__(self, log=None):
        if log is None:
            log = []
        self._log = list(log)

    def log(self, **kwargs):
        self._log.extend(kwargs.items())

    def get(self, name):
        for k, v in self._log:
            if k == name:
                return v
            
    def filter(self, **kwargs):
        newlog = []
        name, value = kwargs.items()[0]
        it = iter(self._log)
        for k, v in it:
            if k == name and v == value:
                newlog.append((k, v))
                break

        for k, v in it:
            if k == name:
                break
            newlog.append((k, v))
        return ImageLogger(newlog)

    def iter_filter(self, name):
        # Whoops, O(n^2)
        for k, v in self._log:
            if k == name:
                yield self.filter(**{k: v})


>>> l = ImageLogger()

>>> l.log(name="David")
>>> l.log(surname="Rothlisberger")

>>> l.log(name="William")
>>> l.log(surname="Manley")

>>> l.get("name")
'David'

>>> l.get("surname")
'Rothlisberger'

>>> f = l.filter(name="William")
>>> f.get("surname")
'Manley'
```

wmanley:
I'd already reviewed these on #354 but I've given them a second once-over.  Pretty uncontraversial small improvements.  Merging

PR #362, opened 2016-05-12
FrameObject: Anticipate stbt.get_frame returning a read-only image

wmanley:
In the future I want to make `stbt.get_frame()` return an immutable image.
This way we will be able to reduce copying of data by providing numpy
arrays backed by the original `Gst.Sample`s.  This will be particularly
important for better and more efficient handling of 4K video.

This commit anticipates that change by exercising user's `FrameObject`s
with read-only buffers.  This is good because it makes a future change to
the behaviour of `get_frame()` less likely to break user-scripts.  This is
a non-breaking change itself as `FrameObject` has yet to be released.

### DIFF
--- a/stbt/__init__.py
+++ b/stbt/__init__.py
@@ -715,6 +715,9 @@ class FrameObject(_stbt.core.FrameObject):
     def __init__(self, frame=None):
         if frame is None:
             frame = _dut.get_frame()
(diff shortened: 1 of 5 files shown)


wmanley:
@drothlis: Please review.  I think this is fairly uncontroversial.

PR #363, opened 2016-05-16
Document timestamp

drothlis:


stbt/__init__.py
```diff
@@ -424,21 +424,25 @@ def frames(timeout_secs=None):
       iterating (for example with ``break``) at any time.
 
     :returns:
-      An ``(image, timestamp)`` tuple for each video frame, where ``image`` is
-      a `numpy.ndarray` object (that is, an OpenCV image).
+      An ``(image, timestamp)`` tuple for each video frame:
+
+      * ``image`` is a `numpy.ndarray` object (that is, an OpenCV image).
+      * ``timestamp`` is the stream time in nanoseconds. The absolute value
```
wmanley: Technically it's not necessilary a GStreamer stream time, nor a running time, it's the PTS attached to the `GstBuffer`s.  This can't be interpreted in the general case, apart from looking a differences between timestamps.  Perhaps we should be more vague:

> timestamp is a timestamp in nanoseconds.  The absolute value isn't related to the system time; what's useful is the relative difference between frames.

drothlis: Good idea. Done.

Issue #247, opened 2015-01-06
Unpredictable timeout if source pipeline restarts during wait_for_match


szemate:
Every time the source pipeline is rebuilt, timestamps of the video buffers restart counting from zero.

If the source pipeline restarts while `gst_samples` is running, the actual timeout value is getting unpredictable, as `gst_samples` times out if `timestamp - self.start_timestamp > timeout_secs`, and both `timestamp` and `self.start_timestamp` are reset to 0 while `timeout_secs` stays the same. (https://github.com/drothlis/stb-tester/blob/master/stbt/__init__.py#L1749)

E.g. if we are running `wait_for_match('something.png', timeout_secs=20)` and the source pipeline restarts after 10 seconds, it raises `MatchTimeout` after ~30 seconds.

This issue replaces (part of) #214.

wmanley:
We really shouldn't be using buffer timestamps directly in stbt.  Technically-speaking they're meaningless without a `GstSegment` to interpret them into running-time or stream-time.

Perhaps we can start dealing with stream-time and manipulate GStreamer to set stream-time to wall-clock time when we create the pipeline in the first-place.  That way stream time would start being more meaningful than "time since we started capturing video".  I don't know what effect that would have on the sink-pipeline.  It would probably be fine.

wmanley:
Hmm, maybe setting stream-time to wall-clock time is meaningless as we can send a SEGMENT event in PAUSED state and you'd expect the next buffer out to be timestamped with the time in the SEGMENT event?  Not sure.

PR #364, opened 2016-05-22
Realtime timestamps

wmanley:
The timestamps we currently provide are the PTS data attached to the
`Gst.Sample`s pulled from the source pipeline.  In practice with v4l2src
this will have tended to roughly be the time from the start of the test or
the last pipeline restart.  It could be some other number depending on the
source pipeline used.   It is only useful for comparing the time
difference between two frames.

This commit introduces `sample.realtime_timestamp` for each pulled frame.
This is the time since the unix epoch in seconds as a float.  This can be
used to compare the time of a frame against:

* The current time (for debugging)
* Other events that happen on the system such as IR presses for more-precise latency measurements
* Other events that occur in the world such as remote server timestamps or for measuring synchronisation between different devices-under-test (assuming a correctly set system-clock).

This could help solve #247 too.

TODO:

- [x] Work out an API - perhaps attaching the data as an additional member to `ndarray` would be best.  See http://stackoverflow.com/a/34967782/65031 for how
- [x] Merge #379 
- [x] Add test for `MotionResult` repr
- [x] Add test that provided timestamps are in wall-clock time
- [x] Add release note for new timestamps
- [x] Document `*.time`
- [x] Clarify what `MotionResult.time` refers to
- [x] Clarify documentation.  `{Match,TextMatch}Result.frame` isn't necessarily a `Frame`

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1484,7 +1484,7 @@ class Display(object):
         self.novideo = False
         self.lock = threading.RLock()  # Held by whoever is consuming frames
         self.last_sample = Queue.Queue(maxsize=1)
-        self.last_used_sample = None
+        self.last_used_frame = None
         self.source_pipeline = None
         self.start_timestamp = None
         self.underrun_timeout = None
@@ -1610,7 +1610,7 @@ class Display(object):
             raise UITestError(str(gst_sample))
 
         if isinstance(gst_sample, Gst.Sample):
-            self.last_used_sample = array_from_sample(gst_sample)
+            self.last_used_frame = array_from_sample(gst_sample)
 
         return gst_sample
 
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -36,9 +36,9 @@ from kitchen.text.converters import to_bytes
 
 from _stbt import imgproc_cache, logging, utils
 from _stbt.config import ConfigurationError, get_config
-from _stbt.gst_utils import (array_from_sample, get_frame_timestamp,
-                             gst_iterate, gst_sample_make_writable,
-                             sample_shape)
+from _stbt.gst_utils import (array_from_sample, CapturedFrame,
+                             get_frame_timestamp, gst_iterate,
+                             gst_sample_make_writable, sample_shape)
 from _stbt.logging import ddebug, debug, warn
 
 gi.require_version("Gst", "1.0")
@@ -2615,7 +2615,6 @@ def test_wait_for_motion_half_motion_int():
 def _fake_frames_at_half_motion():
     class FakeDisplay(object):
         def frames(self, _timeout_secs=10):
-            from _stbt.gst_utils import CapturedFrame
             data = [
                 numpy.zeros((2, 2, 3), dtype=numpy.uint8),
                 numpy.ones((2, 2, 3), dtype=numpy.uint8) * 255,
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1481,11 +1481,15 @@ class Display(object):
                  mainloop, save_video="",
                  restart_source=False,
                  transformation_pipeline='identity'):
+
+        import time
+
         self.novideo = False
         self.lock = threading.RLock()  # Held by whoever is consuming frames
         self.last_sample = Queue.Queue(maxsize=1)
         self.last_used_frame = None
         self.source_pipeline = None
+        self.init_time = time.time()
         self.start_timestamp = None
         self.underrun_timeout = None
         self.tearing_down = False
@@ -1648,6 +1652,11 @@ class Display(object):
             Gst.Format.TIME, sample.get_buffer().pts)
         sample.time = (
             float(appsink.base_time + running_time) / 1e9)
+
+        if (sample.time > self.init_time + 31536000 or
+                sample.time < self.init_time - 31536000):  # 1 year
+            warn("Received frame with suspicious timestamp: %f. Check your "
+                 "source-pipeline configuration." % sample.time)
 
         self.tell_user_thread(sample)
         if self.lock.acquire(False):  # non-blocking
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -430,26 +430,13 @@ class MatchResult(object):
     """
     # pylint: disable=W0621
     def __init__(
-            self, time, match, region, first_pass_result, frame=None,
-            image=None, timestamp=None, _first_pass_matched=None):
+            self, time, match, region, first_pass_result, frame, image,
+            timestamp=None, _first_pass_matched=None):
         self.timestamp = timestamp
         self.match = match
         self.region = region
         self.first_pass_result = first_pass_result
-        if frame is None:
-            warnings.warn(
-                "Creating a 'MatchResult' without specifying 'frame' is "
-                "deprecated. In a future release of stb-tester the 'frame' "
-                "parameter will be mandatory.",
-                DeprecationWarning, stacklevel=2)
         self.frame = frame
-        if image is None:
-            warnings.warn(
-                "Creating a 'MatchResult' without specifying 'image' is "
-                "deprecated. In a future release of stb-tester the 'image' "
-                "parameter will be mandatory.",
-                DeprecationWarning, stacklevel=2)
-            image = ""
         self.image = image
         self.time = time
         self._first_pass_matched = _first_pass_matched
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -954,11 +954,21 @@ class DeviceUnderTest(object):
             motion_frames, considered_frames))
 
         matches = deque(maxlen=considered_frames)
+        motion_count = 0
         for res in self.detect_motion(timeout_secs, noise_threshold, mask):
-            matches.append(res.motion)
-            if matches.count(True) >= motion_frames:
+            motion_count += bool(res)
+            if len(matches) == matches.maxlen:
+                motion_count -= bool(matches.popleft())
+            matches.append(res)
+            if motion_count >= motion_frames:
                 debug("Motion detected.")
-                return res
+                # We want to return the first True motion result as this is when
+                # the motion actually started.
+                for result in matches:
+                    if result:
+                        return result
+                assert False, ("Logic error in wait_for_motion: This code "
+                               "should never be reached")
 
         screenshot = self.get_frame()
         raise MotionTimeout(screenshot, mask, timeout_secs)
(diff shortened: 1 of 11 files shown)


wmanley:
With this we could enhance `wait_for_match()`, `get_frame()`, etc. to discard frames before the point that `wait_for_match()` was called.  This way, even if you have high input video latencies we wouldn't be looking at old frames.

wmanley:
Notes on API:

* Currently we expose timestamps in:

    1. `get_frames()` yields timestamp, frame tuples
    2. `MatchResult` and `TextMatchResult` includes a timestamp

* These timestamps are integer nanoseconds taken from the buffer timestamps (usually equivalent to the total time that the test has been running for)

* This isn't very ergonomic to use.  It requires `/1e9` to turn it into seconds and is not directly comparable to the output of `time.time()`.

We could:

1. Replace all timestamps with nanoseconds since the unix epoch.  This is a minor break as time since start of test is not that useful as an absolute number anyway **or**

2. Replace all timestamps with float seconds a la `time.time()`.  This would break existing correct users of these timestamps as users would have to remove their `/1e9` code everywhere **or**

3. Add additional members to `MatchResult` and `TextMatchResult` with the real-time timestamp.  This would be compatible but could cause some confusion as there would then be two timestamp properties.  How would this be named?

    * `result.time` or
    * `result.realtime` or
    * `result.capture_time`?

wmanley:
After discussing with @drothlis we've decided to rename `realtime_timestamp` to just `time`.  This is good because it's shorter and is reminiscent of `time.time()`.

wmanley:
> Clarify what MotionResult.time refers to

This is a little tricky.  Here's some examples in which I would argue the motion starts at time 5.  I can't think off the top of my head of an algorithm that will return 5 in each case though:

**Example 1:**

    time:  1 2 3 4 5 6 7 8 9
    frame: A A A A B C D E F

**Example 2:**

    time:  5 6 7 8 9
    frame: B C D E F

**Example 3:**

    time:  4 5 6 7 8 9
    frame: A B C D E F

And then we need to add the complexity of `consecutive_frames`.

wmanley:
I'll still need to do some end-to-end latency measurements using [latency-clock] with hardware in the loop.  It may be that we will need to apply a hardware-dependent offset to each timestamp.

[latency-clock]: https://github.com/stb-tester/latency-clock

drothlis:
> Clarify what MotionResult.time refers to

Motion isn't really useful for precise timing measurements (such as how long does it take for content to start playing) because the content might start with a black screen that slowly fades in, or a static landscape shot, etc... so it doesn't seem that important to me. I'd say just return the time of frame "B" in all 3 of your examples.

wmanley:
> I'd say just return the time of frame "B" in all 3 of your examples.

Yes, but how to do that in example 2?

drothlis:
Oops, I meant frame "C" in example 2.

wmanley:
> Oops, I meant frame "C" in example 2.

I think I can get on-board with this.  The first frame that `wait_for_motion` sees will probably be from the past anyway as it will be whatever stbt has buffered, so the second frame seems appropriate.

wmanley:
Implemented.  I'm fairly happy with this now. @drothlis: please review

wmanley:
Possible future work:

* Have is_screen_black return a `BlackScreenResult` with a `time` attribute
* Have `press` return the time at which the IR signal was sent (perhaps via a `PressResult`?)

_stbt/gst_utils.py
```diff
@@ -89,8 +89,11 @@ def __del__(self):
 
 class CapturedFrame(numpy.ndarray):
```
drothlis: I'd like to call this `Frame`.

Does it still belong in gst_utils?

_stbt/gst_utils.py
```diff
@@ -93,23 +93,34 @@ class CapturedFrame(numpy.ndarray):
...
+
+    * `time` (float) - the wall-clock time that this video-frame was captured
+      as number of seconds since the unix epoch (1970-01-01T00:00:00Z).  This
+      is the same format as returned from the Python standard library function
+      `time.time()`.
+
     Added in stb-tester v26.
     """
-    def __new__(cls, array, dtype=None, order=None, _gst_pts=None):
+    def __new__(cls, array, dtype=None, order=None, _gst_pts=None, time=None):
```
drothlis: We're not getting rid of `_gst_pts`?

drothlis:
There's no release note for the new `time` parameter of `*Result`.

drothlis:
No discernible performance impact with these changes. Using `tests/run-performance-test.sh`:

* realtime-timestamps: 49.83 fps (501 frames in 10.05s)
* v25: 49.83 fps (501 frames in 10.05s)
* v24: 49.82 fps (501 frames in 10.06s)

wmanley:
> core: Use wall-clock timestamp everywhere instead of pts

Yeah, this is probably a good idea.  It's certainly a lot tidier.

_stbt/core.py
```diff
@@ -1615,16 +1625,14 @@ def frames(self, timeout_secs=None):
                 ddebug("user thread: Getting sample at %s" % time.time())
```
wmanley: At some point in the future we could do `self.start_timestamp = time.time()`.

_stbt/core.py
```diff
@@ -412,20 +412,20 @@ class MatchResult(object):
...
+      system time (`time.time()`).
     * ``match``: Boolean result, the same as evaluating `MatchResult` as a bool.
       That is, ``if match_result:`` will behave the same as
       ``if match_result.match:``.
     * ``region``: The `Region` in the video frame where the image was found.
     * ``first_pass_result``: Value between 0 (poor) and 1.0 (excellent match)
       from the first pass of stb-tester's two-pass image matching algorithm
       (see `MatchParameters` for details).
-    * ``frame``: The video frame that was searched, in OpenCV format.
+    * ``frame`` (`Frame`): The video frame that was searched.
```
wmanley: `frame` is not necessarily a `Frame` if the user has passed in a frame they've loaded themselves.

docs/release-notes.md
```diff
@@ -47,11 +59,14 @@ UNRELEASED
 * The debug log output of `match`, `wait_for_match` and `match_text` shows the
   matching region as (x, y, right, bottom) instead of (x, y, width, height).
 
-##### Developer-visible changes
+##### Maintainer-visible changes
```
wmanley: I like this

wmanley:
> Rename stbt.CapturedFrame => stbt.Frame

I'm not sure that `Frame` is a good name.  We have images:

1. Frames captured from the video-capture device.  Has a `.time`.  Possibly in the future has `.audio` as well
2. Frames saved to disk and loaded later. Currently a `numpy.ndarray` possibly in the future we would want `.filename` and `.png_git_sha` members
3. Templates and masks.  Currently `numpy.ndarray`.  Possibly in the future we would want `.filename` and `.png_git_sha` members and to replace `_AnnotatedTemplate` with this.

`Frame` covers the first two.  They're more similar WRT purpose but the second two are more similar WRT form.  I don't think it makes sense to call (3) `Frame`, but I do think it makes sense that `2` and `3` are the same type.

wmanley:
Calling it `stbt.Image` could be an alternative, although that would jar with having `.audio` attached.

drothlis:
I think we could create a `stbt.Image` in the future, for cases 2 & 3, when we need it. I think we should keep `stbt.Frame` for case 1, i.e. when you have a timestamp and audio.

drothlis:
Fixed the MatchResult.frame documentation, squashed & force-pushed.

drothlis:
Hmmm something's not quite right. [test_detect_motion_visualisation] fails intermittently, and sometimes it hangs. It does pass sometimes. On my laptop the most usual case is that it hangs.

When it fails, it's because it times out immediately:

    stbt-run: timed out: 19913540806.395 - 1466796732.685 > 10.000

(The first timestamp is somewhere in the year 2601. The second timestamp is today, just now.)

When it hangs, only the `stbt run ... verify.py` process is still running. In the log it has raised `NoVideo` but it hasn't exited. Attaching gdb we see this:

```
(gdb) i threads 
  Id   Target Id         Frame 
* 1    Thread 0x7fb27c57b700 (LWP 8399) "python" __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
  2    Thread 0x7fb25dbfa700 (LWP 8436) "_stbt_user_data" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  3    Thread 0x7fb2425fb700 (LWP 8433) "python" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  4    Thread 0x7fb2429fc700 (LWP 8432) "python" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  5    Thread 0x7fb242dfd700 (LWP 8431) "python" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  6    Thread 0x7fb2437fe700 (LWP 8430) "_stbt_raw_frame" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  7    Thread 0x7fb243fff700 (LWP 8425) "python" 0x00007fb27c179cbd in open64 () at ../sysdeps/unix/syscall-template.S:84
  8    Thread 0x7fb250ef9700 (LWP 8423) "appsrc:src" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  9    Thread 0x7fb2516fa700 (LWP 8422) "queue2:src" syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  10   Thread 0x7fb260fe2700 (LWP 8400) "python" 0x00007fb27be9ae8d in poll () at ../sysdeps/unix/syscall-template.S:84
(gdb) py-bt
Traceback (most recent call first):
  File "/tmp/stbt-test-installation.PmhIlX/libexec/stbt/_stbt/core.py", line 1817, in teardown
    elem.send_event(Gst.Event.new_eos())  # pylint: disable=E1120
  File "/tmp/stbt-test-installation.PmhIlX/libexec/stbt/_stbt/core.py", line 700, in __exit__
    self._display.teardown()
  File "/tmp/stbt-test-installation.PmhIlX/libexec/stbt/stbt/__init__.py", line 782, in teardown_run
    _dut.__exit__(None, None, None)
  File "/tmp/stbt-test-installation.PmhIlX/libexec/stbt/stbt-run", line 154, in <module>
    stbt.teardown_run()
```

[test_detect_motion_visualisation]: https://github.com/stb-tester/stb-tester/blob/1b9301d/tests/test-motion.sh#L213

drothlis:
git bisect says that this was introduced in 9c110be (core: Use wall-clock timestamp everywhere instead of pts), as I expected.

drothlis:
The source pipeline in `test_detect_motion_visualisation` wasn't timestamping the buffers. Fixed the test in 1aab295, and added a helpful warning in dd55512.

wmanley:
Yep, nice fixes, I'll merge

PR #367, opened 2016-06-01
stbt control: Truncate keymap filename in interactive mode

drothlis:
test_stbt_control_as_stbt_record_control_recorder__explicit_keymap was
failing with "Unable to print keymap because the terminal is too small"
when run on my PC (but not on Travis) because the test gives
`stbt control` an absolute path ($testdir/stbt-control.keymap) which on
my PC was longer than the 80-column terminal we use during the test.

The fix is to truncate the keymap filename if it's longer than the terminal width.

I've also fixed a bug in the `keymap_fits_terminal` calculation.

### DIFF
--- a/stbt-control
+++ b/stbt-control
@@ -184,9 +184,8 @@ def clear_last_command(term):
 
 def keymap_fits_terminal(term, printed_keymap):
     term_y, term_x = term.getmaxyx()
-    keymap_y = printed_keymap.count("\n")
-    keymap_x = len(printed_keymap.split("\n")[0])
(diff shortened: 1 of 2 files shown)


drothlis:
Hi @szemate, perhaps you'd like to review this since it touches your code (though it was a long time ago!). This fixes a self-test failure for me.

szemate:
It's definitely nicer to use `textwrap` than counting newlines and white spaces. I like it!

wmanley:
+1

Issue #237, opened 2014-11-14
stbt-camera selftests leave behind a lot of processes


drothlis:
gst-launch-1.0 filesrc location=tests/letters-bw.png ! pngdec ! videoconvert ! identity ! videoconvert ! vide
    gst-launch-1.0 uridecodebin uri=file://tests/capture-chessboard.png ! videoconvert ! imagefreeze ! video/x-ra
    gst-launch-1.0 videotestsrc pattern=solid-color ! video/x-raw,width=1280,height=720 ! rsvgoverlay location=te
    gst-launch-1.0 uridecodebin uri=file://tests/capture-logo.png ! videoconvert ! imagefreeze ! video/x-raw,form
    gst-launch-1.0 filesrc location=tests/letters-bw.png ! pngdec ! videoconvert ! videoscale ! video/x-raw,width
    gst-launch-1.0 filesrc location=tests/letters-bw.png ! pngdec ! videoconvert ! videobox top=-2 left=-2 ! vide
    gst-launch-1.0 filesrc location=tests/letters-bw.png ! pngdec ! videoconvert ! videoscale ! video/x-raw,width
    gst-launch-1.0 filesrc location=tests/letters-bw.png ! pngdec ! videoconvert ! rsvgoverlay location=/home/dro
    gst-launch-1.0 uridecodebin uri=file://tests/capture-letters-bw.png ! videoconvert ! imagefreeze ! video/x-ra
    /bin/bash tests/run-tests.sh -i test_that_validation_video_served_over_http_is_correct
     \_ cat uri_playlist
    /bin/bash tests/run-tests.sh -i test_illumination_compensation
     \_ cat uri_playlist
    /usr/bin/python -u tests/fake-video-src.py /tmp/stb-tester.9uS/gst-shm-socket

PR #369, opened 2016-06-02
tests: Fix teardown of stbt-camera test infrastructure

drothlis:
Some tests would only call `stop_fake_video_src` if the test passed;
others wouldn't call it at all.

This fixes https://github.com/stb-tester/stb-tester/issues/237
(stbt-camera selftests leave behind a lot of processes).


wmanley:
So simple.  Thanks Dave.

PR #371, opened 2016-06-08
Python API: Add __repr__ for stbt.MatchResult etc

drothlis:
This makes MatchResult values more convenient to use in auto-selftests.

I've added a `__repr__` to MatchParameters for consistency. All other classes either have a `__repr__` already, or they derive from namedtuple which provides an appropriate `__repr__` implementation.

I made a few other changes for consistency:

* Use %r instead of %s where appropriate (because this is `__repr__` after all) although in most cases the output is the same.

* `stbt.Region` had different output for `str(region)` vs. `repr(region)`: The former printed width & height whereas the latter printed right & bottom. Now both print ~~width & height~~ right & bottom (the repr can't use width & height because those are ambiguous if x or y are -inf).

TODO:

- [x] MatchResult repr.
- [x] Make Region repr & str consistent.
- [x] Make Region repr use bottom & right again.
- [x] Make "frame" output in MatchResult/TextMatchResult reprs have &lt;angle brackets&gt;
- [x] Release notes about change to Region description in `match` & `match_text` debug log.

### DIFF
--- a/tests/test_ocr.py
+++ b/tests/test_ocr.py
@@ -154,7 +154,7 @@ def test_match_text_stringify_result():
 
     assert re.match(
         r"TextMatchResult\(timestamp=None, match=True, region=Region\(.*\), "
-        r"frame=1280x720x3, text=u'Onion Bhaji'\)", str(result))
+        r"frame=<1280x720x3>, text=u'Onion Bhaji'\)", str(result))
 
 
 def test_that_text_region_is_correct_even_with_regions_larger_than_frame():
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -159,6 +159,13 @@ class MatchParameters(object):
         self.confirm_threshold = confirm_threshold
         self.erode_passes = erode_passes
 
+    def __repr__(self):
+        return (
+            "MatchParameters(match_method=%r, match_threshold=%r, "
+            "confirm_method=%r, confirm_threshold=%r, erode_passes=%r)"
+            % (self.match_method, self.match_threshold,
+               self.confirm_method, self.confirm_threshold, self.erode_passes))
+
 
 class Position(namedtuple('Position', 'x y')):
     """A point within the video frame.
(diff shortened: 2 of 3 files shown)


wmanley:
> * `stbt.Region` had different output for `str(region)` vs. `repr(region)`: The former printed width & height whereas the latter printed right & bottom. Now both print width & height (for consistency with the order of non-keyword args in the constructor).

I think the behaviour of `repr(region)` may have been deliberate.  See the commit message for 2901828a1c9ec7ff217f82b1e2d709d18886f197:

> * The string returned by `__repr__` lists right and bottom rather than
>   width and height.  This is necessary as `repr` should be unambiguous.

drothlis:
The test failure makes that clear:

```
======================================================================
1) FAIL: Doctest: _stbt.core.Region
----------------------------------------------------------------------
    /usr/lib/python2.7/doctest.py line 2226 in runTest
      raise self.failureException(self.format_failure(new.getvalue()))
   AssertionError: Failed doctest test for _stbt.core.Region
     File "/home/travis/build/stb-tester/stb-tester/_stbt/core.py", line 179, in Region
   
   ----------------------------------------------------------------------
   File "/home/travis/build/stb-tester/stb-tester/_stbt/core.py", line 228, in _stbt.core.Region
   Failed example:
       quadrant.translate(2, 2)
   Expected:
       Region(x=-inf, y=-inf, right=2, bottom=2)
   Got:
       Region(x=-inf, y=-inf, width=inf, height=inf)
```

drothlis:
I've squashed in a fix to the above comments, and I have fixed any selftests impacted by the change.

The one test failure is `imgproc_cache.test_that_cache_speeds_up_match` -- it seems we have a new flaky test!

wmanley:
Could do with a release note for this.  I guess this shouldn't affect `stbt auto-selftest` as that should be using `repr` already.  Will it affect logging?  Probably

tests/test_ocr.py
```diff
@@ -154,7 +154,7 @@ def test_match_text_stringify_result():
 
     assert re.match(
         r"TextMatchResult\(timestamp=None, match=True, region=Region\(.*\), "
-        r"frame=1280x720x3, text=u'Onion Bhaji'\)", str(result))
+        r"frame=<1280x720x3>, text=u'Onion Bhaji'\)", str(result))
```
wmanley: I do like this angle bracket thing.  We will be able to add some more info when #364 lands too.

drothlis:
> it seems we have a new flaky test!

I fixed that in #376.

PR #373, opened 2016-06-08
stbt auto-selftest: Improve punctuation of warning message

drothlis:


drothlis:
I wonder if it would also be worth returning a failed exit status at the end, if we raised one of these warnings. I had a syntax error in my source file, and I was seeing the below warning, but the corresponding `carousel_selftest.py` file wasn't deleted or changed in any way.

    Received name 'find_selected_tile' is not defined Exception while inspecting tests/carousel.py, Skipping

drothlis:
Or deleting the corresponding selftest file?

drothlis:
My mistake, it was deleting the generated selftest file. So that's working as intended.

This pull request still stands, it's just a fix to the warning message.

PR #374, opened 2016-06-10
stbt lint: Filter out useless warnings

drothlis:


wmanley:
Nice small improvement.

PR #375, opened 2016-06-13
Result object consistency

wmanley:
Broken out of #364.

They now all derive from object and all implement `__nonzero__`.  This will make it easier to extend them in the future in a backwards-compatible way (although this is a weakly backwards-incompatible change).

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -20,8 +20,9 @@ UNRELEASED
 
 ##### Breaking changes since 25
 
-* `TextMatchResult` (as returned from `match_text`) no longer derives from
-  `tuple`. We don't expect that this will break any real-life test scripts.
+* `TextMatchResult` (returned from `match_text`) and `MotionResult` (returned
+  from `wait_for_motion` no longer derive from `tuple`. We don't expect that
+  this will break any real-life test scripts.
 
 ##### New features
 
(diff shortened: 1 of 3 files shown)


wmanley:
@drothlis: Please review

docs/release-notes.md
```diff
@@ -20,8 +20,9 @@ UNRELEASED
 
 ##### Breaking changes since 25
 
-* `TextMatchResult` (as returned from `match_text`) no longer derives from
-  `tuple`. We don't expect that this will break any real-life test scripts.
+* `TextMatchResult` and `MotionResult` (as returned from `match_text`) no longer
```
drothlis: MotionResult isn't returned by match_text. Perhaps:

```
* `TextMatchResult` (returned from `match_text`) and `MotionResult` (returned
  from `wait_for_motion`) no longer derive from `tuple`. We don't expect that
  this will break any real-life test scripts.
```

drothlis:
Looks good to me, other than my above comment. Perhaps we should change the heading of the release notes to read "API changes" instead of "Breaking changes", but that doesn't have to be done in this pull request, we can do it at release time.

drothlis:
One more thing:

drothlis:
`MotionResult` no longer has a decent `__repr__` (namedtuple provides one). It looks like I missed `MotionResult` in my repr fixes (#371). Just rename `__str__` to `__repr__` (if there's no `__str__` it falls back to using `__repr__`, but not vice versa). Also change the `%s`'s to `%r`.

PR #377, opened 2016-06-13
MotionResult: Add region member and visualisation

wmanley:
These commits add a `region` attribute to `MotionResult` to indicate the region within which motion occurred. They also change the visualisation to a rectangle around the region that shows motion.  See:

video 1

**TODO:**

- [x] Merge #375
- [x] Squash
- [x] Write test for `MotionResult.region`
- [x] Write test for visualisation
- [x] Release note
- [x] Come up with a new name for `_nonzero_region`/`_bounding_box`


_stbt/core.py
```diff
@@ -607,6 +610,44 @@ def new_device_under_test_from_config(
...
+    ...     [0, 0, 0, 1, 0, 0],
+    ...     [0, 1, 0, 0, 0, 0],
+    ...     [0, 0, 0, 0, 1, 0],
+    ...     [0, 0, 1, 0, 0, 0],
+    ...     [0, 0, 0, 0, 0, 0]
+    ... ], dtype=numpy.uint8))
+    Region(x=1, y=1, right=5, bottom=5)
+    """
+    if len(img.shape) != 2:
+        raise ValueError("image must have 2D shape")
```
LewisHaley: Perhaps `Single-channel image required` would be more precise.

wmanley: Good idea, fixed.

_stbt/core.py
```diff
@@ -782,7 +823,12 @@ def detect_motion(self, timeout_secs=10, noise_threshold=None, mask=None):
             imglog.imwrite("absdiff_threshold", thresholded)
             imglog.imwrite("absdiff_threshold_erode", eroded)
 
-            motion = (cv2.countNonZero(eroded) > 0)
+            out_region = _nonzero_region(eroded)
+            if out_region:
+                # Undo cv2.erode above:
+                out_region = out_region.extend(x=-3, y=-3, bottom=3, right=3)
```
LewisHaley: I'm not 100% certain but I think this has the possibility of drawing completely outside the frame region, say if the video cuts to black resulting in the `out_region` being the whole frame. Perhaps you could add a translucent fill to the region to compensate.

_stbt/core.py
```diff
@@ -1422,6 +1423,28 @@ def _mainloop():
...
+        if not self.region:
+            return
+        cv2.rectangle(
+            img, (self.region.x, self.region.y),
+            (self.region.right, self.region.bottom), self.colour,
+            thickness=3)
+
+        # Slightly above the match annotation
+        label_loc = (self.region.x, self.region.y - 10)
+        _draw_text(img, self.label, label_loc, (255, 255, 255), font_scale=0.5)
```
LewisHaley: Probably not a huge inefficiency but maybe worth querying if `self.label` is empty-string before doing these two lines.

_stbt/core.py
wmanley: Yes, good thinking.  Here's what full-screen motion looks like:

video 2

So the highlight is still there but the label disappears off the top of the screen.

_stbt/core.py
wmanley: I'm not concerned about efficiency, but I think it would be a good idea to avoid drawing a black box with no text in if `label == ""`. I'll add that to `_draw_text` though rather than here.

wmanley:
Squashed

_stbt/core.py
```diff
@@ -511,17 +511,20 @@ class MotionResult(object):
...
       bool. That is, ``if result:`` will behave the same as
       ``if result.motion:``.
+    * ``region``: The region of the video frame that contained the motion.
+      `None` if no motion detected.
     """
-    def __init__(self, timestamp, motion):
+    def __init__(self, timestamp, motion, region):
         self.timestamp = timestamp
         self.motion = motion
+        self.region = region
```
drothlis: In the past we've made new parameters to these constructors optional, for backwards compatibility. See the `MatchResult` constructor.  I can imagine someone having written a `wait_for_no_motion` helper function that constructs their own `MotionResult` object. What do you think, was this a conscious decision?

_stbt/core.py
```diff
@@ -511,17 +511,20 @@ class MotionResult(object):
...
 
     def __nonzero__(self):
         return self.motion
 
     def __str__(self):
-        return "MotionResult(timestamp=%s, motion=%s)" % (
-            self.timestamp, self.motion)
+        return "MotionResult(timestamp=%s, motion=%s, region=%s)" % (
+            self.timestamp, self.motion, self.region)
```
drothlis: This should be %r as per my comment on #375.

_stbt/core.py
```diff
@@ -782,7 +824,12 @@ def detect_motion(self, timeout_secs=10, noise_threshold=None, mask=None):
             imglog.imwrite("absdiff_threshold", thresholded)
             imglog.imwrite("absdiff_threshold_erode", eroded)
 
-            motion = (cv2.countNonZero(eroded) > 0)
+            out_region = _nonzero_region(eroded)
```
drothlis: "bounding box" might be a better name than "nonzero region".

tests/test-motion.sh
```diff
@@ -77,6 +77,9 @@ test_detect_motion_reports_motion() {
 	for motion_result in detect_motion():
 	    assert bool(motion_result) == motion_result.motion
 	    if motion_result:
+	        # videotestsrc has motion in bottom right corner:
+	        assert motion_result.region == stbt.Region(
+	            240, 180, right=320, bottom=240)
```
drothlis: I ran this test 1,000 times and it always passed, so it seems that the random noise generated by videotestsrc will always have some motion along the edge rows/columns of this region. Or maybe videotestsrc is deterministic. All good, in any case.

_stbt/core.py
```diff
@@ -1423,6 +1424,28 @@ def _mainloop():
               "is still alive!" if thread.isAlive() else "ok"))
 
 
+class _Annotation(namedtuple("_Annotation", "pts region label colour")):
+    MATCHED = (32, 0, 255)  # Red
+    NO_MATCH = (32, 255, 255)  # Yello
```
drothlis: ...w

_stbt/core.py
```diff
@@ -1630,8 +1653,9 @@ def draw(self, obj, duration_secs):
                 self.text_annotations.append(
                     {"text": obj, "duration": duration_secs * Gst.SECOND})
             elif isinstance(obj, MatchResult):
-                if obj.timestamp is not None:
```
drothlis: I imagine this check for `is not None` was so that the very first frame (with timestamp 0?) wouldn't be excluded. It's unlikely to matter though.

_stbt/core.py
```diff
@@ -1593,7 +1593,6 @@ def gst_samples(self, timeout_secs=None):
                             timeout_secs * Gst.SECOND))
                         return
 
-                sample = gst_sample_make_writable(sample)
```
drothlis: Nice!

drothlis:
+1

_stbt/core.py
wmanley: I wasn't too worried about this break, but I suppose defaulting to `None` doesn't hurt.

_stbt/core.py
wmanley: Fixed

_stbt/core.py
wmanley: Good idea, fixed

_stbt/core.py
wmanley: Fixed

wmanley:
I'll merge once Travis is happy

wmanley:
No, hold on, it's missing a release note.

wmanley:
Release notes written

_stbt/core.py
wmanley: Ah, my fix was not so good:

> _stbt/core.py:613: [E0102(function-redefined), _bounding_box] function already defined line 387

Any other suggestions for a name?

drothlis: Not really. "bounding box of two regions" and "bounding box of nonzero pixels"? Or just leave it as is.

wmanley: I've gone for `_pixel_bounding_box`

wmanley:
Thanks both @LewisHaley and @drothlis for the reviews.

Issue #365, opened 2016-05-26
Control device using CEC


danielandersson:
I think it would be a great feature to have the possibility to control the connected device using the CEC feature in HDMI. It might at first look like a weird feature as most of the capture devices lacks CEC, however with an USB - CEC adapter ([for example](https://www.pulse-eight.com/p/104/usb-hdmi-cec-adapter#)) it would be possible to reach devices that cannot be controlled through IR nor IP.

In my case I'm facing a situation where I would like to use STB-Tester on Playstation 4, a device without IR nor IP controllability.

TODO:

- [x] Add stb-tester package for GPL dependent features
- [x] Add HdmiCecControl to `_stbt/control_gpl.py`
- [x] Implement `__init__()` and `press()`
- [ ] CEC addressing
- [ ] Write tests to make sure it works
    - Please discuss test strategy below, e.g. mocks vs fakes, etc.
- [ ] Add release note

wmanley:
That sounds like it could be really useful, particularly for the PS4 and other devices that don't have traditional remote controls like the Chromecast.

There are some licencing issues as libcec is GPLv2, whereas stb-tester is LGPLv2+.  As we would have to link against libcec we should be careful to ensure that the dependency is optional.  Perhaps it could be a separate Debian package so users who are happy with GPL test-packs can install the additional package.  I can help out with this bit.  I'll write some scaffolding that you can fill in.

I'm not sure what the best testing strategy is in this instance.  Perhaps mocking out the libcec API?  Usually I'm not a fan of mocking because it doesn't actually check that you've got the integration right, but it's better than no tests at all.  Unless you can think of a more representative test?

We look forward to your patches.

wmanley:
I've created the packaging scaffolding to help you on your way. It solves the boring build bits so you can get on with the functionality proper.  See the hdmi-cec-control branch of stb-tester:

https://github.com/stb-tester/stb-tester/commits/hdmi-cec-control

So the next step is to fill in the `HdmiCecControl` class from c87c71887326af1bbad10c1372b13bdb6ac8bbce.

Let us know by leaving a comment here if you need any more help.

alexeykorobov:
Hi there, I'm experimenting with the PS4 control as well. Have you managed to turn on the PS4 using the that USB-to-CEC device? I tried the CEC control with several TVs but haven't succeeded

danielandersson:
I just received the device and will start experimenting with it in the upcoming days.
[...]

PR #378, opened 2016-06-14
Hdmi cec control

danielandersson:
Implementation for #365 

TODO:

* [ ] Add some automated tests  – only been tested in practice.
* [ ] Need to workaround the CEC-addressing.
* [ ] Make source and destination addresses configurable

### DIFF
--- a/_stbt/control.py
+++ b/_stbt/control.py
@@ -42,7 +42,10 @@ def uri_to_remote(uri, display=None):
         (r'x11:(?P<display>[^,]+)?(,(?P<mapping>.+)?)?', _X11Remote),
     ]
     if gpl_controls is not None:
+        print "GPL Controls ACTIVATED"
         remotes += gpl_controls
+
+    print(remotes)
     for regex, factory in remotes:
         m = re.match(regex, uri, re.VERBOSE | re.IGNORECASE)
         if m:
--- a/_stbt/control.py
+++ b/_stbt/control.py
@@ -42,10 +42,8 @@ def uri_to_remote(uri, display=None):
         (r'x11:(?P<display>[^,]+)?(,(?P<mapping>.+)?)?', _X11Remote),
     ]
     if gpl_controls is not None:
-        print "GPL Controls ACTIVATED"
         remotes += gpl_controls
 
-    print(remotes)
     for regex, factory in remotes:
         m = re.match(regex, uri, re.VERBOSE | re.IGNORECASE)
         if m:
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -1,4 +1,5 @@
-from .logging import debug, scoped_debug_level
+from .logging import debug
+
 
 class HdmiCecControl(object):
     cecconfig = {}
@@ -7,28 +8,29 @@ class HdmiCecControl(object):
     # Map our recommended keynames (from linux input-event-codes.h) to the
     # equivalent CEC commands.
     _KEYNAMES = {
-        "KEY_OK"    : "14:44:00",
-        "KEY_UP"    : "14:44:01",
-        "KEY_DOWN"  : "14:44:02",
-        "KEY_LEFT"  : "14:44:03",
-        "KEY_RIGHT" : "14:44:04",
-        "KEY_BACK"  : "14:44:0D"
+        "KEY_OK": "14:44:00",
+        "KEY_UP": "14:44:01",
+        "KEY_DOWN": "14:44:02",
+        "KEY_LEFT": "14:44:03",
+        "KEY_RIGHT": "14:44:04",
+        "KEY_BACK": "14:44:0D"
     }
 
     def __init__(self, device):
         import cec
         self.cecconfig = cec.libcec_configuration()
-        self.cecconfig.strDeviceName   = "STB-Tester"
+        self.cecconfig.strDeviceName = "STB-Tester"
         self.cecconfig.bActivateSource = 0
         self.cecconfig.deviceTypes.Add(cec.CEC_DEVICE_TYPE_RECORDING_DEVICE)
         self.cecconfig.clientVersion = cec.LIBCEC_VERSION_CURRENT
         self.lib = cec.ICECAdapter.Create(self.cecconfig)
         # print libCEC version and compilation information
-        debug("libCEC version " + self.lib.VersionToString(self.cecconfig.serverVersion) + " loaded: " + self.lib.GetLibInfo())
+        debug("libCEC version " + self.lib.VersionToString(self.cecconfig.serverVersion) +
+              " loaded: " + self.lib.GetLibInfo())
 
         if device == 'auto':
             device = self.DetectAdapter()
-            if device == None:
+            if device is None:
                 debug("No adapters found")
         if self.lib.Open(device):
             debug("connection opened")
@@ -44,7 +46,6 @@ class HdmiCecControl(object):
             debug("command sent")
         else:
             debug("failed to send command")
-    
 
     # detect an adapter and return the com port path
     def DetectAdapter(self):
@@ -57,11 +58,11 @@ class HdmiCecControl(object):
             debug("product:  " + hex(adapter.iProductId))
             retval = adapter.strComName
         return retval
-        
+
 
 def test_hdmi_cec_control():
-    from .control import uri_to_remote
-    control = uri_to_remote('hdmi-cec:192.168.1.3')
+    # from .control import uri_to_remote
+    # control = uri_to_remote('hdmi-cec:192.168.1.3')
     assert False, "Tests not written"
 
 controls = [
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -1,7 +1,9 @@
 from .logging import debug
 
+
 class HdmiCecError(Exception):
     pass
+
 
 class HdmiCecControl(object):
 
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -1,12 +1,14 @@
 from .logging import debug
 
+class HdmiCecError(Exception):
+    pass
 
 class HdmiCecControl(object):
-    cecconfig = {}
-    lib = {}
 
     # Map our recommended keynames (from linux input-event-codes.h) to the
     # equivalent CEC commands.
+    # The mapping between CEC commands and code can be found at http://www.cec-o-matic.com
+    # or in HDMI-CEC specification 1.3a
     _KEYNAMES = {
         "KEY_OK": "14:44:00",
         "KEY_UP": "14:44:01",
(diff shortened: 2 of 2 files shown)


_stbt/control.py
```diff
@@ -42,7 +42,10 @@ def uri_to_remote(uri, display=None):
         (r'x11:(?P<display>[^,]+)?(,(?P<mapping>.+)?)?', _X11Remote),
     ]
     if gpl_controls is not None:
+        print "GPL Controls ACTIVATED"
         remotes += gpl_controls
+
+    print(remotes)
```
wmanley: You'll want to remove these debug statements.

_stbt/control_gpl.py
```diff
@@ -1,16 +1,68 @@
+from .logging import debug
+
+
 class HdmiCecControl(object):
+    cecconfig = {}
+    lib = {}
```
wmanley: I don't think either of these should be here.  They look like they should be attached to the object rather than the class.  Removing these two lines should be sufficient.

_stbt/control_gpl.py
```diff
@@ -1,16 +1,68 @@
...
+        self.cecconfig.strDeviceName = "STB-Tester"
+        self.cecconfig.bActivateSource = 0
+        self.cecconfig.deviceTypes.Add(cec.CEC_DEVICE_TYPE_RECORDING_DEVICE)
+        self.cecconfig.clientVersion = cec.LIBCEC_VERSION_CURRENT
+        self.lib = cec.ICECAdapter.Create(self.cecconfig)
+        # print libCEC version and compilation information
+        debug("libCEC version " + self.lib.VersionToString(self.cecconfig.serverVersion) +
+              " loaded: " + self.lib.GetLibInfo())
+
+        if device == 'auto':
```
wmanley: I would suggest making this conditional `device is None`.  We can change the regex below to pass `device=None` if you specify `hdmi-cec:` as your control.  This will be more consistent with the other remotes.

_stbt/control_gpl.py
```diff
@@ -1,16 +1,68 @@
...
+            device = self.DetectAdapter()
+            if device is None:
+                debug("No adapters found")
```
wmanley: This should probably be raising an exception rather than printing to debug.  Perhaps create `HdmiCecException` with:

    class HdmiCecError(Exception):
        pass

and then do `raise HdmiCecError("No adaptors found")` here.

_stbt/control_gpl.py
```diff
@@ -1,16 +1,68 @@
...
+        if self.lib.Open(device):
+            debug("connection opened")
+        else:
+            debug("failed to open a connection to the CEC adapter")
```
wmanley: This should also be raising an exception.

_stbt/control_gpl.py
```diff
@@ -1,16 +1,68 @@
...
+
+    # Map our recommended keynames (from linux input-event-codes.h) to the
+    # equivalent CEC commands.
+    _KEYNAMES = {
+        "KEY_OK": "14:44:00",
+        "KEY_UP": "14:44:01",
+        "KEY_DOWN": "14:44:02",
+        "KEY_LEFT": "14:44:03",
+        "KEY_RIGHT": "14:44:04",
+        "KEY_BACK": "14:44:0D"
```
wmanley: Do you have a reference for these codes here?  It would be good to link to it in the comment you've made above.

_stbt/control_gpl.py
```diff
@@ -1,16 +1,68 @@
...
+                debug("No adapters found")
+        if self.lib.Open(device):
+            debug("connection opened")
+        else:
+            debug("failed to open a connection to the CEC adapter")
 
     def press(self, key):
-        import cec
-        raise NotImplementedError()
+        cec_command = self._KEYNAMES.get(key, key)
```
wmanley: Right, so key can be either one of our symbolic keys like `KEY_UP` or a CEC command of the form `14:xx`? What use-case had you in mind for the latter?

_stbt/control_gpl.py
```diff
@@ -1,16 +1,68 @@
...
+        cmd = self.lib.CommandFromString(cec_command)
+        debug("transmit " + key + " as " + cec_command)
+        if self.lib.Transmit(cmd):
+            self.lib.Transmit(self.lib.CommandFromString('14:45'))
+            debug("command sent")
+        else:
+            debug("failed to send command")
```
wmanley: This should probably be `raise HdmiCecError("failed to send command")`.

_stbt/control_gpl.py
```diff
@@ -1,16 +1,68 @@
...
+
+    # detect an adapter and return the com port path
+    def DetectAdapter(self):
```
wmanley: Could you rename this `detect_adapter` for [pep8 compliance](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables)?

_stbt/control_gpl.py
```diff
@@ -1,16 +1,68 @@
...
+        retval = None
+        adapters = self.lib.DetectAdapters()
+        for adapter in adapters:
+            debug("found a CEC adapter:")
+            debug("port:     " + adapter.strComName)
+            debug("vendor:   " + hex(adapter.iVendorId))
+            debug("product:  " + hex(adapter.iProductId))
+            retval = adapter.strComName
```
wmanley: This method will return the last adapter that `lib.DetectAdapters` lists.  Is this intentional?  Perhaps this should just be `return adapter.strComName` unless I've misunderstood?

danielandersson:
Hey @wmanley, 
I've adjusted the code according to your comments. Most of it was straightforward and leftovers from the pyCecClient that was the inspiration. 

Great comment with the new exception.
For the CEC codes, I added a short description and a link to a extremely nice site to lookup the codes for a specific command. Moreover I think a Wiki-page with a wider description of CEC might be necessary.
As for the "lib.DetectAdapters" you're absolutely correct is lists all connected adapters. In the above mentioned [pyCeceClient](https://github.com/Pulse-Eight/libcec/blob/master/src/pyCecClient/pyCecClient.py) they uses it like this. Thats is the reason I left it. If you're having multiple adapters you have to specify the correct one in _stbt.conf_.

Do you have any suggestions on how to solve the CEC addressing issue? The first two numbers in the CEC-codes, in all of the examples in the code now - 14, is first the source and secondly the destination. These two number have to be configured, at least the destination, by the end-user. The problem is that it isn't that user-friendly to identify what ID the device-under-testing have. Any ideas?

wmanley:
> Do you have any suggestions on how to solve the CEC addressing issue? The first two numbers in the CEC-codes, in all of the examples in the code now - 14, is first the source and secondly the destination.

I was having a look at the excellent CEC website you shared.  What about source: TV, destination: Broadcast (0F).  Would that work?  I think we'd also want to make either option configurable.  I'll make a comment in-line describing how to do this.

_stbt/control_gpl.py
```diff
@@ -1,16 +1,72 @@
...
+    _KEYNAMES = {
+        "KEY_OK": "14:44:00",
+        "KEY_UP": "14:44:01",
+        "KEY_DOWN": "14:44:02",
+        "KEY_LEFT": "14:44:03",
+        "KEY_RIGHT": "14:44:04",
+        "KEY_BACK": "14:44:0D"
+    }
+
     def __init__(self, device):
```
wmanley: To make the source and destinations configurable make this

    def __init__(self, device=None, source=None, destination=None):

and add

    if source is None:
        source = 0
    if destination is None:
        destination = 0xf

here

_stbt/control_gpl.py
```diff
@@ -1,16 +1,72 @@
...
 
 
 def test_hdmi_cec_control():
-    from .control import uri_to_remote
-    control = uri_to_remote('hdmi-cec:192.168.1.3')
+    # from .control import uri_to_remote
+    # control = uri_to_remote('hdmi-cec:192.168.1.3')
     assert False, "Tests not written"
 
 controls = [
```
wmanley: and modify the regex here to say:

     (r'hdmi-cec:(?P<device>[^:]+)(:(?P<source>[^:]+)(:(?P<destination>[^:]+))?)?', HdmiCecControl),

(not tested)

wmanley:
I think you'll need to open a new pull request too.  Above it says:

> @danielandersson wants to merge 4 commits into `stb-tester:hdmi-cec-control` from `danielandersson:hdmi-cec-control`

Whereas it should say:

> @danielandersson wants to merge 4 commits into `stb-tester:master` from `danielandersson:hdmi-cec-control`

(merging into stb-tester:master).

I don't know how to change this without closing this pull-request and opening a new one.

danielandersson:
Wouldn't it be more suitable to have "Recording 1" as default source since it describes the device best? If we decide to use TV it might collide in case a TV is connected and I know from my experiments that for example a Playstation 4 depends on having a device with CEC+EDID to control it, something TV usually have but not all USB-CEC adapters.

As for the destination, having broadcast as default feels like a dirty solution but also the most user-friendly one. Not optimal if multiple devices are connected to the CEC chain.

wmanley:
> Wouldn't it be more suitable to have "Recording 1" as default source since it describes the device best

I don't think so.  stb-tester is intended to simulate actual usage of set-top boxes as if a human were using it.  Humans use them via TVs so stb-tester should pretend to be a TV.

> I know from my experiments that for example a Playstation 4 depends on having a device with CEC+EDID to control it, something TV usually have but not all USB-CEC adapters.

So you're saying that it doesn't work if the USB-CEC adapter pretends to be a TV, whereas it does work when identifying itself as "Recording 1"?

> As for the destination, having broadcast as default feels like a dirty solution but also the most user-friendly one. Not optimal if multiple devices are connected to the CEC chain.

I admit I know little about HDMI-CEC, but I would have thought that if you can have multiple devices in a CEC chain there must be some way that they avoid sender ID conflicts?  Do you know how these numbers get assigned in practice?  Could we query that somehow?

OTOH, if broadcast works for all the cases we care about then maybe it's not that big a deal?

danielandersson:
I understand and agree that the "consuming side" of the HDMI, which is the capture card and USB-CEC adapter, should be considered the TV preferable. As you say, the purpose is to simulate and end-to-end usage using a remote to control and output image to validate. However, for the perspective of the HDMI signal and CEC – the capture card "side" records the signal.

I have experimented with Apple TV and Playstation 4 and for the first mentioned it worked as intended. The Apple TV requested a CEC-address (Playback 1) and the USB-CEC adapter requested Recorder 1. No problems.
When I tried with Playstation 4 however, it didn't request a CEC-address when only having the USB-CEC adapter and PS4 in the HDMI chain. The PS4 was totally quite and didn't listen nor say anything. After some investigation I found that when connecting the PS4 to a TV the CEC, using the TV remote, was fully working. To debug this I connected the USB-CEC adapter to the very same TV, having the adapter and PS4 connected at the same time.

                                                    TV
                                   TV HDMI 1  -------------  TV HDMI 2
                                  |                                   |
                                  |                                   |
                                  |                                   |
                                Playstation 4                   USB-CEC 

This allowed me to listen to the CEC channel and it was using this I discovered that the PS4 must be communicating using another channel before initing CEC.
I found after searching online that PS4 listens on the EDID channel before enabling CEC. And since the USB-CEC adapter does not have the EDID channel it can't enable CEC on PS4.

So with this issue with PS4, if we set the USB-CEC adapter as TV by default in STB-Tester and since there is only one slot for TV in the CEC-addressing (apart from the reserved address "E") we will occupy and it might lead to potential issues.
However, as the PS4 was connected to the TV the capture card couldn't capture the image so thing might be a corner case. (Since it is dependent on stbt camera)


TLDR;
> Humans use them via TVs so stb-tester should pretend to be a TV.

I agree but I think we should make an exception and have it as recording device to not interfere with TV.

> So you're saying that it doesn't work if the USB-CEC adapter pretends to be a TV ...

From my tries, it has worked both as TV and recorder.

> they avoid sender ID conflicts?

I'm do not understand this fully but from what I understand they mitigate the conflict by taking another slot. That is why you have multiple addresses for most of the types, "Recording 1 & Recording 2" etc.

`self.cecconfig.deviceTypes.Add(cec.CEC_DEVICE_TYPE_RECORDING_DEVICE)`

SidePlace:
Hi,
Anyone still looking into this branch? Any chance we still having this? I really want the CEC support.

drothlis:
Moved to #393.

PR #379, opened 2016-06-14
Refactor: Hide `Gst.Sample` in `_Display` for simplicity

wmanley:
We were previously passing around `Gst.Sample`s because:

* There are timestamp metadata attached to them
* It was easier to handle lifetime management of the buffers, specifically WRT calling `gst_buffer_unmap`

These commits create a new numpy array type and a new backing store to solve each of those problems respectively.

Now this is done:

* We now no longer need to put `numpy_from_sample` everywhere.
* It opens the door to reducing copying, in the future we will be able to give clients these frames directly
* We have a place to attach arbitrary extra metadata to frames

I've implemented this for the last point.  I'll be needing that for #364.

TODO:

- [x] Release note: internal change

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -38,7 +38,7 @@ from _stbt import imgproc_cache, logging, utils
 from _stbt.config import ConfigurationError, get_config
 from _stbt.gst_utils import (array_from_sample, get_frame_timestamp,
                              gst_iterate, gst_sample_make_writable,
-                             numpy_from_sample, sample_shape)
+                             sample_shape)
 from _stbt.logging import ddebug, debug, warn
 
 gi.require_version("Gst", "1.0")
@@ -1658,20 +1658,20 @@ class Display(object):
                     self.annotations.remove(annotation)
 
         sample = gst_sample_make_writable(sample)
-        with numpy_from_sample(sample) as img:
-            # Text:
-            _draw_text(
-                img, datetime.datetime.now().strftime("%H:%M:%S.%f")[:-4],
-                (10, 30), (255, 255, 255))
-            for i, x in enumerate(reversed(texts)):
-                origin = (10, (i + 2) * 30)
-                age = float(now - x['start_time']) / (3 * Gst.SECOND)
-                color = (int(255 * max([1 - age, 0.5])),) * 3
-                _draw_text(img, x['text'], origin, color)
+        img = array_from_sample(sample, readwrite=True)
+        # Text:
+        _draw_text(
+            img, datetime.datetime.now().strftime("%H:%M:%S.%f")[:-4],
+            (10, 30), (255, 255, 255))
+        for i, x in enumerate(reversed(texts)):
+            origin = (10, (i + 2) * 30)
+            age = float(now - x['start_time']) / (3 * Gst.SECOND)
+            color = (int(255 * max([1 - age, 0.5])),) * 3
+            _draw_text(img, x['text'], origin, color)
 
-            # Regions:
-            for annotation in annotations:
-                annotation.draw(img)
+        # Regions:
+        for annotation in annotations:
+            annotation.draw(img)
 
         self.appsrc.props.caps = sample.get_caps()
         self.appsrc.emit("push-buffer", sample.get_buffer())
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -36,9 +36,9 @@ from kitchen.text.converters import to_bytes
 
 from _stbt import imgproc_cache, logging, utils
 from _stbt.config import ConfigurationError, get_config
-from _stbt.gst_utils import (get_frame_timestamp, gst_iterate,
-                             gst_sample_make_writable, numpy_from_sample,
-                             sample_shape)
+from _stbt.gst_utils import (array_from_sample, get_frame_timestamp,
+                             gst_iterate, gst_sample_make_writable,
+                             numpy_from_sample, sample_shape)
 from _stbt.logging import ddebug, debug, warn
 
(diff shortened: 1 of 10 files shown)


wmanley:
I've left a bunch of copies in that could potentially be removed, such as in `DeviceUnderTest.get_frame()`, `DeviceUnderTest.frames()` and `DeviceUnderTest._match_all()`.

wmanley:
@drothlis: Please review

_stbt/gst_utils.py
```diff
@@ -168,8 +168,27 @@ def __del__(self):
...
+    """
+    def __new__(cls, array, dtype=None, order=None, _gst_pts=None):
+        obj = numpy.asarray(array, dtype=dtype, order=order).view(cls)
+        obj._gst_pts = _gst_pts  # pylint: disable=protected-access
+        return obj
+
+    def __array_finalize__(self, obj):
+        if obj is None:
+            return
+        self._gst_pts = getattr(obj, '_gst_pts', None)  # pylint: disable=protected-access,attribute-defined-outside-init
```
drothlis: Was there documentation you followed on how to subclass an ndarray? Would be useful to mention that link in the commit message.

_stbt/gst_utils.py
```diff
@@ -168,8 +168,27 @@ def __del__(self):
...
+
+
 def array_from_sample(sample, readwrite=False):
-    return numpy.array(_MappedSample(sample, readwrite), copy=False)
+    f = numpy.array(
+        _MappedSample(sample, readwrite), copy=False).view(CapturedFrame)
```
drothlis: Any particular reason why you do `numpy.array(args).view(CapturedFrame)` instead of `CapturedFrame(args)`?

_stbt/core.py
```diff
@@ -738,7 +738,7 @@ def _match_all(self, image, frame, match_parameters, region):
             # Remove this copy once we make numpy_from_sample *not* unmap the
             # sample when the contextmanager exits, but when no further
             # references are held to the numpy image.
-            npframe = numpy.copy(npframe)
+            npframe = npframe.copy()
```
drothlis: This comment has now been implemented, but we still haven't removed the copy -- we should modify the comment to state when it will be safe to remove the copy.

drothlis: Hmmm...  or has it. Why are we still using `numpy_from_sample`? (Here and in `Display.push_sample`.)

_stbt/gst_utils.py
```diff
@@ -168,8 +168,27 @@ def __del__(self):
             self._ctx.__exit__(None, None, None)
 
 
+class CapturedFrame(numpy.ndarray):
```
drothlis: Presumably we're going to make this a public API at some future point (#364 perhaps).

_stbt/core.py
```diff
@@ -2619,17 +2616,16 @@ def test_wait_for_motion_half_motion_int():
...
                 numpy.ones((2, 2, 3), dtype=numpy.uint8) * 255,
             ]
             for i in range(10):
-                buf = Gst.Buffer.new_wrapped(data[(i // 2) % 2].flatten())
-                buf.pts = i * 1000000000
-                yield gst_sample_make_writable(
-                    Gst.Sample.new(buf, Gst.Caps.from_string(
-                        'video/x-raw,format=BGR,width=2,height=2'), None, None))
+                yield CapturedFrame(
+                    data[(i // 2) % 2].flatten().reshape((2, 2, 3)),
```
drothlis: Why is this flatten + reshape necessary, if `data[0]` and `data[1]` are already (2,2,3) numpy arrays?

drothlis:
Why do we still use `numpy_from_sample` in `Display.push_sample`? (And in `stbt_camera_calibrate.py` and `stbt_camera_validate.py`).

_stbt/gst_utils.py
wmanley: Yep: http://docs.scipy.org/doc/numpy-1.10.1/user/basics.subclassing.html#slightly-more-realistic-example-attribute-added-to-existing-array I'll add it when I rebase.

_stbt/gst_utils.py
wmanley: I'll try it.

wmanley: It works and looks much nicer.  Thanks

_stbt/gst_utils.py
wmanley: Yes

_stbt/core.py
wmanley: Yep, nicer

wmanley:
> Why do we still use numpy_from_sample in Display.push_sample? (And in stbt_camera_calibrate.py and stbt_camera_validate.py).

In these instances we're actually dealing with `Gst.Sample`s directly.  I guess they could be refactored too, I just didn't want to fiddle with code that needs to be `Gst.Sample` aware anyway.  I'll have a quick play.

wmanley:
> Why do we still use numpy_from_sample in Display.push_sample? (And in stbt_camera_calibrate.py and stbt_camera_validate.py).

Ok, I've replaced these too and removed `numpy_from_sample`.

wmanley:
Squashed.  @drothlis: You might like to look at the last commit before merging.

drothlis:
Looks good to me. Feel free to merge once travis has run.

wmanley:
Merged.  Some notes on possible future work:

- Replace `_AnnotatedTemplate` with an `ndarray` subclass
- Remove the copying when giving frames to users (grep for `.copy(`).  This will require validation that `push_sample` isn't doodling on frames that have already been passed to users.

PR #380, opened 2016-06-15
`match` debug image logging: Don't show second pass

drothlis:
...if confirm_method="none".

Previously the generated html would show a table for the second pass,
full of broken links to images that didn't exist. Now it only shows the
table if confirm_method isn't "none".

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -2234,49 +2234,53 @@ def _log_match_image_debug(imglog):
         </table>
 
         {% if show_second_pass %}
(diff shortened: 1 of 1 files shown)


drothlis:
Note that most of the change is an indentation change, so view the diff [here](https://github.com/stb-tester/stb-tester/pull/380/files?w=1) with `?w=1`.

wmanley:
LGTM.  I'll take your word for it that it works :)

PR #387, opened 2016-10-19
stbt.match_text: Fix exception for single-channel images

drothlis:
The `__repr__` of `TextMatchResult` was raising an `IndexError` if
`match_text` was given a single-channel (black-and-white or greyscale)
image. This means that you couldn't do custom thresholding before
running the image through `match_text`. This was the stack trace:

     [...]
     _stbt/core.py line 1036 in match_text
       debug("match_text: Match found: %s" % str(result))
     _stbt/core.py line 612 in __repr__
       self.frame.shape[2]),
    IndexError: tuple index out of range

Note that I have also changed `MatchResult.__repr__` for consistency,
but `match` probably needs other changes too before it can support
single-channel images.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -447,9 +447,7 @@ class MatchResult(object):
                 self.match,
                 self.region,
                 self.first_pass_result,
-                "None" if self.frame is None else "%dx%dx%d" % (
-                    self.frame.shape[1], self.frame.shape[0],
-                    self.frame.shape[2]),
+                _str_frame_dimensions(self.frame),
                 "<Custom Image>" if isinstance(self.image, numpy.ndarray)
                 else repr(self.image)))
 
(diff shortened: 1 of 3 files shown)


_stbt/core.py
```diff
@@ -468,6 +466,15 @@ def __nonzero__(self):
         return self.match
 
 
+def _str_frame_dimensions(frame):
+    if frame is None:
+        return "None"
+    if len(frame.shape) == 3:
+        return "%dx%dx%d" % (frame.shape[1], frame.shape[0], frame.shape[2])
+    else:
+        return "%dx%dx" % (frame.shape[1], frame.shape[0])
```
LewisHaley: Do you perhaps what `"%dx%d"` not `"%dx%d*x*"`?

drothlis: Good catch, thanks! Fixed.

PR #388, opened 2016-10-19
stbt.ocr & stbt.match_text: Add `text_color` parameter

drothlis:
In low-contrast situations (for example white text on a blue background,
such as the "Summary" tab in `tests/action-panel.png`) tesseract's
binarisation algorithm often picks a threshold that removes the desired
text. So far we have worked around this in our test scripts by
specifying a region, so that the blue background is the only background
visible anywhere in the image, which usually helps tesseract to pick a
suitable threshold. But we can't always specify a region, for example
when the location of a menu varies depending on the contents of the
menu.

When you specify `text_color`, we perform binarisation ourselves based
on this information.

We use a simple euclidean distance in the BGR colorspace, which perhaps
isn't ideal but it seems to work. It might be worth testing the
weightings suggested in [1](http://www.compuphase.com/cmetric.htm), or for something more computationally
expensive use CIE DeltaE-1994 [2](http://stackoverflow.com/a/25726414/606705). I've picked an arbitrary threshold of
25, which is 10% of the possible range (0 - 255).

Note that we do the binarisation _after_ scaling the image up, to
retain (some of) the information in the anti-aliasing of the text.

---

TODO:
- [x] Tests.


drothlis:
@LewisHaley I want to add some more test cases but in the meantime it'd be great to get your thoughts on this.

LewisHaley:
It looks potentially very valuable but it all depends on the (increased) accuracy this option provided. It might also be better (rather than having 1 colour and a threshold), to specify an upper and lower bound for the colour tuples.

drothlis:
@LewisHaley In theory this should help for text over dynamic backgrounds (e.g. where the background is a translucent overlay over live TV, or even over live TV directly with no background). It would be great if you could try it out with such cases in your own UIs.

Re. the hardcoded threshold it has worked well so far in everything I've tried it on, and I'd like to avoid having to specify more parameters than necessary. @wmanley wants to add (eventually) an optional `background_color` parameter so that we could do a contrast-stretch from (min=background_color, max=text_color) to (min=0, max=255) and then let tesseract pick its own threshold. But that can be added later (not in this pull request) and you'd still be able to specify `text_color` without `background_color` to get this pull request's behaviour. Hope that makes sense.

_stbt/core.py
```diff
@@ -2527,21 +2526,33 @@ def _tesseract(frame, region, mode, lang, _config,
...
     # uses bilinear interpolation which produces the best results.  See
     # http://stb-tester.com/blog/2014/04/14/improving-ocr-accuracy.html
     outsize = (frame.shape[1] * 3, frame.shape[0] * 3)
-    subframe = cv2.resize(frame, outsize, interpolation=cv2.INTER_LINEAR)
+    frame = cv2.resize(frame, outsize, interpolation=cv2.INTER_LINEAR)
+
+    if text_color is not None:
+        # Calculate distance of each pixel from `text_color`, then discard
+        # everything further than `threshold` distance away.
+        diff = frame - text_color
```
wmanley: What about overflow?  Presumably these are both `uint8`?

_stbt/core.py
```diff
@@ -2527,21 +2526,33 @@ def _tesseract(frame, region, mode, lang, _config,
...
+        frame = numpy.sqrt((diff[:, :, 0] ** 2 +
+                            diff[:, :, 1] ** 2 +
+                            diff[:, :, 2] ** 2) / 3) \
+                     .astype(numpy.uint8)
+        threshold = 25
+        _, frame = cv2.threshold(frame, threshold, 255, cv2.THRESH_BINARY)
```
drothlis: This means that pixels close to the text color will go to black, and pixels far away from the text color will go to white. This is the right way round for tesseract (it likes black text on white backgrounds).

drothlis:
Rebased onto master to resolve conflicts with #407.

_stbt/core.py
drothlis: > >         diff = frame - text_color
> What about overflow? Presumably these are both uint8?

`frame` is `uint8`. `text_color` is just a (non-numpy) tuple.

If they were both `uint8` numpy arrays the result would be `uint8` and there would be overflow. Since `text_color` isn't a numpy array the result is `int64`.

I'll push a fix to make this more explicit by using `diff = numpy.subtract(frame, text_color, dtype=numpy.int32)`.

drothlis: P.S. `int32` is sufficient to hold the largest possible diff value (255 or -255), squared, times 3.

drothlis:
Specifying `text_color` can even improve the run-time performance of ocr:

```
f = cv2.imread("tests/action-panel.png")
%timeit stbt.ocr(f)
1 loop, best of 3: 2.96 s per loop
%timeit stbt.ocr(f, text_color=(235,235,235))
1 loop, best of 3: 1.19 s per loop
```

drothlis:
We now have 2 real-world examples where `text_color` can help. See the comments in [test_ocr_text_color](https://github.com/stb-tester/stb-tester/pull/388/files#diff-f387e57a67f3eb52177a29382ad8fac3R226) for details.

PR #390, opened 2016-10-20
stbt.match_text: Normalise unicode punctuation & ligatures

drothlis:
If tesseract reads the text "stb-tester" as u"stb\u2014tester" (\u2014
is the unicode code point for "em dash") then `stbt.ocr()` already
normalises it and returns "stb-tester" with an ascii dash.

But `stbt.match_text("stb-tester")` was returning False, because it
wasn't doing the same normalisation. Note that we have to normalise the
text that tesseract returns *and* the text that you give to match_text,
for backwards compatibility with users who previously worked around this
by saying `match_text(u"stb\u2014tester")`.


drothlis:
@wmanley please review.

_stbt/core.py
```diff
@@ -2628,14 +2628,15 @@ def _hocr_iterate(hocr):
 
 
 def _hocr_find_phrase(hocr, phrase):
-    l = list(_hocr_iterate(hocr))
-    words_only = [(w, elem) for w, elem in l if w.strip() != u'']
+    words_only = [(w.lower().translate(_ocr_transtab), elem)
+                  for w, elem in _hocr_iterate(hocr) if w.strip() != u'']
+    phrase = [unicode(w).lower().translate(_ocr_transtab) for w in phrase]
```
wmanley: I'm a little worried about this unicode(). Passing in a utf-8 encoded string to match_text would have worked before, but now will raise.

wmanley:
Yes, this great fix to a subtle bug.  It must have been tricky to work out that this was causing whatever problem you were looking at.

_stbt/core.py
drothlis: If it's a byte string, the call to `translate(_ocr_transtab)` would fail with:

    TypeError: expected a string or other character buffer object

I need to look into what `translate` is actually doing -- maybe we'll need an explicit type check and 2 different code paths.

wmanley: Yes, something like:

```
def to_unicode(text):
    if isinstance(text, string):
        return text.decode('utf-8')
    else:
        return unicode(text)
```

drothlis: Fixed. Thanks for the code, it worked. I added a test too.

PR #391, opened 2016-10-20
Add stbt control-relay

wmanley:
`stbt control-relay` allows you to use any stbt remote control using the lirc socket protocol. This for example allows you to control a roku via its HTTP REST API using `irsend`.

### DIFF
--- a/_stbt/control.py
+++ b/_stbt/control.py
@@ -51,6 +51,7 @@ def uri_to_remote_recorder(uri):
          lirc_remote_listen_tcp),
         (r'lirc:(?P<lircd_socket>[^:]+)?:(?P<control_name>.+)',
          lirc_remote_listen),
+        (r'lircd(:(?P<lircd_socket>[^:]+))?', fake_lircd_listen),
         (r'stbt-control(:(?P<keymap_file>.+))?', stbt_control_listen),
         (r'vr:(?P<address>[^:/]*)(:(?P<port>\d+))?', virtual_remote_listen),
     ]
@@ -498,7 +499,8 @@ def virtual_remote_listen(address, port=None):
 
 def lirc_remote_listen(lircd_socket, control_name):
     """Returns an iterator yielding keypresses received from a lircd file
-    socket.
+    socket -- that is, the keypresses that lircd received from a hardware
+    infrared receiver and is now sending on to us.
 
     See http://www.lirc.org/html/technical.html#applications
     """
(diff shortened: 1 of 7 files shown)


_stbt/control.py
```diff
@@ -67,6 +68,22 @@ def press(key):
         debug('NullRemote: Ignoring request to press "%s"' % key)
 
 
+class FileRemote(object):
```
drothlis: We want to be using the word "Control" rather than "Remote" in any new code.

_stbt/control.py
```diff
@@ -557,6 +558,24 @@ def lirc_key_reader(cmd_iter, control_name):
             yield m.group('key')
 
 
+def lircd_remote_listen(lircd_socket=None):
```
drothlis: We want to be using the word "control" instead of "remote" in any new code. Perhaps "fake_lircd_listen"?

_stbt/control.py
```diff
@@ -51,6 +51,7 @@ def uri_to_remote_recorder(uri):
          lirc_remote_listen_tcp),
         (r'lirc:(?P<lircd_socket>[^:]+)?:(?P<control_name>.+)',
          lirc_remote_listen),
+        (r'lircd(:(?P<lircd_socket>[^:]+))?', lircd_remote_listen),
```
drothlis: I wondered if this needs to be available to `--contol-recorder` -- it might be less confusing just to have it internally. But then I saw that your `stbt control-relay` can take any of these as inputs, which could be useful I suppose.

_stbt/control.py
drothlis: I found the difference between `lircd_remote_listen` and `lirc_remote_listen` quite confusing, so I've pushed a fixup to the docstrings and the man page.

wmanley: Yes, your changes are much clearer

wmanley: > Perhaps "fake_lircd_listen"?

Done

_stbt/control.py
wmanley: Fixed

wmanley:
Rebased to resolve conflicts and squashed

wmanley:
@drothlis: Merge?

stbt_control_relay.py
```diff
@@ -0,0 +1,55 @@
...
+Will write the text "KEY_UP" to the file `example`.
+
+    $ stbt control-relay --input=lircd:lircd.sock \\
+          roku:192.168.1.13 samsung:192.168.1.14
+
+Listens on lircd.sock and will forward keypresses to the roku at 192.168.1.13
+using its HTTP protocol and to the Samsung TV at 192.168.1.14 using its TCP
+protocol.  So
+
+    $ irsend -d lircd.sock SEND_ONCE stbt KEY_OK
```
drothlis: I guess the `stbt` argument is just any dummy value that will get ignored by non-lirc backends?

drothlis:
+1.

stbt_control_relay.py
drothlis: Actually it will get ignored by all backends because `fake_lircd_listen` only tells us the key. Each backend's configuration tells it which control to use (which lirc config file if it's a lirc backend, etc).

PR #392, opened 2016-10-20
Add stbt.Region.above/below/right_of/left_of

drothlis:
These are useful when you want to read text adjacent to some image:

    m = stbt.match("blue-button.png")
    print stbt.ocr(region=m.region.right_of(width=150))


_stbt/core.py
```diff
@@ -378,8 +389,73 @@ def translate(self, x=0, y=0):
...
+            x = self.x
+        if width is None and right is None:
+            right = self.right
+        if y is None:
+            if height is None:
+                y = -inf
+            else:
+                y = bottom - height
+                height = None
+        return Region(x, y, width, height, right, bottom)
```
drothlis: Good idea.

drothlis: Done. So much better!

PR #393, opened 2016-10-20
Hdmi cec control

wmanley:
This is an implementation of #365.  This implementation is based on the work by @danielandersson in #378, but has been fairly heavily modified and reworked such that it's mergable into stb-tester proper.

HdmiCecControl: Implemented CEC support using libCEC
In combination with a USB-CEC adaptor this allows stb-tester to send
keypresses to devices over HDMI-CEC.  This is useful for testing devices
which don't have an IR input like the Sony PlayStation 4.

The URI is of the form:

```
hdmi-cec:[<device>]:[<source>]:[<destination>]
```

If device is not specified then any available CEC adapter will be used.

Similar to the keymapping for Roku, a map is used to lookup the CEC
commands.  We raise `UnknownKeyError` if it's not one we've heard of.  The
list of keys is from http://www.cec-o-matic.com .  I've done my own
interpretation on how they should be mapped onto linux input keynames.

The tests are implemented by monkey-patching out the bits of libcec that
actually interact with the hardware.  I'm not a great fan of this style
of testing, but in the absence of hardware-in-the-loop testing this is
the best we can do.  At least some of this code will actually get run
through during `make check`.

The tests are particularly ugly because the SWIG generated libcec bindings
makes inspecting `cec_commmand` objects inconvenient so I've had to use
`ctypes` :(.

This is based on an initial prototype and research by Daniel Andersson
d.andersson92@gmail.com (@danielandersson).  See PR #378 for more
information.

TODO:
- [x] Release note
- [x] Documentation
- [x] Run tests on Travis
- [x] Review
- [x] Manual hardware-in-the-loop testing of final implementation 

Future work:

- Allow matching the device we want to talk to by osd name e.g. "PlayStation 4".  See the output of

        echo scan | cec-client -s

  for more information

### DIFF
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -1,2 +1,239 @@
+from contextlib import contextmanager
+from textwrap import dedent
+
+from .logging import debug
+
+
+class HdmiCecError(Exception):
+    pass
+
+
+class HdmiCecControl(object):
+    # Map our recommended keynames (from linux input-event-codes.h) to the
+    # equivalent CEC commands.
+    # The mapping between CEC commands and code can be found at
+    # http://www.cec-o-matic.com or in HDMI-CEC specification 1.3a
+    _KEYNAMES = {
+        "KEY_OK": 0,
+        "KEY_UP": 1,
+        "KEY_DOWN": 2,
+        "KEY_LEFT": 3,
+        "KEY_RIGHT": 4,
+        "KEY_RIGHT_UP": 5,
+        "KEY_RIGHT_DOWN": 6,
+        "KEY_LEFT_UP": 7,
+        "KEY_LEFT_DOWN": 8,
+        "KEY_ROOT_MENU": 9,
+        "KEY_SETUP": 10,
+        "KEY_CONTENTS_MENU": 11,  # <- not in input-event-codes.h
+        "KEY_FAVORITE_MENU": 12,  # <- not in input-event-codes.h
+        "KEY_BACK": 13,
+        "KEY_0": 32,
+        "KEY_1": 33,
+        "KEY_2": 34,
+        "KEY_3": 35,
+        "KEY_4": 36,
+        "KEY_5": 37,
+        "KEY_6": 38,
+        "KEY_7": 39,
+        "KEY_8": 40,
+        "KEY_9": 41,
+        "KEY_DOT": 42,
+        "KEY_ENTER": 43,
+        "KEY_CLEAR": 44,
+        "KEY_NEXT_FAVORITE": 47,  # <- not in input-event-codes.h
+        "KEY_CHANNELUP": 48,
+        "KEY_CHANNELDOWN": 49,
+        "KEY_PREVIOUS": 50,
+        "KEY_SOUND_SELECT": 51,  # <- not in input-event-codes.h
+        "KEY_INPUT_SELECT": 52,  # <- not in input-event-codes.h
+        "KEY_INFO": 53,
+        "KEY_HELP": 54,
+        "KEY_PAGEUP": 55,
+        "KEY_PAGEDOWN": 56,
+        "KEY_POWER": 64,
+        "KEY_VOLUMEUP": 65,
+        "KEY_VOLUMEDOWN": 66,
+        "KEY_MUTE": 67,
+        "KEY_PLAY": 68,
+        "KEY_STOP": 69,
+        "KEY_PAUSE": 70,
+        "KEY_RECORD": 71,
+        "KEY_REWIND": 72,
+        "KEY_FASTFORWARD": 73,
+        "KEY_EJECT": 74,
+        "KEY_FORWARD": 75,
+        "KEY_BACKWARD": 76,
+        "KEY_STOP_RECORD": 77,
+        "KEY_PAUSE_RECORD": 78,
+        "KEY_ANGLE": 80,
+        "KEY_SUB_PICTURE": 81,  # <- not in input-event-codes.h
+        "KEY_VOD": 82,
+        "KEY_EPG": 83,
+        "KEY_TIMER_PROGRAMMING": 84,  # <- not in input-event-codes.h
+        "KEY_CONFIG": 85,
+
+        # Not sure what the difference is between KEY_PLAY and KEY_PLAY_FUNCTION
+        # is but none of these _FUNCTION keys are in linux-event-codes.h:
+        "KEY_PLAY_FUNCTION": 96,
+        "KEY_PAUSE_PLAY_FUNCTION": 97,
+        "KEY_RECORD_FUNCTION": 98,
+        "KEY_PAUSE_RECORD_FUNCTION": 99,
+        "KEY_STOP_FUNCTION": 100,
+        "KEY_MUTE_FUNCTION": 101,
+        "KEY_RESTORE_VOLUME_FUNCTION": 102,
+        "KEY_TUNE_FUNCTION": 103,
+        "KEY_SELECT_MEDIA_FUNCTION": 104,
+        "KEY_SELECT_AV_INPUT_FUNCTION": 105,
+        "KEY_SELECT_AUDIO_INPUT_FUNCTION": 106,
+        "KEY_POWER_TOGGLE_FUNCTION": 107,
+        "KEY_POWER_OFF_FUNCTION": 108,
+        "KEY_POWER_ON_FUNCTION": 109,
+
+        # Back to normal keys.  We duplicate these colour buttons because HDMI
+        # doesn't make a distinction:
+        "KEY_F1": 113,
+        "KEY_BLUE": 113,
+        "KEY_F2": 114,
+        "KEY_RED": 114,
+        "KEY_F3": 115,
+        "KEY_GREEN": 115,
+        "KEY_F4": 116,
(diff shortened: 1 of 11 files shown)


wmanley:
@drothlis: I'm tempted to squash all these commits (except the first) into a single commit and credit @danielandersson in the commit message.  I'm not sure that all these separate commits actually buy us anything.

drothlis:
> I'm tempted to squash all these commits (except the first) into a single commit

Agreed.

wmanley:
Well, the tests aren't passing on Travis.  My best guess is that it's due to a different version of libcec.  Ubuntu 14.04 has `libcec2` and we install `python-cec` from pip.  Locally I'm testing on Debian with libcec3 and Debian's python-cec package where it passes.  Not sure what to do.  Tempted to skip the test if `import cec` fails.  Thoughts?

drothlis:
On Ubuntu 16.04 (which has libcec3):

| libcec              | python bindings           | test passes |
|---------------------|---------------------------|-------------|
| apt install libcec3 | apt install python-libcec | ✓           |
| apt install libcec3 | pip install cec==0.2.5    | ✗           |

drothlis:
Since we rarely run these tests locally I really want Travis to run them. We could build our own copy of libcec?

I'll have a go at this, since I already got it building locally.

drothlis:
Hold on, is it *just* the tests that fail, or will the CEC control not work at all on Ubuntu 14.04?

Does that mean that we can only build the `stb-tester-gpl` deb for Ubuntu 16.04?

wmanley:
> Hold on, is it just the tests that fail, or will the CEC control not work at all on Ubuntu 14.04?

Not sure, nor do I care very much.  If someone tries it on 14.04 and it doesn't work they can fix it.

drothlis:
At least it's in a separate deb package (coincidentally) so we don't have to handle the different dependencies when building for different Ubuntu releases -- we'll just build `stb-tester-gpl` for Ubuntu 16.04 only.

_stbt/control_gpl.py
```diff
@@ -0,0 +1,217 @@
...
+        if not self.lib.Open(device):
+            raise HdmiCecError("Failed to open a connection to the CEC adapter")
+        debug("Connection to CEC adapter opened")
+
+    def press(self, key):
+        from .control import UnknownKeyError
+        keycode = self._KEYNAMES.get(key)
+        if keycode is None:
+            raise UnknownKeyError("HdmiCecControl: Unknown key %r" % key)
+        cec_command = "%X%X:44:%02X" % (self.source, self.destination, keycode)
```
danielandersson: Wouldn't it be better to create a dictionary for those magic CEC numbers?

_stbt/control_gpl.py
```diff
@@ -0,0 +1,217 @@
...
+        key_down_cmd = self.lib.CommandFromString(cec_command)
+        key_up_cmd = self.lib.CommandFromString(
+            '%X%X:45' % (self.source, self.destination))
```
danielandersson: Same as the comment above, create a dictionary with magic CEC numbers.

SidePlace:
Hi @wmanley would you resolve the conflict and merge it? Thank you so much.

extra/pylint.conf
```diff
@@ -9,7 +9,7 @@ reports=no
 disable=C0102,C0103,C0111,C0112,C0121,C0202,C0203,C0302,C0321,C0322,C0323,C0324,C0330,E0012,E1103,I0011,I0012,R0201,R0401,R0801,R0901,R0902,R0903,R0904,R0911,R0912,R0913,R0914,R0915,R0921,R0922,R0923,W0142,W0232,W0603
 
 [TYPECHECK]
-ignored-classes=Buffer,Caps,Frame,Client,Event,gi.repository.Gst,MainLoop,numpy,numpy.linalg,Sample,scipy.interpolate,scipy.optimize,_socketobject
+ignored-classes=Buffer,Caps,cec,Frame,Client,Event,gi.repository.Gst,MainLoop,numpy,numpy.linalg,pytest,Sample,scipy.interpolate,scipy.optimize,_socketobject
```
SidePlace: Seems the only conflict is here, maybe I can help a bit? Just add cec and pytest into it right? Can we have this PR merged?
[pylint.zip](https://github.com/stb-tester/stb-tester/files/568551/pylint.zip)

drothlis:
I've rebased onto master to solve conflicts in pylint.conf.

@SidePlace the main thing remaining is to test the final implementation with real hardware (see the TODO list at the top of the pull request). It would be great if you could help test this by building this branch & testing against your own hardware.

_stbt/control_gpl.py
wmanley: I don't want to do this for the time being.  The magic CEC numbers are not in themselves interesting.  In the future I think it would be good to match devices by OSD name as this is more likely to be meaningful and stable.

This would involve doing an CEC scan to work out which device should be used.  This wouldn't be difficult and it can be added in a future PR in a backwards compatible way, but I don't have time to spend on this right now.

SidePlace:
Hi @drothlis  
I just gave a try and I think basically it's working. 
I can use navigation keys, ok, back and even PS button (KEY_ROOT_MENU/14:44:09). 
I also wrote a simple test case involving navigation keys, KEY_OK and stbt.wait_for_match, all working.

Some points worth mentioning:
1. I was testing on the 'stbt-control-relay2' branch as I can't find the codes under either branch master or branch hdmi-cec-control. 
2. Seems the make file didn't install correctly. After 'sudo make' and 'sudo make install', I cannot execute with 'stbt control'. Instead, I would have to cd to the source code and type './stbt-control'

I think we can merge.

wmanley:
Hi @SidePlace, Thanks for testing this

> 1. I was testing on the 'stbt-control-relay2' branch as I can't find the codes under either branch master or branch hdmi-cec-control. 
>
> ...
>
> I think we can merge.

We can't merge if this branch isn't working.  Perhaps I misunderstood.  What was the git SHA of the version you tested?

danielandersson:
@SidePlace For point 2 and 3, I think that you have installed stb-tester from apt-get and is using that version instead of the one you installed using make. You can test this by using `which stbt-control` to see where the executable is located.

drothlis:
> 1. I was testing on the 'stbt-control-relay2' branch as I can't find the codes under either branch master or branch hdmi-cec-control.

@SidePlace The hdmi-cec-control branch is on wmanley's branch, so you'll have to do something like this:

```
git remote add wmanley https://github.com/wmanley/stb-tester.git
git fetch wmanley
git checkout -b hdmi-cec-control wmanley/hdmi-cec-control
```

SidePlace:
I'm finally able to pull the codes and test against it. The test result this time is quite good. I can use it to control and execute test cases.

Hi @drothlis , thank you for your your help, those git commands really helped me a lot. I'm able to pull Wmanley's branch with that.
-----
Hi @wmanley , the SHA I'm testing this time is d1638479e7511b90e693464f5049a8c01333ca9b
-----
Hi @danielandersson , thank you for your reminder, but I don't know how to fix that. :( 
No respond when I type which stbt-control. But I can see it's installed on different places when I run 'whereis stbt' more details are:
`$:~/TestcasesForStb/ps4$ which stbt-control
$:~/TestcasesForStb/ps4$ which stbt
/usr/local/bin/stbt
$:~/TestcasesForStb/ps4$ which stb-tester
$:~/TestcasesForStb/ps4$ whereis stbt-control
stbt-control:
$:~/TestcasesForStb/ps4$ whereis stbt
stbt: /usr/bin/stbt /usr/lib/stbt /etc/stbt /usr/local/bin/stbt /usr/local/etc/stbt /usr/share/man/man1/stbt.1.gz
`

wmanley:
Great, thanks for the testing.  Merging now.

drothlis:
N.B. Since I rebased & reviewed 4 days ago, @wmanley squashed in the following changes (plus some test changes not shown):

```
diff --git a/_stbt/control_gpl.py b/_stbt/control_gpl.py
index e853a5d..934249a 100644
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -111,8 +112,13 @@ class HdmiCecControl(object):
             source = 1
         if destination is None:
             destination = 4
-        self.source = int(source, 16)
-        self.destination = int(destination, 16)
+        if isinstance(source, (str, unicode)):
+            source = int(source, 16)
+        if isinstance(destination, (str, unicode)):
+            destination = int(destination, 16)
+
+        self.source = source
+        self.destination = destination
 
         self.cecconfig = cec.libcec_configuration()
         self.cecconfig.strDeviceName = "stb-tester"
```

This is included in the commit d163847 that @SidePlace tested.

danielandersson:
Just to give an insight in how Android have implemented CEC support I'm posting two interesting links. I like how they do the CEC key mapping and how they support repeatable keys.

https://android.googlesource.com/platform/frameworks/base.git/+/master/services/core/java/com/android/server/hdmi/HdmiCecKeycode.java
https://android.googlesource.com/platform/frameworks/base.git/+/master/services/core/java/com/android/server/hdmi/SendKeyAction.java

drothlis:
I've soak-tested this implementation with the PS4 and there wasn't a single "missed keypress" in 55,000 keypresses.

PR #394, opened 2016-10-21
stbt auto-selftest generate: Accept optional source filenames

drothlis:
Even with our caching, running `stbt auto-selftest generate` on a large
test-pack can take several seconds. When you're iterating on a single
FrameObject it is useful to re-generate the selftests for just the file
you're working on.

Even with a small test-pack this reduces the run-time from 4 seconds to
1 second (when regenerating a single file).

### DIFF
--- a/stbt_auto_selftest.py
+++ b/stbt_auto_selftest.py
@@ -171,16 +171,20 @@ def init_worker():
 
 
 def iterate_with_progress(sequence, width=20, stream=sys.stderr):
-    ANSI_ERASE_LINE = '\033[K'
     stream.write('\n')
     total = len(sequence)
     for n, v in enumerate(sequence):
-        progress = (n * width) // total
-        stream.write(
-            ANSI_ERASE_LINE + '[%s] %8d / %d - Processing %s\r' % (
-                '#' * progress + ' ' * (width - progress), n, total, str(v)))
+        stream.write(_progress_line(width, n, total) +
+                     " - Processing %s\r" % v)
         yield v
-    stream.write('\n')
+    stream.write(_progress_line(width, total, total) + "\n")
+
+
+def _progress_line(width, n, total):
+    ANSI_ERASE_LINE = '\033[K'
+    progress = (n * width) // total
+    return '\r' + ANSI_ERASE_LINE + '[%s] %3d / %d' % (
+        '#' * progress + ' ' * (width - progress), n, total)
 
 
 def generate_into_tmpdir(source_files=None):
--- a/stbt_auto_selftest.py
+++ b/stbt_auto_selftest.py
@@ -175,8 +175,6 @@ def iterate_with_progress(sequence, width=20, stream=sys.stderr):
     stream.write('\n')
     total = len(sequence)
     for n, v in enumerate(sequence):
-        if n == total:
-            break
         progress = (n * width) // total
         stream.write(
             ANSI_ERASE_LINE + '[%s] %8d / %d - Processing %s\r' % (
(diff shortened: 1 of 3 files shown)


drothlis:
@wmanley Please review.

stbt_auto_selftest.py
```diff
@@ -92,23 +96,33 @@ def main(argv):
...
-        os.rename(tmpdir, target)
-    except:
-        shutil.rmtree(tmpdir)
-        raise
+        if source_files:
+            for f in valid_source_files:
+                target = os.path.join(os.curdir, "selftest/auto_selftest",
+                                      _selftest_filename(f))
+                if os.path.exists(target):
+                    os.remove(target)
```
wmanley: Why `remove` if you're about to `rename` over it anyway?

stbt_auto_selftest.py
```diff
@@ -92,23 +96,33 @@ def main(argv):
...
     try:
-        target = "%s/selftest/auto_selftest" % os.curdir
-        if os.path.exists(target):
-            shutil.rmtree(target)
-        os.rename(tmpdir, target)
-    except:
-        shutil.rmtree(tmpdir)
-        raise
+        if source_files:
+            for f in valid_source_files:
```
wmanley: Oh I see.  `valid_source_files` is a filtered version of `source_files`, it's not a list of all valid source files.

stbt_auto_selftest.py
```diff
@@ -92,23 +96,33 @@ def main(argv):
...
     elif cmdline_args.command == 'validate':
         return validate()
     else:
         assert False
 
 
-def generate():
-    tmpdir = generate_into_tmpdir()
+def generate(source_files):
+    valid_source_files = _valid_source_files(source_files)
```
wmanley: You probably want to error for any source files that don't exist.  You may wish to error for any source files that won't or don't produce an auto-selftest file.  This would probably be best for interactive use.

stbt_auto_selftest.py
```diff
@@ -163,16 +163,20 @@ def init_worker():
...
     for n, v in enumerate(sequence):
-        progress = (n * width) // total
-        stream.write(
-            ANSI_ERASE_LINE + '[%s] %8d / %d - Processing %s\r' % (
-                '#' * progress + ' ' * (width - progress), n, total, str(v)))
+        stream.write(_progress_line(width, n, total) +
+                     " - Processing %s\r" % v)
         yield v
-    stream.write('\n')
+    stream.write(_progress_line(width, total, total) + "\n")
```
wmanley: Yes, this is the kind of nice little polish that makes a tool feel mature.

stbt_auto_selftest.py
drothlis: I think I just got confused -- at some point I thought os.rename would fail if target file existed, but it must've been a permissions issue or something else. I've removed this.

stbt_auto_selftest.py
drothlis: I've removed the `valid_source_files` list now that I error early for missing files.

stbt_auto_selftest.py
drothlis: Good idea. I've added a check that the file exists. I still need to error for source files that have invalid python syntax, or that don't produce an auto-selftest file. For those cases I will need to do some bigger refactorings.

drothlis: I added a quick & easy check for invalid source files by running through the whole process, and checking if there is an auto-selftest output file.

PR #395, opened 2016-10-22
stbt lint: Compatibility with latest pylint/astroid version

drothlis:
The current version (pylint 1.6.4, astroid 1.5.0) has renamed
`astroid.node_classes.{CallFunc => Call}`. Luckily they have added an
alias for backwards compatibility; this alias is a proxy class that
implements `__instancecheck__` so `isinstance(..., CallFunc)` still
works.

I don't know why we were using `type(x) == X` instead of `isinstance`,
probably just an oversight.

TODO:

- [x] Compatibility with latest pylint (1.6.4)
- [x] Future compatibility with pylint 2.0
- [x] Test this change with Ubuntu 14.04's pylint 1.1.0
- [x] Test this change with Ubuntu 16.04's pylint 1.5.2
- [x] Release notes.


drothlis:
Note that we haven't been running pylint's `no-member` checker for at least 2 years! (see 4e9f79f for details). However re-enabling it didn't find any true positives.

PR #398, opened 2016-11-11
_stbt/control.py: Remove VirtualRemote implementation

drothlis:
This was a YouView-specific TCP-based remote control that YouView never
got around to releasing; I don't think even they use it any more.


drothlis:
@LewisHaley you're one of the few people who *might* be using this -- would removing this affect you?

LewisHaley:
No.

PR #399, opened 2016-11-11
stbt-control-relay: Add --verbose parameter

drothlis:
So that we can see debug output from each of the remote-control
implementations that are sending output.

Also require at last 1 output control to be specified.


drothlis:
@wmanley please review.

PR #400, opened 2016-11-25
stbt-control: Truncate long remote control URI to 80 characters

szemate:
`stbt control` raises the error "Unable to print keymap because the
terminal is too small" if any of the printed lines is longer than the
width of the terminal. If the remote control URI is very long then it
is becoming difficult to launch the app on small screens. (The terminal
has to be made full screen and/or the font size has to be decreased.)


drothlis:
Merged. Thanks!

PR #401, opened 2016-12-13
irnetbox-proxy: set SO_REUSEADDR on listening socket

lmaneos:
After the service had terminated due to a dropped connection from
the irnetbox, restarting it was failing for several minutes with:

irnetbox-proxy: Traceback (most recent call last):
irnetbox-proxy: File "/usr/local/bin/irnetbox-proxy", line 225, in connect
irnetbox-proxy: self.listen_sock.bind(self.listen_addr)
irnetbox-proxy: File "/usr/lib64/python2.7/socket.py", line 224, in meth
irnetbox-proxy: return getattr(self._sock,name)(*args)
irnetbox-proxy: error: [Errno 98] Address already in use
irnetbox-proxy: Could not bind to local address

because the listening socket was left in TIME_WAIT state.  Set
the SO_REUSEADDR option on the socket to avoid this.

### DIFF
--- a/irnetbox-proxy
+++ b/irnetbox-proxy
@@ -222,6 +222,8 @@ class IRNetBoxProxy(object):
     def connect(self):
         try:
             self.listen_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            self.listen_sock.setsockopt(socket.SOL_SOCKET,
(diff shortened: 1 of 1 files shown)


drothlis:
Merged. Thanks!

PR #402, opened 2016-12-14
HDMI CEC: Automatically choose device to talk to

wmanley:
TODO:

- [x] Update man page.

### DIFF
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -110,15 +110,10 @@ class HdmiCecControl(object):
         import cec
         if source is None:
             source = 1
-        if destination is None:
-            destination = 4
         if isinstance(source, (str, unicode)):
             source = int(source, 16)
         if isinstance(destination, (str, unicode)):
             destination = int(destination, 16)
-
-        self.source = source
-        self.destination = destination
 
         self.cecconfig = cec.libcec_configuration()
         self.cecconfig.strDeviceName = "stb-tester"
@@ -137,6 +132,21 @@ class HdmiCecControl(object):
         if not self.lib.Open(device):
             raise HdmiCecError("Failed to open a connection to the CEC adapter")
         debug("Connection to CEC adapter opened")
+
+        if destination is None:
+            ds = list(self._list_active_devices())
+            debug("HDMI-CEC scan complete.  Found %r" % ds)
+            if len(ds) == 0:
+                raise HdmiCecError(
+                    "Failed to find a device on the CEC bus to talk to.")
+            # Choose the last one, the first one is likely to be a TV if there's
+            # one plugged in
+            destination = ds[-1]
+            debug("HDMI-CEC: Chose to talk to device %i %r" % (
+                destination, self.lib.GetDeviceOSDName(destination)))
+
+        self.source = source
+        self.destination = destination
 
     def press(self, key):
         from .control import UnknownKeyError
@@ -171,6 +181,17 @@ class HdmiCecControl(object):
             retval = adapter.strComName
         return retval
 
+    def _list_active_devices(self):
+        self.lib.RescanActiveDevices()
+        active = self.lib.GetActiveDevices()
+
+        # We get a fixed size array back.  libcec-python doesn't implement
+        # iteration or bounds checking:
+        for n in range(16):
+            # active.primary is us
+            if n != active.primary and active[n]:
+                yield n
+
 
 def test_hdmi_cec_control():
     from .control import uri_to_remote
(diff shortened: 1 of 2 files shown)


wmanley:
One limitation of this implementation is that if the device isn't there when we create `HdmiCecControl` we will fail, even though it might be a transient failure e.g. due to the DUT rebooting.

wmanley:
Manual hardware-in-the-loop test results:

```
$ ipython
Python 2.7.12+ (default, Nov 22 2016, 00:48:54) 
Type "copyright", "credits" or "license" for more information.

IPython 4.1.1 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import _stbt.logging

In [2]: _stbt.logging._debug_level = 2

In [3]: from  _stbt.control import uri_to_remote

In [4]: c = uri_to_remote("hdmi-cec:")
ipython: libCEC version 3.1.0 loaded: compiled on Linux-3.16.0-4-amd64 ... , features: P8_USB, DRM, P8_detect, randr
ipython: Found a CEC adapter:"
Port:     /dev/ttyACM0
Vendor:   2548
Product:  1002
ipython: Connection to CEC adapter opened
ipython: HDMI-CEC scan complete.  Found [4]
ipython: HDMI-CEC: Chose to talk to device 4 'PlayStation 3'

In [5]: c.press('KEY_RIGHT')
ipython: cec: Transmit KEY_RIGHT as 14:44:04
```

`KEY_RIGHT` was transmitted successfully.

wmanley:
@drothlis: Please review

wmanley:
> One limitation of this implementation is that if the device isn't there when we create HdmiCecControl we will fail, even though it might be a transient failure e.g. due to the DUT rebooting.

An alternative would be to try the lookup on startup but if it fails carry on, trying again for each key-press.  Each keypress before we know who to talk to would fail until we find a candidate.

drothlis:
Looks good to me. Need to update the documentation in `docs/stbt.1.rst` though.

> An alternative would be to try the lookup on startup but if it fails carry on, trying again for each key-press. Each keypress before we know who to talk to would fail until we find a candidate.

What would a keypress failure look like? An exception? At least you'd be able to catch the exception in the test script (when you call `press`) unlike the current initialisation exception which is called in `stbt run` before it gets to the test script.

But given that the PS4 can't be controlled after rebooting until you press a button on the physical controller, I think that it's OK as is.

drothlis:
In `stbt.1.rst`:

>     hdmi-cec:[<device>]:[<source>]:[<destination>]

I don't think that's quite right either. It should be:

    hdmi-cec[:<device>[:<source>[:<destination>]]]

The implementation doesn't actually let you leave one of the intermediate things blank, e.g. `hdmi-cec:::<destination>` isn't allowed. See the regex in `control_gpl.py`:

>     r'hdmi-cec(:(?P<device>[^:]+)(:(?P<source>[^:]+)(:(?P<destination>[^:]+))?)?)?'

drothlis:
I mean in the man page the following needs to be updated:

> destination defaults to 4 (Playback 1).

but while you're there you might as well fix the `hdmi-cec[:<device>[:<source>[:<destination>]]]` as well.

danielandersson:
@drothlis 
> But given that the PS4 can't be controlled after rebooting until you press a button on the physical controller, I think that it's OK as is.

What do you mean? You need to press a button on the PS4 controller in order to control it over CEC?
@SidePlace did you know about this?

drothlis:
@danielandersson Yes at the login screen you need to press a button on the PS4 controller to log in; CEC doesn't work.

SidePlace:
Hi @danielandersson thank you for the mentioning.

> at the login screen you need to press a button on the PS4 controller to log in; CEC doesn't work.
@drothlis do you mean the ps button (button H in [this image](https://www.playstation.com/en-gb/content/dam/get-help/help-library/scee/accessories/using-dualshock-4/using-dualshock-bespoke-ps4-01-eu-22jul15.jpg)) on the controller?
If so, you don't have to, the cec command you can use for that is 'tx 14:44:09', you don't need a real controller to at all.

drothlis:
Thanks @SidePlace. You're right, login via CEC does work, I just tested it. My confusion was because it doesn't work if you wake the PS4 by pressing the PS button on the physical controller -- then you have to log in (select the user) using the X button on the controller. But if you wake the PS4 using CEC, or by pressing the power button on the PS4 itself, then you can log in with CEC.

PR #404, opened 2016-12-30
stbt-control-relay: Report success/error of relayed keypress

drothlis:


drothlis:
@wmanley please review.

stbt_control_relay.py
```diff
@@ -18,46 +18,84 @@
...
-            r.press(key)
-        except Exception as e:  # pylint: disable=broad-except
-            sys.stderr.write("Error pressing key %r: %s\n" % (key, e))
+    s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    s.bind(args.socket)
+    s.listen(5)
+
+    control = uri_to_remote(args.output)
+
+    while True:
```
wmanley: This is no-longer using a thread-pool so it won't be able to cope with multiple connections and a single connection could cause the daemon to hang.

wmanley: This isn't a new issue actually.

wmanley: Yes, the threadpool was just on the sending side.  The receiving side was always blocking.  To fix this we could derive from [`SocketServer.TCPServer`](https://docs.python.org/2/library/socketserver.html) or spin each connection off onto a `ThreadPool`.

This won't be an issue with `stbt` or `irsend` because both use short-lived connections sending a single keypress.

wmanley:
> Fixing this required .. removing ... relaying to more than one remote control simultaneously

I don't think this is necessarily true, although I have no objection as it's not clear that multiple outputs would ever be useful.

wmanley:
Yep, I'm happy with this for the time being.  It's become a narrower tool, but it supports the common/intended use-case better than before making it overall more useful.

PR #405, opened 2016-12-30
wait_until: Add `stable_secs` parameter

drothlis:
If specified, waits for the return value to stabilise. This is useful
when you're waiting for an animation to complete, for a selection to
finish moving, etc.


drothlis:
@wmanley please review.

drothlis:
Rebased onto master to fix conflict in release notes.

_stbt/core.py
```diff
@@ -1183,10 +1183,11 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0):
     import time
     expiry_time = time.time() + timeout_secs
     while True:
+        t = time.time()
         value = callable_()
         if value:
             return value
-        if time.time() > expiry_time:
+        if t > expiry_time:
```
wmanley: Neat

_stbt/core.py
```diff
@@ -1128,7 +1128,7 @@ def save_frame(image, filename):
     cv2.imwrite(filename, image)
 
 
-def wait_until(callable_, timeout_secs=10, interval_secs=0):
+def wait_until(callable_, timeout_secs=10, interval_secs=0, consecutive_secs=0):
```
wmanley: Should be `stable_secs`.

_stbt/core.py
```diff
@@ -1137,6 +1137,15 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0):
     truthy value) or until ``timeout_secs`` seconds have passed. In both cases,
     ``wait_until`` returns the value that ``callable_`` returns.
 
+    If ``consecutive_secs`` is specified then ``wait_until`` returns:
+
+    * ``callable_``'s return value if it was truthy and remained the same
+      (as determined by ``==``) for ``consecutive_secs`` seconds.
+    * ``None`` if ``callable_``'s return value was truthy but didn't stay
+      the same for ``consecutive_secs`` before we reached ``timeout_secs``.
```
wmanley: This is a little ugly.

wmanley: An alternative: Return a `TimeoutExpired` type:

```
class TimeoutExpired(object):
    def __init__(self, last_value):
        self.last_value = last_value

    def __nonzero__(self):
        return False
```

wmanley: or perhaps `NotStable`

_stbt/core.py
```diff
@@ -1178,18 +1187,38 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0):
...
+    stable_value = NoValue
+    stable_since = None
     expiry_time = time.time() + timeout_secs
+
     while True:
         t = time.time()
         value = callable_()
-        if value:
+
+        if value != stable_value:
```
wmanley: If we're getting `Falsey` values I think this will compare them even if we're not interested in `consecutive_secs`.  I don't think this will be a problem for `FrameObject`s as all `False`y `FrameObject`s of the same concrete type are equal and only `is_visible` will be invoked, and the result of that should be cached.  I don't know if this could be a problem with other types.  Probably fine.

tests/test_core.py
```diff
@@ -32,3 +33,10 @@ def t():
 
 def test_that_wait_until_with_zero_timeout_tries_once():
     assert stbt.wait_until(lambda: True, timeout_secs=0)
+
+
+def test_wait_until_consecutive_secs():
```
wmanley: Could probably do with some more tests here.  `wait_until` is becoming a little complex with multiple interacting parts.  It's good to centralise this complexity in one place but it's also good to sure how it works.

Perhaps some tests that mock out `time.time`?  Something like:

```
@contextmanager
def faketime_fn(sequence):
    it = iter(sequence)
    t = [0]
    def fake_time():
        return t[0]
    def fn():
        v = it.next()
        t[0] += 0.1
        return v
    with mock.mock("time.time", faketime):
        yield fn

def test():
    with faketime_fn([0,1,2,3,4,5,5,5,5,5,6,7,8,9,10]) as fn:
        assert wait_until(fn, consecutive_secs=0.5)
    with faketime_fn([0,1,2,3,4,5,5,5,5,5,6,7,8,9,10]) as fn:
        assert not wait_until(fn, consecutive_secs=0.6)
```

wmanley: Oh, I see there are more tests in the next commit.

tests/test_core.py
```diff
@@ -1,42 +1,96 @@
...
+
 
-    assert stbt.wait_until(t().next, timeout_secs=0.1)
+def test_that_wait_until_tries_one_last_time_with_consecutive_secs():
+    #                      consecutive_secs reached
+    #           Timeout reached     |
+    #                         |     |
+    #                         v     v
+    assert wait_until(f("a b b b b b b b b b b"),
+                      timeout_secs=0.03, consecutive_secs=0.05) is None
```
wmanley: This comes as a surprise to me.  I was expecting that the semantics was it must have got stable within the timeout, but they are it must have begun getting stable within the timeout.  Not staying that the current semantics are wrong, just surprising to me.

tests/test_core.py
```diff
@@ -1,42 +1,96 @@
...
+
+
+def test_wait_until_false_value():
+    assert wait_until(f("F F F"), timeout_secs=0.1) is False
+    assert wait_until(f("F F F"), timeout_secs=0.1, consecutive_secs=0.01) \
+        is None
+
+
+def test_wait_until_falsey_value():
+    class NonZero(object):
```
wmanley: It's called `NonZero` but it is `Zero`.

tests/test_core.py
```diff
@@ -1,42 +1,96 @@
...
+    assert wait_until(f("a b b")) == "a"
+    assert wait_until(f("a b b"), timeout_secs=0.1, consecutive_secs=0.01) == "b"
+    assert wait_until(f("a b c"), timeout_secs=0.1, consecutive_secs=0.01) \
+        is None
+
+
+def test_wait_until_false_value():
+    assert wait_until(f("F F F"), timeout_secs=0.1) is False
+    assert wait_until(f("F F F"), timeout_secs=0.1, consecutive_secs=0.01) \
+        is None
```
wmanley: Some of these tests make me a little nervous.  I do wonder if mocking out `time.time` and `time.sleep` would be better.

_stbt/core.py
```diff
@@ -490,6 +490,19 @@ def __repr__(self):
...
+
+    def __eq__(self, other):
+        return (self.match == other.match and self.region == other.region and
+                self.image == other.image)
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __hash__(self):
+        return hash((self.match, self.region, self.image))
```
wmanley: I have to admit, this makes me a little nervous.  I can't think of any particular scenario that's someone's likely to do in practice where it would go wrong but it would certainly be a surprising if you were to discover these semantics accidentally.

We're a bit cheeky and override equality in `FrameObject`s too, ignoring the `_frame` and other protected members of the class even though sometimes we assume their existence, but those members are at least protected.

Maybe it's fine.  These semantics certainly seem useful - more useful in practice than including the `first_pass_result`.  Maybe the real mistake was making `frame` a member at all, but we're unlikely to do anything about that at this point.

docs/release-notes.md
```diff
@@ -22,6 +22,10 @@ UNRELEASED
 
 ##### Breaking changes since v26
 
+* The definition of equality for `stbt.MatchResult` objects has changed (see
+  below for details). This should only affect you if you were storing
+  `MatchResult` objects in a dict or a set.
```
wmanley: in a set or as the keys of a `dict`

tests/test_core.py
wmanley: My mistake, I missed the `is None`.

wmanley:
Review complete

tests/test_core.py
drothlis: Done.

drothlis:
Rebased onto master to fix conflicts in release notes.

drothlis:
I've reworked the selftests to mock `time.time` and `time.sleep`, and I've tidied up the git history.

_stbt/core.py
drothlis: Note that `MatchResult.image` can be a string, or a `numpy.ndarray` (whatever you passed to `stbt.match`). For example you can try to match an image that you saw earlier, instead of matching an image loaded from disk. In that case the `MatchResult` won't be hashable -- whereas before, it would have Python's default hash value, based on the object's id (its address in memory).

_stbt/core.py
```diff
@@ -490,6 +490,19 @@ def __repr__(self):
                 "<Custom Image>" if isinstance(self.image, numpy.ndarray)
                 else repr(self.image)))
 
+    def __nonzero__(self):
+        return self.match
+
+    def __eq__(self, other):
+        return (self.match == other.match and self.region == other.region and
+                self.image == other.image)
```
drothlis: Similarly, if `MatchResult.image` is a `numpy.ndarray` this will result in a numpy array of bools, which will raise ValueError if you try to use it as a boolean ("The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()").

drothlis: I'm removing `MatchResult.__eq__`, `__ne__` and `__hash__` in #443.

PR #406, opened 2016-12-30
stbt lint: Check that frameobjects pass self._frame

drothlis:
...to stbt.match etc.

TODO:

- [x] Only if call is inside a frameobject member
- [x] Only inside properties
- [x] Test with older pylint?
- [x] Tests.

### DIFF
--- a/_stbt/pylint_plugin.py
+++ b/_stbt/pylint_plugin.py
@@ -86,7 +86,7 @@ class StbtChecker(BaseChecker):
                 else:
                     self.add_message('E7003', node=node, args=arg.as_string())
 
-        if _in_frameobject(node):
+        if _in_frameobject(node) and _in_property(node):
             for funcdef in node.func.infer():
                 if (isinstance(funcdef, FunctionDef) and
                         funcdef != YES and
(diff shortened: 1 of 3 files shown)


wmanley:
I really like the idea of this.  It should only apply to `@property`s though rather than all methods.

Another useful lint would be checking referring to other public properties from `is_visible`, but increasingly I am coming round to your viewpoint that  we should just fix that so it's not a problem.

_stbt/pylint_plugin.py
```diff
@@ -80,6 +86,30 @@ def visit_callfunc(self, node):
...
+            for funcdef in node.func.infer():
+                if (isinstance(funcdef, FunctionDef) and
+                        funcdef != YES and
+                        "frame" in funcdef.argnames()):
+                    index = funcdef.argnames().index("frame")
+                    if node.keywords:
+                        kwargs = [k.arg for k in node.keywords]
+                    else:
+                        kwargs = []
+                    if len(node.args) <= index and "frame" not in kwargs:
```
wmanley: This is elegant.  I like how you don't have to hard-code the list of functions to check against.

On a slightly related note I wonder if df9009eb6f6f8ffb1ebfc20e2e7bc5a850574667 went too far.  Perhaps `match()` should take a `dut=` parameter rather than be a member of `DeviceUnderTest`.  We could pare `DeviceUnderTest` back.

wmanley: We would then have to make `DeviceUnderTest` public.  Hmm, could be reasonable, not sure what it would look like.  Could provide a bunch of wrapping methods like `press()` and `frames()`, or just provide access to a `Remote` and a `Display` object.

drothlis:
Force-pushed a fix so that it works with astroid < 1.4 now (e.g. astroid 1.0.1 from Ubuntu 14.04).

drothlis:
Added a commit that makes it check only FrameObject properties.

PR #409, opened 2017-01-16
_stbt/control.py: Add "error" remote control 

drothlis:
It raises `RuntimeError` when the test script calls `stbt.press`.

If you don't want to use any of the built-in remote controls, this
allows you to catch unintended uses of `stbt.press` in your test script,
which would be silently ignored if you used the `none` remote control.


wmanley:
There's an argument for this becoming the default remote control too, but we can worry about this elsewhere.

PR #410, opened 2017-01-16
Change the default remote control to "error"

drothlis:
If your test calls `stbt.press` but you haven't configured a remote
control, it should fail noisily instead of silently swallowing the
keypress.


wmanley:
:+1:

PR #412, opened 2017-01-24
stbt lint: Catch astroid inference exceptions

drothlis:
I don't think this is a regression in v27 because we were using `infer`
for some of the checkers before then. Maybe it's a change in newer
astroid versions.

When you had an undefined variable, `stbt lint` would fail with the
following exception:

>   File "/libexec/stbt/_stbt/pylint_plugin.py", line 104, in _is_callable
>     for inferred in node.infer():
>   File "/lib/python2.7/site-packages/astroid/bases.py", line 302, in wrapped
>     for res in _func(node, context, **kwargs):
>   File "/lib/python2.7/site-packages/astroid/inference.py", line 84, in infer_name
>     raise exceptions.UnresolvableName(self.name)
> astroid.exceptions.UnresolvableName: <name of undefined variable>


drothlis:
Rebased onto master to fix conflict in release notes.

Issue #413, opened 2017-01-25
irnetbox.py does not support "double signals" config files


rinaldomerlo:
Files generated by RedRat irnetbox signal utility cannot be correctly parsed if they are of the "double signals" variety.

rinaldomerlo:
pull request #414 submitted for this.

PR #414, opened 2017-01-25
Update irnetbox.py for double signals configs (issue #413)

rinaldomerlo:
Add the ability to parse RedRat irnetbox config files with "double signals".


rinaldomerlo:
Tested on my own machine. Did not break existing behaviour with old single signal (MOD_SIG) style file, and seemed to work fine with double signal (DMOD) file (after change in 2nd commit).

drothlis:
Merged via #415. Thanks Rinaldo!

PR #417, opened 2017-02-21
Add `stbt.android` with control & screenshot mechanism for Android devices

drothlis:
See the docstring in `stbt/android.py` for details.

TODO:

- ~~Make a separate `stbt-mobile` package~~
- [x] Install it.
- [x] Release notes.
- [x] Take default values for `AdbDevice` constructor from `stbt.conf`.
- [x] Integration tests with a real phone. (Doesn't run on Travis)
- [x] Test the rendering of the rst documentation.
- ~~Obey `$ANDROID_SERIAL` just like adb client does?~~ I suppose we do implicitly: If we don't specify `adb_device` in the constructor or stbt.conf, we call `adb` without `-s` so adb will obey `$ANDROID_SERIAL`.
- ~~Rename `adb_device` constructor parameter to `adb_serial` or `android_serial`?~~ Nah. I've improved the documentation though.
- [x] Rename `screenshot` to `get_frame` for consistency with `_stbt.core.DeviceUnderTest`?
- [x] Add `press` for pressing HOME & BACK buttons for phones with hardware buttons (not on-screen buttons).
- [x] Soak test a stbt test over adb tcpip.
- [x] Don't log invalid screenshot data.
- ~~Retry screenshot if times out over tcpip?~~ Maybe later.


drothlis:
`AdbDevice.devices()` is an odd one out, because it isn't about a specific Android device, it's just talking to adb to list all available devices.

It's convenient to have it in `AdbDevice` because `AdbDevice` knows the necessary paramaters to talk to the adb server (`adb_server` (the hostname) and `adb_binary`).

wmanley:
> AdbDevice.devices() is an odd one out, because it isn't about a specific Android device, it's just talking to adb to list all available devices.

Yes.  If you really wanted to fix this you could have a seperate `Adb` class representing a connection to an adb server, but given we're just wrapping a command line tool I don't think it matters.

stbt/android.py
```diff
@@ -0,0 +1,406 @@
...
+
+from __future__ import division
+
+import re
+import subprocess
+import sys
+import time
+from collections import namedtuple
+
+import cv2
```
wmanley: I'd prefer if you just import cv2 in the screenshot function.  That way this module is usable without that dependency.

stbt/android.py
```diff
@@ -0,0 +1,406 @@
...
+import numpy
```
wmanley: Same is true of `numpy`

stbt/android.py
```diff
@@ -0,0 +1,406 @@
...
+
+        w, h = img.shape[1], img.shape[0]
+        stbt.debug("Screenshot is %d px wide x %d px high" % (w, h))
+
+        if self.coordinate_system == CoordinateSystem.ADB_720P:
+            # Resize to 720p preserving orientation
+            if w > h:
+                img = cv2.resize(img, (1280, 720))
+            else:
+                img = cv2.resize(img, (720, 1280))
```
drothlis: I wonder if we should make `screenshot` resize --and rotate portait images-- if CoordinateSystem.CAMERA_720P. That way the screenshots would be compatible with the screenshots you're getting from `stbt.get_frame()`.

On the other hand it can be a useful debugging tool to get the ADB screenshots in the same orientation as the device. But you could always instantiate a different `AdbDevice` with `CoordinateSystem.ADB_NATIVE` for those debugging screenshots.

stbt/android.py
```diff
@@ -0,0 +1,406 @@
...
+                    .replace("\r\n", "\n"))
+            img = cv2.imdecode(
+                numpy.asarray(bytearray(data), dtype=numpy.uint8),
+                cv2.CV_LOAD_IMAGE_COLOR)
+            if img is None:
+                sys.stderr.write(
+                    "AdbDevice.screenshot: error: Failed to get screenshot "
+                    "via adb (attempt %d/10)\n" % attempt)
+                sys.stderr.write("Length of data: %d\n" % len(data))
+                sys.stderr.write("Data: %r\n" % data)
```
wmanley: I've been coming round to using `logging.warning` for this kind of thing. It will end up on stderr by default anyway, but it gives us additional options too like structured logging.  So this would be:

    logging.warning("AdbDevice.screenshot: error: Failed to get screenshot "
                    "via adb (attempt %d/10)\n"
                    "Length of data: %d\n"
                    "Data: %r", attempt, len(data), data)

stbt/android.py
```diff
@@ -0,0 +1,406 @@
...
+            return _to_native_coordinates(
+                x, y, self.coordinate_system, self._get_display_dimensions())
+
+    def _get_display_dimensions(self):
+        return _parse_display_dimensions(
+            self.adb(["shell", "dumpsys", "window"],
+                     timeout_secs=10, capture_output=True))
+
+
+class AdbError(Exception):
```
wmanley: Should this subclass `subprocess.CalledProcessError`?

stbt/android.py
```diff
@@ -0,0 +1,406 @@
...
+            else:
+                break
+        else:
+            raise RuntimeError(
+                "Failed to capture screenshot from android device")
+
+        w, h = img.shape[1], img.shape[0]
+        stbt.debug("Screenshot is %d px wide x %d px high" % (w, h))
+
+        if self.coordinate_system == CoordinateSystem.ADB_720P:
```
wmanley: So it's clear what we should be doing here in `ADB_720P` mode and `ADB_NATIVE`, what about `HDMI_720P` and `CAMERA_720P`?

stbt/android.py
```diff
@@ -0,0 +1,406 @@
...
+            # Resize to 720p preserving orientation
+            if w > h:
+                img = cv2.resize(img, (1280, 720))
+            else:
+                img = cv2.resize(img, (720, 1280))
+
+        return img
```
wmanley: Would there be any value in returning a `stbt.Frame` here?  We would have to do timestamping which would probably only be accurate to <0.5s or so.  Probably not worth it at this point, we can be on the lookout for usecases.

tests/test_android.py
```diff
@@ -0,0 +1,124 @@
...
+        coordinate_system=CoordinateSystem.ADB_NATIVE)
+    if not any(
+            serial in _adb.devices() for serial in [
+                "7278681B045C937CEB770FD31542B16",  # Our CI Samsung Galaxy Ace2
+                os.environ.get("ANDROID_SERIAL", None),
+            ] if serial is not None):
+        raise SkipTest(
+            "CI Android device not connected. You can set ADB_SERVER, "
+            "ANDROID_SERIAL, and ADB_BINARY environment variables.")
+    return _adb
```
wmanley: Nice tests.

tests/test_android.py
```diff
@@ -0,0 +1,124 @@
...
+
+          Display: init=480x800 cur=480x800 app=480x800 rng=480x442-800x762
+          [...]
+        """)
+    assert _parse_display_dimensions(samsung_galaxy_ace_2) == \
+        _Dimensions(width=480, height=800)
+
+
+def test_screenshot_tap_and_swipe(adb):  # pylint:disable=redefined-outer-name
+    adb.adb(["shell", "input", "keyevent", "KEYCODE_HOME"])
```
wmanley: In the future we should consider adding a `press` function or similar for this usecase.

stbt/android.py
```diff
@@ -0,0 +1,406 @@
...
+        if self.adb_server:
+            _command += ["-H", self.adb_server]
+        if self.adb_device:
+            _command += ["-s", self.adb_device]
+        _command += command
+        stbt.debug("AdbDevice.adb: About to run command: %r\n" % _command)
+        output = subprocess.check_output(
+            _command, stderr=subprocess.STDOUT, **kwargs)
+        if command[0] in ("connect", "tcpip", "usb"):
+            time.sleep(5)
```
wmanley: What's up with this?  This feels like a layering violation.

stbt/android.py
```diff
@@ -0,0 +1,406 @@
...
+        x, y = _region_to_tuple(position)
+        stbt.debug("AdbDevice.tap((%d,%d))" % (x, y))
+
+        x, y = self._to_native_coordinates(x, y)
+        self.adb(["shell", "input", "tap", str(x), str(y)], timeout_secs=10)
+
+    def _adb(self, command, timeout_secs=None, **kwargs):
+        _command = []
+        if timeout_secs is not None:
+            _command += ["timeout", "%fs" % timeout_secs]
```
wmanley: It's a shame that `subprocess` is so limited that we need to use an additional helper binary for this.  Incidentally Python 3.5 has fixed this with their new [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run).

stbt/android.py
wmanley: Sounds reasonable to me if it's not too much extra work.

wmanley:
> - [ ] Make a separate stbt-mobile package?

What's the motivation for this?

drothlis:
> What's the motivation for [Make a separate stbt-mobile package]

1. Avoid adding dependencies like `adb` on the standard stb-tester package.
2. Allow installing it without installing stb-tester.

But now that I think about it:

Re (1), we might not want to add a dependency on the Ubuntu adb package, so that users can install whatever version they want (the adb client and adb server version must match exactly, and users may want to plug their phones into a remote adb server, or integrate with Appium which manages its own adb version).

Re (2), I'm not even sure how that would work, given that you need a `stbt/__init__.py` file present if you wanted to install this package and have `import stbt.android` work; the `stbt/__init__.py` would conflict with the one from the main stb-tester package.

So I'm tempted to conclude that it isn't worth making a separate package, and we'll just stuff it into the main stb-tester package.

stbt/android.py
drothlis: It's the only way I could get `test_AdbDevice_tcpip` to pass. `adb tcpip` and `adb usb` both restart the adbd process on the phone). The test has passed 80/80 times. Without any sleep the test fails almost every time. With a 1 second sleep it fails 2/10 times. This was with a modern phone (Moto X 2nd gen); on our old CI phone (Galaxy Ace 2) `adb tcpip` doesn't seem to work.

Here's how I ran the test:

```
ANDROID_SERIAL=model:XT1092 PYTHONPATH=`pwd` py.test -vv -rs -x --count=100 tests/test_android.py -k test_AdbDevice_tcpip
```

wmanley:
> So I'm tempted to conclude that it isn't worth making a separate package, and we'll just stuff it into the main stb-tester package.

Sounds reasonable to me.  It depends on `stbt.Region` anyway.

stbt/android.py
wmanley: Well then I think the sleep belongs in `_connect` and the `test_AdbDevice_tcpip` rather than here.

stbt/android.py
drothlis: `import stbt` isn't going to work if you don't have cv2 & numpy, is it?

stbt/android.py
drothlis: So with HDMI_720P it should also shrink portrait images to 720px high, and add letterboxing on the sides?

stbt/android.py
drothlis: Yeah, I suppose in practice you'll never be using `adb tcpip` and `adb usb` from test scripts.

stbt/android.py
drothlis: Hmmm *all* of the subprocess functions have `timeout` since Python 3.3, and the `subprocess32` backport includes this. Unfortunately `subprocess32` isn't available in the Ubuntu 12.04 or 14.04 repos.

drothlis: I suppose we could copy `subprocess32.py` into our repo. It's a self-contained 1900 line file.

stbt/android.py
drothlis: Done.

stbt/android.py
wmanley: Sounds like a bit of a pain.  Perhaps for now it should raise `NotImplementedError`.

stbt/android.py
wmanley: I believe it has a native component too.

drothlis: You're right: https://github.com/google/python-subprocess32/blob/master/subprocess32.py#L465-L472

stbt/android.py
```diff
@@ -0,0 +1,407 @@
...
+            self.cmd, self.returncode, self.output, self.adb_devices)
+
+
+def _region_to_tuple(r):
+    if isinstance(r, stbt.Region):
+        return ((r.x + r.right) // 2, (r.y + r.bottom) // 2)
+    elif isinstance(r, tuple) and len(r) == 2:
+        return r
+    else:
+        raise ValueError("Expected stbt.Region or (x,y) tuple but got %r" % r)
```
wmanley: I think this should be `TypeError`.

An alternative implementation that takes advantage of Python's duck typing might look like:

```
try:
    return (r.x + getattr(r, "width", 0) // 2, r.y + getattr(r, "height", 0) // 2)
except AttributeError:
    pass
try:
    x, y = r
    return (x, y)
except ValueError:
    raise TypeError("Expected stbt.Region or (x, y) coordinates but got %r", r)
```

stbt/android.py
drothlis: Or maybe this function should just return the native adb screenshot in all cases. After all you have `stbt.get_frame` for the camera / hdmi capture.

I'm starting to think that making this look like a `DeviceUnderTest` is a bit misleading if you're actually going to be grabbing frames from `stbt.get_frame`.

drothlis: > Or maybe this function should just return the native adb screenshot in all cases

No, because then what's the point of ADB_720P.

stbt/android.py
drothlis: > Would there be any value in returning a stbt.Frame here

Good idea. Done.

I did a quick test by starting the Android stopwatch when my computer ticked over to 0 seconds of a new minute, and printed the time before & after calling `AdbDevice.screenshot` as well as saving the frame. Seems to be within 0.1s variance. For now I've documented it as "probably to <0.5s or so". To quantify it we could capture frames from HDMI and from ADB at the same time, and do OCR on the time on the screen. But that's outside the scope of this pull request.

stbt/android.py
drothlis: Done.

stbt/android.py
drothlis: I don't like it because `_region_to_tuple("25")` returns `(2, 5)`.

drothlis: I changed the exception to TypeError, and added some tests.

stbt/android.py
drothlis: In `screenshot` I've implemented rotating & scaling for CAMERA_720P, and it raises NotImplementedError for HDMI_720P.

stbt/android.py
drothlis: Not sure. I think of subprocess as an implementation detail and I have considered changing the `__str__` to be less CalledProcessError-like, and possibly changing `cmd` to only show the adb command not the full shell command.

tests/test_android.py
drothlis: It's on the todo list at the top.

stbt/android.py
drothlis: For the record here are my test results of `test_adb_tcpip`:

* connect sleep(1), tcpip/usb sleep(5): Failed 1/7: Device offline
* connect sleep(2), tcpip/usb sleep(2): Failed 1/2: Unable to connect
* connect sleep(2), tcpip/usb sleep(3): Failed 1/6: Unable to connect
* connect sleep(2), tcpip/usb sleep(4): Failed 1/108: Unable to connect
* connect sleep(2), tcpip/usb sleep(5): Failed 0/181.

drothlis:
I think I've addressed all your comments so far except the cv2 & numpy import (because `import stbt` would fail anyway if those aren't installed) and the AdbError parent class.

stbt/android.py
drothlis: I've implemented rotating & scaling for CAMERA_720P, and it raises NotImplementedError for HDMI_720P.

tests/test_android.py
drothlis: Done.

drothlis:
I've squashed all my changes & force-pushed.

drothlis:
I soaked the following stbt test overnight for 10 hours:

```
import os

import cv2

from stbt.android import AdbDevice, CoordinateSystem
from stbt import match, wait_until


def test_get_frame_press_tap_and_swipe():
    adb = AdbDevice(
        adb_server=os.environ.get("ADB_SERVER", "localhost"),
        adb_device=os.environ.get("ANDROID_SERIAL", None),
        adb_binary=os.environ.get("ADB_BINARY", "adb"),
        tcpip=os.environ.get("ADB_TCPIP", False),
        coordinate_system=CoordinateSystem.ADB_NATIVE)

    f = [None]

    def match_any(basename):
        f[0] = adb.get_frame()
        return (match("tests/images/galaxy-ace-2/" + basename, f[0]) or
                match("tests/images/moto-x2/" + basename, f[0]))

    try:
        adb.press("HOME")
        m = wait_until(lambda: match_any("app-icon.png"))
        assert m
        adb.tap(m.region)
        assert wait_until(lambda: match_any("app.png"))
        adb.swipe((240, 0), (240, 600))
        assert wait_until(lambda: match_any("settings-icon.png"))
    except AssertionError:
        cv2.imwrite("adb-screenshot.png", f[0])
        raise
```

With:

### 1. TCPIP

```
ANDROID_SERIAL=192.168.2.162:5555 ADB_TCPIP=True \
./stbt-batch run -kk -o results/adb-tcpip test_android.py::test_get_frame_press_tap_and_swipe
```

There were 29 errors out of 1698 test runs. All were "AdbError: Command '['timeout', '60.000000s', 'adb', '-H', 'localhost', '-s', '192.168.2.162:5555', 'shell', 'screencap', '-p']' failed with exit status 124." (124 is `timeout`'s exit status.) Phone was still reachable & tests were still passing in the morning.

### 2. USB

```
ANDROID_SERIAL=7278681B045C937CEB770FD31542B16 \
./stbt-batch run -kk -o results/adb-usb test_android.py::test_get_frame_press_tap_and_swipe
```

100% passed out of 2618 test runs.

stbt/android.py
```diff
@@ -0,0 +1,784 @@
...
+        """Run any ADB command.
+
+        ``command`` and ``kwargs`` are the same as `subprocess.check_output`,
+        except that ``shell``, ``stdout`` and ``stderr`` are not allowed.
+
+        Raises `AdbError` if the command fails.
+        """
+        try:
+            if self.tcpip:
+                self._connect(timeout_secs)
```
wmanley: So we call `adb connect` before every invocation of `adb`?

stbt/android.py
```diff
@@ -0,0 +1,784 @@
...
+                raise ValueError(
+                    "Invalid value '%s' for android.coordinate_system in "
+                    "config file. Valid values are %s."
+                    % (name, ", ".join("'%s'" % k for k in
+                                       CoordinateSystem.__members__)))  # pylint:disable=no-member
+            coordinate_system = CoordinateSystem[name]
+        self.coordinate_system = coordinate_system
+
+        if self.tcpip:
+            self._connect(timeout_secs=60)
```
wmanley: I'm not a big fan of this constructor.  It feels a little overloaded.  What about:

```
new_adb_device_usb(adb_device=None, adb_server=None, adb_binary=None, coordinate_system=None)
new_adb_device_tcpip(address, adb_server=None, adb_binary=None, coordinate_system=None)
new_adb_device_from_config(config=None)
```

Hmm, seems a bit much.

stbt/android.py
```diff
@@ -0,0 +1,784 @@
...
+            https://developer.android.com/reference/android/view/KeyEvent.html
+            (you can specify it with or without the "KEYCODE_" prefix).
+            Particularly useful key codes are "HOME" and "BACK", which are
+            physical buttons on some phones so you can't hit them with
+            `AdbDevice.tap`.
+        """
+        # "adb shell input keyevent xxx" always returns success, so we need to
+        # validate key names.
+        if key not in _KEYCODES and "KEYCODE_" + key not in _KEYCODES:
+            raise ValueError("Unknown key code %r" % (key,))
```
wmanley: `press("KEY_OK")` will fail. It'll look up `KEYCODE_KEY_OK`.

I would say require the `KEYCODE_` prefix for now.  That way we can sort out a mapping for our standard keycodes in the future without clashes or backwards compatibility concerns.

Is there a reason you need to check the list before forwarding it on to ADB?  Presumably `adb` will fail if the key is missing?

drothlis: Yes, as per the comment in the code: "adb shell input keyevent xxx" always returns success, so we need to validate key names.

I have implemented the mandatory KEYCODE_ prefix.

PR #418, opened 2017-02-23
stbt.android: Implement AdbDevice.get_frame for CoordinateSystem.HDMI_720P

drothlis:
TODO:

- [x] Tests.


drothlis:
I've added tests, fixed some other bugs I noticed, and rebased onto master.

wmanley:
I like the Colombian flag, but it is left-right symmetrical, so these tests wouldn't detect a undesired flipping.  Perhaps the New Zealand flag would have been more appropriate.

PR #420, opened 2017-04-03
Disable sink pipeline and video

wmanley:
For lower resource utilisation when it's not needed.

TODO:

- [x] Tests
- [x] It's not `"disable"` it's `""`


wmanley:
@drothlis: Please review

_stbt/core.py
```diff
@@ -687,8 +687,11 @@ def raise_in_user_thread(exception):
         display[0].tell_user_thread(exception)
     mainloop = _mainloop()
 
-    sink_pipeline = SinkPipeline(
-        gst_sink_pipeline, raise_in_user_thread, save_video)
+    if not gst_sink_pipeline and not save_video:
+        sink_pipeline = NoSinkPipeline()
```
drothlis: Where is the code that checks for the value "disable"?

wmanley: Oh yeah, it's not `"disable"` it's `""`. How silly of me. `""` is better because it's not a valid GStreamer pipeline description.

_stbt/core.py
```diff
@@ -687,8 +687,11 @@ def raise_in_user_thread(exception):
...
+    else:
+        sink_pipeline = SinkPipeline(  # pylint: disable=redefined-variable-type
+            gst_sink_pipeline, raise_in_user_thread, save_video)
```
drothlis: What happens if `gst_sink_pipeline` is `None` or `""` or `"disable"` but `save_video` is truthy?

Should we fall back to fakesink + a pipeline that encodes the video? Should we error?

Maybe instead of "disable" we can grep the sink_pipeline for "fakesink" -- something like `if "fakesink" in gst_sink_pipeline and not save_video`. Maybe that's too magical though. It would break if someone put in a tee with a fakesink on one branch and something useful on the other, but it seems unlikely that anyone would do that.

_stbt/core.py
```diff
@@ -1668,6 +1671,25 @@ def draw(self, obj, duration_secs=None, label=""):
                     "Can't draw object of type '%s'" % type(obj).__name__)
 
 
+class NoSinkPipeline(object):
+    """
+    Used in place of a SinkPipeline when no video output is required.  Is a lot
+    faster because it doesn't do anything.  It especially doesn't do any copying
+    nor video encoding :).
```
drothlis: We don't do video encoding either if `--save-video` isn't set. Do you have a rough indication of what the performance gain is? What copying are we doing without this -- I assume the only copy is in `gst_sample_make_writable` called from `push_sample`.

_stbt/core.py
```diff
@@ -681,11 +681,20 @@ def new_device_under_test_from_config(
     if transformation_pipeline is None:
         gst_sink_pipeline = get_config('global', 'transformation_pipeline')
 
-    display = Display(
-        gst_source_pipeline, gst_sink_pipeline, _mainloop(), save_video,
-        restart_source, transformation_pipeline)
+    display = [None]
+
+    def raise_in_user_thread(exception):
+        display[0].tell_user_thread(exception)
```
drothlis: You don't need to make `display` be an array because the only place we access it from a different context (here) is for calling one of its methods, not for *setting* `display[0]`.

drothlis:
I'm a bit confused about what the `Display` class is nowadays. I suppose it's essentially a `SourcePipeline` except that it also has a `sink_pipeline` member because it needs to push samples through to the sink pipeline.

I suppose when we initially called it `Display` we were thinking it represents the display of the *device under test*, not the output/monitoring display (that's the sink pipeline).

This whole comment is just an aside. If we ever get around to making `DeviceUnderTest` a public API then we can worry about a better name for `Display`.

drothlis:
Rebased onto master (fixing conflicts in `docs/release-notes.md` and `stbt-batch.d/run-one`), fixed incorrect `sink_pipeline="disable"` wording in commit message, and added a minor code cleanup.

_stbt/core.py
wmanley: Fakesink and pipeline sounds like the right behaviour

drothlis:
I'm not sure if the commit *sink-pipeline: Disable sink pipeline if `sink_pipeline=""`* is strictly necessary -- I imagine that most of the performance improvement comes from `--no-save-video`.

Furthermore it might make more sense to implement it as a list of pipelines so you could have a separate "save video" pipeline, a "display the video with xvimagesink" pipeline, and possibly even non-GStreamer sinks like "log draw_text annotations to a file for use with DASH".

However I don't intend to do that work at the moment, so we should probably merge this as is to enable us to progress #449.

@wmanley please review.

_stbt/core.py
```diff
@@ -732,7 +732,8 @@ def new_device_under_test_from_config(
     if restart_source is None:
         restart_source = get_config('global', 'restart_source')
     if transformation_pipeline is None:
-        gst_sink_pipeline = get_config('global', 'transformation_pipeline')
+        transformation_pipeline = get_config('global',
+                                             'transformation_pipeline')
```
wmanley: Whoops!

PR #421, opened 2017-04-04
power: fix ATEN oid for ports above 8

lmaneos:
The ATEN MIB is using 1.3.6.1.4.1.21317.1.3.2.2.2.2.10 for
`outletConfigTable`, so on models with more than 8 ports `outlet9Status`
and up (at least up to 36 currently) are offset by 2.


drothlis:
I'll take your word for it. :-)  Thanks!

PR #425, opened 2017-05-29
virtual-stb selftests: Use KeyboardEvent.key instead of .keyIdentifier

drothlis:
`keyIdentifier` is non-standard, deprecated, and it has been removed
from chromium, which is causing our `stbt virtual-stb` selftests to
fail. See
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyIdentifier


wmanley:
Nice one @drothlis

PR #428, opened 2017-06-16
Automatically build & publish Ubuntu packages to ppa:stb-tester/stb-tester-nightly

drothlis:
**TODO:**

- [x] Create PPA https://launchpad.net/~stb-tester/+archive/ubuntu/stb-tester-nightly
- [x] Test publishing manually: From c3e69db I did `extra/debian/ubuntu-shell.sh -c "make deb" && DPUT_HOST=ppa:stb-tester/stb-tester-nightly make ppa-publish`
- [ ] Configure Travis to build & publish package from master branch.
    - [ ] `ubuntu-shell.sh` uses docker. Can we run this on Travis?
    - [ ] How to authenticate to Launchpad: Create a new user? Add a new GPG key to my user?
    - [ ] Encrypt credentials (GPG password) using [Travis encryption keys].
- [ ] Document in our installation instructions.
- [ ] Tell the mailing list.

[Travis encryption keys]: https://docs.travis-ci.com/user/encryption-keys/


drothlis:
I'm having trouble remembering / finding out how Launchpad authentication works. My current theory is that upload itself is anonymous, but Launchpad checks that the package you upload was signed by a GPG associated with an authorized user.

drothlis:
I couldn't think of a better name for the PPA than "nightly". It's a fairly standard term, although in this case the builds would be triggered by merges to master rather than running on a nightly schedule. However "dev" isn't as evocative as "nightly", and "unstable" does us a disservice because we have fairly extensive automated testing that has to pass before any build would be published.

PR #430, opened 2017-06-20
stbt-batch: Performance improvements: Record screenshots in stbt-run

wmanley:
stbt batch run wants a JPG thumbnail of the system under test at the end
of a test run.  Previously it would spawn an additional python instance
after the test run, load the PNG screenshot and then save it as a jpeg.
If the test run succeeded it would even spawn yet another stbt run just
to create a screenshot.png before deleting it after converting it to JPG.

This commit tidies this up.  stbt-run is responsible for creating all
screenshots and thumbnails as it has access to the uncompressed source
screenshot to start with.  This can save 1s for each test run by
`stbt batch`.

### DIFF
--- a/stbt-run
+++ b/stbt-run
@@ -25,6 +25,14 @@ parser.add_argument(
     '--save-trace', metavar='FILE', default=None,
     help='Write state to this file, as xz compressed newline-seperated JSON')
 parser.add_argument(
+    '--save-screenshot', default='on-failure',
+    choices=['always', 'on-failure', 'never'],
+    help="Save a screenshot at the end of the test to screenshot.png")
+parser.add_argument(
+    '--save-thumbnail', default='never',
+    choices=['always', 'on-failure', 'never'],
+    help="Save a thumbnail at the end of the test to thumbnail.jpg")
+parser.add_argument(
     'script', metavar='FILE[::TESTCASE]', help=(
         "The python test script to run. Optionally specify a python function "
         "name to run that function; otherwise only the script's top-level will "
@@ -70,6 +78,35 @@ def _print_exc_utf8(file_=None):
         traceback.print_exc(file=file_)
     finally:
         traceback._some_str = _old_some_str
+
+
+def _save_screenshot(exception):
+    import cv2
+
+    jpg = (args.save_thumbnail == 'always' or
+           (args.save_thumbnail == 'on-failure' and exception))
+    png = (args.save_screenshot == 'always' or
+           (args.save_screenshot == 'on-failure' and exception))
+
+    if not jpg and not png:
+        return
+
+    screenshot = getattr(exception, "screenshot", None)
+    if screenshot is None and stbt._dut._display:  # pylint: disable=protected-access
+        screenshot = stbt._dut._display.last_used_frame  # pylint: disable=protected-access
+    if screenshot is None:
+        screenshot = stbt._dut.get_frame()  # pylint: disable=protected-access
+
+    if png:
+        cv2.imwrite("screenshot.png", screenshot)
+        sys.stderr.write("Saved screenshot to 'screenshot.png'.\n")
+
+    if jpg:
+        cv2.imwrite(
+            'thumbnail.jpg',
+            cv2.resize(screenshot, (
+                640, 640 * screenshot.shape[0] // screenshot.shape[1])),
+            [cv2.cv.CV_IMWRITE_JPEG_QUALITY, 50])
 
 
 def import_by_filename(filename_):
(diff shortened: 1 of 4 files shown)


stbt-run
```diff
@@ -72,6 +80,35 @@ def _print_exc_utf8(file_=None):
         traceback._some_str = _old_some_str
 
 
+def _record_screenshot(exception):
+    import cv2
+
+    jpg = (args.write_thumbnail == 'always' or
+           (args.write_thumbnail == 'on-failure' and exception))
+    png = (args.write_screenshot == 'always' or
+           (args.write_screenshot == 'on-failure' and exception))
```
wmanley: May as well put an early exit in here.

wmanley:
Should add some tests for this stuff.  We've already got `test_that_stbt_run_saves_screenshot_on_match_timeout`, `test_that_stbt_run_saves_screenshot_on_precondition_error` and `test_that_stbt_run_saves_last_grabbed_screenshot_on_error`.

stbt-run
```diff
@@ -25,6 +25,14 @@ parser.add_argument(
     '--save-trace', metavar='FILE', default=None,
     help='Write state to this file, as xz compressed newline-seperated JSON')
 parser.add_argument(
+    '--write-screenshot', default='on-failure',
+    choices=['always', 'on-failure', 'never'],
+    help="Record a screenshot at the end of the test to screenshot.png")
+parser.add_argument(
+    '--write-thumbnail', default='never',
+    choices=['always', 'on-failure', 'never'],
+    help="Record a thumbnail at the end of the test to thumbnail.jpg")
```
drothlis: Did you choose "write" instead of "save" (like the existing "save-trace" and "save-video" on purpose?

stbt-run
```diff
@@ -72,6 +80,35 @@ def _print_exc_utf8(file_=None):
         traceback._some_str = _old_some_str
 
 
+def _record_screenshot(exception):
```
drothlis: Might as well call this `_write_screenshot` (or `_save_screenshot`) to be consistent with the wording of the command-line argument.

stbt-run
```diff
@@ -72,6 +80,35 @@ def _print_exc_utf8(file_=None):
...
+           (args.write_screenshot == 'on-failure' and exception))
+
+    screenshot = getattr(exception, "screenshot", None)
+    if screenshot is None and stbt._dut._display:  # pylint: disable=protected-access
+        screenshot = stbt._dut._display.last_used_frame  # pylint: disable=protected-access
+    if screenshot is None:
+        screenshot = stbt._dut.get_frame()  # pylint: disable=protected-access
+
+    if screenshot is None:
+        return
```
drothlis: Can `screenshot` still be `None` after we've done `stbt._dut.get_frame()`?

drothlis:
> Should add some tests for this stuff.

Meh. These tests are relatively slow (they have to start up a GStreamer pipeline each time) and all combinations of `--save-screenshot={never,on-failure,always}`, `--save-thumbnail={never,on-failure,always}`, and `test_passes={yes,no}` are 18 tests. The default values for `stbt run` are already tested in the tests you mentioned above (I added checks that the thumbnail isn't written in those cases), and the values we use with `stbt batch` are tested in `test_that_stbt_batch_run_runs_until_failure`:

```
    [[ -f $first/thumbnail.jpg ]] || fail "Expected successful testrun to create thumbnail"
    [[ ! -f $first/screenshot.png ]] || fail "Expected successful testrun to not create screenshot"
    [[ -f latest/thumbnail.jpg ]] || fail "Expected failed testrun to create thumbnail"
    [[ -f latest/screenshot.png ]] || fail "Expected failed testrun to create thumbnail"
```

stbt-run
```diff
@@ -72,6 +80,35 @@ def _print_exc_utf8(file_=None):
...
+    jpg = (args.write_thumbnail == 'always' or
+           (args.write_thumbnail == 'on-failure' and exception))
+    png = (args.write_screenshot == 'always' or
+           (args.write_screenshot == 'on-failure' and exception))
+
+    screenshot = getattr(exception, "screenshot", None)
+    if screenshot is None and stbt._dut._display:  # pylint: disable=protected-access
+        screenshot = stbt._dut._display.last_used_frame  # pylint: disable=protected-access
+    if screenshot is None:
+        screenshot = stbt._dut.get_frame()  # pylint: disable=protected-access
```
drothlis: This last clause (if `stbt._dut._display.last_used_frame` is None then grab a new frame by calling `get_frame()`) causes a regression in `test_invalid_source_pipeline`: Instead of the error message about the invalid source pipeline, you get a stack trace from `get_frame()`.

drothlis:
@wmanley I think this is ready to merge.

PR #432, opened 2017-06-23
WIP: Add `remove_transparency` pre-processing function

wmanley:
Often stb UIs consist of a transparent overlay over live TV or some other
changable image.  The transparency can cause OCR and match failures due to
the background bleeding through.  This function replaces these transparent
background with solid colour which can be used as a pre-processing step
before performing a `match` or `ocr`.

color is an HTML-style hex colour string of the form #rrggbbaa.

It assumes that an overlay image was alpha-blended according to:

    frame = color[rgb] * color[alpha] + underlay * (1 - color[alpha])

where color specifies overlay and alpha.

See https://en.wikipedia.org/wiki/Alpha_compositing#Alpha_blending

Usage:

    remove_transparency('#bf000079', frame=...)

TODO:

- [ ] Tests


drothlis:
Should it be called "remove **translucency**" instead of "transparency"?

Should we create a new `stbt.preprocessing` module or just keep jamming stuff into `stbt`? This doesn't have any dependencies except numpy (not even other stbt functions).

PR #433, opened 2017-06-26
stbt.ocr: Passing region=None means no region, not Region.ALL

drothlis:
This was added for backwards compatibility in v0.21 (two and a half
years ago) but since then it has printed a warning that this change
would be made eventually.

This is consistent with `stbt.match_text`. In fact my motivation for
making this change is that the difference in behaviour caused me some
confusion when I was writing a test for the recent `text_color` pull
request. It's also consistent with the output of `Region.intersect`
which means that it's safe to pass regions created from
`Region.intersect` into `stbt.ocr`.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1037,14 +1037,6 @@ class DeviceUnderTest(object):
         if frame is None:
             frame = self._display.pull_frame()
 
-        if region is None:
-            warnings.warn(
-                "Passing region=None to ocr is deprecated since 0.21 and the "
-                "meaning will change in a future version.  To OCR an entire "
-                "video frame pass region=Region.ALL instead",
-                DeprecationWarning, stacklevel=2)
-            region = Region.ALL
-
         text, region = _tesseract(
             frame, region, mode, lang, tesseract_config,
             tesseract_user_patterns, tesseract_user_words, text_color)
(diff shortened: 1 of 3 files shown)


_stbt/core.py
```diff
@@ -1037,14 +1037,6 @@ def ocr(self, frame=None, region=Region.ALL,
         if frame is None:
             frame = self._display.pull_frame()
 
-        if region is None:
-            warnings.warn(
-                "Passing region=None to ocr is deprecated since 0.21 and the "
-                "meaning will change in a future version.  To OCR an entire "
-                "video frame pass region=Region.ALL instead",
-                DeprecationWarning, stacklevel=2)
-            region = Region.ALL
```
wmanley: I wonder if we should have gone through a transition where this `raise`d an exception.  There may be people out there who have just not been seeing this warning.

_stbt/core.py
```diff
@@ -2822,12 +2814,13 @@ def draw(self, *_args, **_kwargs):
...
+        # pylint: disable=line-too-long
+        ("Connection-status--white-on-dark-blue.png", "Connection status: Connected", Region.ALL, None),
         ("Connection-status--white-on-dark-blue.png", "Connected", Region(x=210, y=0, width=120, height=40), None),
-        ("programme--white-on-black.png", "programme", None, None),
-        ("UJJM--white-text-on-grey-boxes.png", "", None, None),
-        ("UJJM--white-text-on-grey-boxes.png", "UJJM", None, OcrMode.SINGLE_LINE),
+        ("Connection-status--white-on-dark-blue.png", "", None, None),
+        ("programme--white-on-black.png", "programme", Region.ALL, None),
+        ("UJJM--white-text-on-grey-boxes.png", "", Region.ALL, None),
+        ("UJJM--white-text-on-grey-boxes.png", "UJJM", Region.ALL, OcrMode.SINGLE_LINE),
```
wmanley: Hah, including in our own selftest-suite!

_stbt/core.py
drothlis: Good idea. I've pushed a new commit that makes this a TypeError.

PR #436, opened 2017-06-30
stbt.wait_until: Return the first stable value, not the last

drothlis:
If `stable_secs` is specified, wait_until should return the first stable
value. This is similar to how `wait_for_motion` reports the timestamp of
the first frame where motion was detected, even when you've asked to
validate continuous motion for several frames.

This allows uses like waiting for a selection to finish moving, and then
report how long the animation took.

**TODO:**

- [x] Tests.
- [x] Docs

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -507,8 +507,8 @@ class MatchResult(object):
         return self.match
 
     def __eq__(self, other):
-        return (self.match == other.match and self.region == other.region and
-                self.image == other.image)
+        return (isinstance(other, MatchResult) and self.match == other.match and
+                self.region == other.region and self.image == other.image)
 
     def __ne__(self, other):
         return not self.__eq__(other)
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1189,7 +1189,10 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
 
         If ``stable_secs`` is specified then ``wait_until`` returns:
 
-        * ``callable_``'s return value if it was truthy and stable.
+        * ``callable_``'s return value if it was truthy and stable. This will
+          be the *first* stable value, so that you can use ``wait_until`` for
+          performance measurements (for example to measure the time for an
+          animation to complete).
         * ``None`` if ``callable_``'s return value was truthy but not stable
           before we reached ``timeout_secs``.
         * ``callable_``'s last return value if it was falsey when we reached
@@ -1241,11 +1244,7 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
     """
     import time
 
-    class NoValue(object):
-        pass
-
-    stable_value = NoValue
-    stable_since = None
+    stable_value = None
     expiry_time = time.time() + timeout_secs
 
     while True:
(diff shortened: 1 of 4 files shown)


drothlis:
Hmmm this makes pylint think that `wait_until` returns `NoValue`, even though that's impossible. We can tell pylint to ignore `NoValue`, but each of our users would have to do the same in their own pylint config.

drothlis:
It's safe to use `None` for the sentinel value (instead of `NoValue`) because stability isn't considered for falsey values.

wmanley:
So did using `None` as the sentinel help pylint not get confused?

tests/test_core.py
```diff
@@ -106,3 +107,34 @@ def test_wait_until(mock_time, f, kwargs, expected):
 def test_that_wait_until_times_out(mock_time):
     assert not wait_until(Zero, interval_secs=1)
     assert time.time() == 1497000010
+
+
+def test_that_wait_until_returns_first_stable_value(mock_time):
+
+    def MR(match, x):
+        time.sleep(1)  # advance the mock time by 1 second
```
wmanley: Where is time mocked out?

wmanley: I see, it's a pytest fixture

_stbt/core.py
```diff
@@ -1241,11 +1244,8 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
     """
     import time
 
-    class NoValue(object):
-        pass
-
-    stable_value = NoValue
-    stable_since = None
+    stable_value = None
+    stable_since = time.time()
```
wmanley: I think this `time.time()` is misleading.  According to my reading of the function the value will never be used.  You could leave it as `None` and change the code below (see comment below)

_stbt/core.py
```diff
@@ -1259,7 +1259,7 @@ class NoValue(object):
         stable_duration = t - stable_since
 
         if value and stable_duration >= stable_secs:
```
wmanley: This could be `if value and t - stable_since >= stable_secs`.  That way `stable_since` can be `None` (or even not set) until `value`.

drothlis:
> So did using `None` as the sentinel help pylint not get confused?

Yes.

_stbt/core.py
drothlis: Done.

drothlis:
Hmmm all of a sudden the virtual-stb tests are failing on Travis. Probably an unrelated change in the Travis infrastructure. The failing tests use `wait_for_match`, not `wait_until`.

drothlis:
Yep, Travis is failing on master too. https://travis-ci.org/stb-tester/stb-tester/builds/249930696

wmanley:
Travis has a notice:

> This job ran on our Trusty, `sudo: required` environment which was updated on Wednesday, June 21st. Read all about it on our [blog](https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch) and take note that you can add `group: deprecated-2017Q2` in your .travis.yml file to use the previous version.

Not sure if it's relevant

wmanley:
Chromium has been updated from 53 to 58.  Could be the cause?

drothlis:
> Travis has a notice:

I'd seen that notice before, when the tests were still passing.

wmanley:
Yep, still fails with the workaround.

drothlis:
The virtual-stb tests pass on my laptop. I'm disabling them on Travis.

.travis.yml
```diff
@@ -8,7 +8,6 @@ install:
  - export DEBIAN_FRONTEND=noninteractive
  - sudo apt-get update
  - sudo apt-get -y install
-        chromium-browser
```
drothlis: Heh, chromium-browser is already installed on Travis so it runs the virtual-stb tests anyway.

drothlis:
> The virtual-stb tests pass on my laptop

Nevermind, after upgrading chromium-browser the tests fail on my laptop too. I guess the update just came out today.

drothlis:
I've raised #440 for the virtual-stb tests.

PR #440, opened 2017-07-04
Fix virtual-stb with Chrome 59

drothlis:
They've started failing since the latest Chromium release (59.0.3071.109). I disabled the tests temporarily in commit 1e490e5. 

Using `stbt tv` to debug the running virtual-stb shows a blank white page. Loading the same page in chromium-browser directly works fine.

```
./stbt_virtual_stb.py run tests/vstb-example-html5/run.sh &
./stbt-tv
```

vs.

```
chromium-browser --no-sandbox --app=file://`pwd`/tests/vstb-example-html5/virtual-stb/index.html
```

### DIFF
--- a/tests/test-virtual-stb.sh
+++ b/tests/test-virtual-stb.sh
@@ -13,8 +13,6 @@ with_html5_vstb()
 
 test_that_virtual_stb_configures_stb_tester_for_testing_virtual_stbs()
 {
-    skip "virtual-stb tests don't work with chromium-browser 59.0.3071.109"
-
     with_html5_vstb --x-keymap=$testdir/vstb-example-html5/key-mapping.conf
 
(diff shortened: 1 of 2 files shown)


drothlis:
@wmanley please review. The tests now pass because I've added `--disable-gpu`, but would you consider this a bug in virtual-stb?

drothlis:
Maybe it isn't fixed after all. It passed on 1 travis job and failed on the other. On my laptop it failed 1/5 times, with the same symptom (screenshot is a plain white screen).

drothlis:
Seems to be working again, perhaps they fixed it in Chromium 60. (On my laptop: Failed 1/5 times with Chromium 59, 0/22 times with Chromium 60.)

PR #441, opened 2017-07-13
wait_until: Add `ignored_exceptions` parameter

drothlis:
Useful if you've started an HTTP server in a background thread and
you're waiting for it to start up by trying to connect to it; or you've
just sent a key-press to the device-under-test and the FrameObject
you're waiting for doesn't handle unknown states very well; etc.


drothlis:
N.B. The name is consistent with Selenium's [WebDriverWait] which also has an `ignored_exceptions` parameter.

[WebDriverWait]: https://selenium-python.readthedocs.io/api.html#selenium.webdriver.support.wait.WebDriverWait

drothlis:
Rebased on top of #444.

_stbt/core.py
```diff
@@ -1213,43 +1217,67 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
...
+    stable_predicate_value = None
     expiry_time = time.time() + timeout_secs
 
     while True:
         t = time.time()
-        value = callable_()
+        try:
+            value = callable_()
+        except ignored_exceptions as e:
+            debug("wait_until: Ignoring exception %r" % e)
```
drothlis: Need to test unicode behaviour here? We've had trouble with printing exceptions in the past.

_stbt/core.py
```diff
@@ -1213,43 +1217,67 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
...
+            value = None
+            predicate_value = None
+        else:
+            try:
+                predicate_value = predicate(value)
+            except ignored_exceptions as e:
+                debug("wait_until: Ignoring exception %r" % e)
```
drothlis: And here.

drothlis:
Rebased onto master after merging #444.

_stbt/core.py
```diff
@@ -1244,8 +1249,18 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, predicate=None,
 
     while True:
         t = time.time()
-        value = callable_()
-        predicate_value = predicate(value)
+        try:
+            value = callable_()
+        except ignored_exceptions as e:
```
wmanley: This works?!?  Fancy that, I guess it makes sense.

_stbt/core.py
```diff
@@ -1244,8 +1249,18 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, predicate=None,
 
     while True:
         t = time.time()
-        value = callable_()
-        predicate_value = predicate(value)
+        try:
+            value = callable_()
```
wmanley: Could this be

```
try:
    value = callable_()
    predicate_value = predicate(value)
except ...
```

and then you could avoid the `else:` block and the second `try...except`?

Unless this has something to do with not setting `value = None` if predicate fails?

wmanley:
i'd like a little bit more discussion of use-cases here.  Can you provide an example of where this would have been useful?

drothlis:
We'll need to test that it raises (or returns None?) if the callable continues to raise exceptions after the timeout.

_stbt/core.py
drothlis: This comment was before we dropped support for Python 2.

_stbt/core.py
drothlis: This comment was before we dropped support for Python 2.

PR #442, opened 2017-07-13
Add stbt.load_image

drothlis:
It looks up the given path in the caller's directory, then the caller's caller, etc.  Also raises instead of returning None -- having to check the return value of `cv2.imread` is super annoying.


_stbt/core.py
```diff
@@ -2491,24 +2527,32 @@ def _iter_frames(depth=1):
...
-    # stack()[2] is detect_match's caller (the user script).
+    # Start searching from the first parent stack-frame that is outside of
+    # the _stbt installation directory (this file's directory). We can ignore
+    # the first 2 stack-frames:
+    #
+    # * stack()[0] is _find_path;
+    # * stack()[1] is _find_path's caller: load_image or _load_template;
+    # * stack()[2] is load_image's caller (the user script). It could also be
+    #   _load_template's caller (e.g. `match`) so we still need to check until
+    #   we're outside of the _stbt directory.
```
wmanley: I'm not a huge fan of the way we do image lookup.  It makes caching difficult and the result of calling match in a function non-deterministic.

Some alternatives:

* Use a similar trick to pytest which has a [`__tracebackhide__`](https://docs.pytest.org/en/latest/example/simple.html#writing-well-integrated-assertion-helpers) attribute for helpers.  We would then consider only the first stack frame that isn't marked with `__tracebackhide__`.
* Expect to find all images under some prefix in the test-pack - this would bind our implementation to the concept of a test-pack though.

Oh well, it's too late now, we have to preserve the algorithm as it is to avoid breaking test-scripts.

drothlis:
Rebased onto master to fix conflict with #444 (in release-notes.md and test_core.py).

_stbt/core.py
```diff
@@ -2499,43 +2502,43 @@ def _iter_frames(depth=1):
...
-    debug("Using mask %s" % mask_path)
-    if not mask_path or not os.path.isfile(mask_path):
-        raise UITestError("No such mask file: %s" % mask)
-    mask_image = cv2.imread(mask_path, cv2.CV_LOAD_IMAGE_GRAYSCALE)
-    if mask_image is None:
-        raise UITestError("Failed to load mask file: %s" % mask_path)
-    return mask_image
+    absolute_filename = _find_path(filename)
+    debug("Using mask %s" % absolute_filename)
+    if not absolute_filename or not os.path.isfile(absolute_filename):
```
wmanley: it's guaranteed to be a file isn't it?

_stbt/core.py
```diff
@@ -546,6 +546,42 @@ def _load_template(template):
...
+
+    :returns: An image in OpenCV format (a `numpy.ndarray` of 8-bit values, 3
+        channel BGR).
+    :raises: `ValueError` if the specified path doesn't exist or isn't a valid
+        image file.
+    """
+
+    absolute_filename = _find_path(filename)
+    if not absolute_filename or not os.path.isfile(absolute_filename):
+        raise ValueError("No such file: %s" % filename)
```
wmanley: Perhaps this should be `IOError` - this is what is usually raised for file not found errors. Arguably `_find_path` should actually be `_open_file_on_path`.  It would avoid races (probably not a problem in practice) and centralise more error handling. You could then use `cv2.imdecode` to get the image out and `file.name` to get the name of the file.

_stbt/core.py
```diff
@@ -546,6 +546,42 @@ def _load_template(template):
...
+    image = cv2.imread(absolute_filename, flags)
+    if image is None:
+        raise ValueError("Failed to load image: %s" % absolute_filename)
```
wmanley: `ValueError` doesn't quite seem right here either.  There's something wrong with the contents of a file - this doesn't mean that the value of the filename is incorrect.  Perhaps `RuntimeError` would be more appropriate.

_stbt/core.py
wmanley: Ah, but `imdecode` takes a numpy array, not a file object

_stbt/core.py
wmanley: I see you've fixed this later in this PR

wmanley:
Review complete.  I think the exceptions should probably be changed.  I don't think there's much we can do about the rest of my concerns without breaking backwards compatibility.

_stbt/core.py
drothlis: As a data point, `PIL.Image.open` from the Python Imaging Library raises `IOError` for an invalid PNG.

`cv2.imread` just silently returns None, of course.

drothlis:
Thanks for the review. I've fixed the exceptions & merged.

_stbt/core.py
```diff
@@ -533,14 +535,51 @@ def _load_template(template):
...
         return _AnnotatedTemplate(template, None, None)
     else:
-        template_name = _find_path(template)
-        if not template_name or not os.path.isfile(template_name):
-            raise UITestError("No such template file: %s" % template)
-        image = cv2.imread(template_name, cv2.CV_LOAD_IMAGE_COLOR)
+        relative_filename = template
+        absolute_filename = _find_user_file(relative_filename)
+        if not absolute_filename:
+            raise IOError("No such template file: %s" % relative_filename)
```
wmanley: This should be `IOError(errno.ENOENT, "No such template file: %s" % relative_filename, relative_filename)`

[Python docs say](https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError):

> When exceptions of this type are created [...], the first item is available on the instance’s errno attribute (it is assumed to be an error number), and the second item is available on the strerror attribute \[...\]  the third item is available on the filename attribute.

drothlis: Huh. I'm surprised because the tests passed. It seems that it falls back to the normal error message formatting if you only specify one argument:

```
>>> str(IOError("cannot identify image file 'dave.png'"))
"cannot identify image file 'dave.png'"

>>> str(IOError(errno.ENOENT, "cannot identify image file 'dave.png'", 'dave.png'))
"[Errno 2] cannot identify image file 'dave.png': 'dave.png'"
```

drothlis: We should use ENOENT for "file not found". What about "invalid png file"?

* EINVAL - Invalid argument
* EIO - Input/output error
* EPROTO - Protocol error

`PIL.Image.open` raises an IOError of a single string argument, so that's no help.

drothlis: > It seems that it falls back to the normal error message formatting if you only specify one argument

And that is documented under [EnvironmentError] (IOError's base class):

> The filename attribute is None when this exception is created with other than 3 arguments. The errno and strerror attributes are also None when the instance was created with other than 2 or 3 arguments. In this last case, args contains the verbatim constructor arguments as a tuple.

[EnvironmentError]: https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError

drothlis: Raised #447 but see the caveats in the comments there.

PR #443, opened 2017-07-14
Fix stbt.MatchResult and wait_until's stable_secs

drothlis:
TODO:

- [x] Remove MatchResult `__eq__` and `__hash__`.
- [x] Don't use value comparison in `wait_until` if you aren't using `stable_secs`.


wmanley:
Yes, I think this is a nice safe correctness improvement. I'm a lot more comfortable with this implementation

PR #444, opened 2017-07-14
stbt.wait_until: Add `predicate` parameter

drothlis:
TODO:

- [x] Merge #443.
- [x] Tests.
- [x] Docs.
- [x] Release notes.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1237,10 +1237,10 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
 
         if t >= expiry_time:
             debug("wait_until timed out: %s" % _callable_description(callable_))
-            if stable_secs:
-                return None
-            else:
+            if not value:
                 return value  # it's falsey
+            else:
+                return None  # must have failed stable_secs check
 
         time.sleep(interval_secs)
 
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1213,15 +1213,6 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
         if not wait_until(lambda: match("xyz.png")):
             do_something_else()
 
-    There are some drawbacks to using `assert` instead of `wait_for_match`:
-
-    * The exception message won't contain the reason why the match failed
-      (unless you specify it as a second parameter to `assert`, which is
-      tedious and we don't expect you to do it), and
-    * The exception won't have the offending video-frame attached.
-
-    We hope to solve both of the above drawbacks at some point in the future.
-
     ``wait_until`` was added in stb-tester v22. The ``stable_secs`` parameter
     was added in v28.
     """
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1143,7 +1143,8 @@ def save_frame(image, filename):
     cv2.imwrite(filename, image)
 
 
-def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
+def wait_until(callable_, timeout_secs=10, interval_secs=0, predicate=None,
+               stable_secs=0):
     """Wait until a condition becomes true, or until a timeout.
 
     Calls ``callable_`` repeatedly (with a delay of ``interval_secs`` seconds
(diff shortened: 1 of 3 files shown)


_stbt/core.py
```diff
@@ -1143,7 +1143,8 @@ def save_frame(image, filename):
     cv2.imwrite(filename, image)
 
 
-def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
+def wait_until(callable_, timeout_secs=10, interval_secs=0, predicate=None,
+               stable_secs=0):
```
wmanley: `predicate` should probably be the last parameter.  I like this for two reasons:

* Backwards compatibility - in case anyone's using `wait_until` without keyword arguments.
* Having callables last makes code easier to read because they can often involve brackets and it can get confusing where the parameter ends

_stbt/core.py
```diff
@@ -1162,24 +1163,22 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
...
-          performance measurements (for example to measure the time for an
-          animation to complete).
-        * ``None`` if ``callable_``'s return value was truthy but not stable
-          before we reached ``timeout_secs``.
-        * ``callable_``'s last return value if it was falsey when we reached
-          ``timeout_secs``.
+        succeeded, or falsey if ``wait_until`` timed out). If the value was
+        truthy when the timeout was reached but it failed the ``predicate`` or
+        ``stable_secs`` conditions (if any) then ``wait_until`` returns
+        ``None``.
```
wmanley: So the motivation for returning `None` is for use cases like this:

```
assert wait_until(Guide, predicate=lambda x: x.selection)
```

but it would prevent use cases like:

```
mr = wait_until(is_motion, predicate=lambda x: not x)
assert not mr, "Motion found in region %r" % mr.region
```

Seems reasonable.  I guess that our `stable_secs` policy of returning the first truthy result is fundamentally incompatible with returning a truthy value when the predicate fails.  I suppose that predicate just cannot be used when waiting for `not` something because of the difficulties distinguishing whether a timeout happened or not.

tests/test_core.py
```diff
@@ -134,11 +157,13 @@ def g():
...
 
     def match():
-        match_result = next(results)
-        return match_result and match_result.region
+        return next(results)
 
-    result = wait_until(match, stable_secs=2)
-    assert result == stbt.Region(x=4, y=0, width=10, height=2)
+    result = wait_until(match, predicate=lambda x: x and x.region,
+                        stable_secs=2)
```
wmanley: You could ditch `match` and write:

        result = wait_until(g().next, predicate=lambda x: x and x.region,
                             stable_secs=2)

but that's probably too confusing.

_stbt/core.py
```diff
@@ -1213,34 +1212,57 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, stable_secs=0):
...
+        end_time = match_result.time  # this is the first stable frame
+        print "Transition took %s seconds" % (end_time - start_time)
+
+    ``wait_until`` was added in stb-tester v22. The ``predicate`` and
+    ``stable_secs`` parameters were added in v28.
     """
     import time
 
+    if predicate is None:
+        predicate = lambda x: x
```
wmanley: Yes, this is a nice simple implementation.

wmanley:
Right, so the motivation is that before you'd have to write:

```
def f():
    m = stbt.match("template.png")
    return m and m.region
assert wait_until(f, stable_secs=4)
```

and now it's:

```
assert wait_until(lambda: stbt.match('template.png'),
                      predicate=lambda x: x and x.region,
                      stable_secs=4)
```

wmanley:
This looks ok to me.  I'm having difficulty visualising all the use-cases so it's difficult for me to comment whether the trade-offs are right.  It doesn't interfere with the existing use-cases of `wait_until` so it's difficult to see the harm.

The only risks I can see is that there could be a future violation of the "There should be one-- and preferably only one --obvious way to do it" rule.

+1

drothlis:
Thanks for the review. Yes, I'm much more comfortable with this than the previous `stable_secs` + `MatchResult.__ne__` implementation.

drothlis:
Oops forgot to fix the order of the parameters. I'll do that in another pull request.

_stbt/core.py
drothlis: `stable_secs` hasn't been released yet, so there's no backward compatibility issue there. I'll change it anyway so that it comes last, though it seems unlikely that anyone would pass all of these as positional parameters.

drothlis: Hmm now I remember why I put it in that order -- it makes more sense to document `predicate` first, since `predicate` can be documented without mentioning `stable_secs`, but `stable_secs` needs to talk about `predicate`. I think I'll leave it as is.

Issue #246, opened 2015-01-06
get_frame provides an old frame after a sleep while video wasn't rolling


szemate:
`stbt._display.last_sample` stores the most recent GSt buffer until `stbt.gst_samples` removes it or it's replaced by a newer buffer. To put it other way, if the video stream stops then `gst_samples`
yields the very last sample of the video stream even though it's called minutes later.

This behaviour has unwanted effects in the following situations:

1.

``` py
stbt.press("POWER_OFF")
time.sleep(5)
# Make sure that the device under test is off
try:
    stbt.get_frame()
except stbt.NoVideo:
    pass
else:
    raise stbt.UITestFailure("The device is still on")
```

In this case `stbt.get_frame` never raises `NoVideo`, as it returns the
sample that had been stored _before_ the box was powered off.

2.

``` py
stbt.wait_for_match("menu.png")
stbt.press("POWER_OFF")
time.sleep(5)
stbt.press("POWER_ON")
# Make sure that we no longer see menu.png because
# it should disappear if the device under test is restarted
try:
    stbt.wait_for_match("menu.png")
except stbt.MatchTimeout:
    pass
else:
    raise stbt.UITestFailure("menu.png hasn't disappeared")
```

In this case we have a race condition between stbt and the device under
test. The device needs some time to start delivering frames after it's
powered on; if it doesn't happen before `gst_samples` is called then
`wait_for_match("menu.png")` returns false positive as the first sample
that `gst_samples` yields is the one that had been stored _before_ the
device was powered off.

drothlis:
There's the more general issue of video-capture latency. For example in `wait_for_transition` (see #445) we do something like this:

```
stbt.press("KEY_DOWN")
press_time = time.time()

# Wait for animation to start
while True:
    f = stbt.get_frame()
    if f.time < press_time:
        continue  # Discard frame to work around latency in video-capture pipeline
    ...
```

I'm not sure what to do about this in the general case. Perhaps `stbt.get_frame` and `stbt.frames` could have similar logic as the above, and discard frames older than the time at which `get_frame` was called. But that would probably break things for people with badly-configured source pipelines (with incorrect timestamps). I don't know how common such bad configurations are, but it's not trivial to get right, so probably quite common.

wmanley:
>Perhaps `stbt.get_frame` could have similar logic as the above

[...]

PR #445, opened 2017-07-19
[WIP] Add wait_for_transition

drothlis:
Useful for performance tests involving animation (e.g. Selection moving) or waiting for a screen (e.g. Guide) to populate fully.

TODO:

- [ ] Merge #442 and rebase.
- [ ] Make it a class, since so much is shared between
  `wait_for_transition` and `wait_for_transition_to_end`.
- [ ] Tests.


wmanley:
I've been thinking about this and I wonder if it would be better to have just a single `wait_for_transition` function that doesn't press.   The `DeviceUnderTest` class already records a `_time_of_last_press`.  We could even enhance the `press` function to store the last frame before the press happened if we want to avoid race conditions.

The mental model I'm basing this on is the same one from [this comment](https://github.com/stb-tester/stb-tester/issues/246#issuecomment-317482218) on #246.

wmanley:
These functions need to be made threadsafe using `stbt.frames()` rather than `get_frame()` like in #449

wmanley:
I think this should be moved into the `stbt` namespace.  I think it's easier for users if all our functions are under `stbt` - that way they don't need to look up where they should be importing them from.

drothlis:
Rebased onto master to fix conflicts.

stbt/transition.py
```diff
@@ -148,21 +148,21 @@ def wait_for_transition(self, key):
...
-            initial_frame = self.dut.get_frame()
+            initial_frame, _ = next(self.frames)
         if self.expiry_time is None:
             self.expiry_time = initial_frame.time + self.timeout_secs
 
         f = first_stable_frame = initial_frame
         while True:
             prev = f
-            f = self.dut.get_frame()
+            f, _ = next(self.frames)
```
wmanley: This possibly could be clearer as a for loop.  Not important.

stbt/__init__.py
```diff
@@ -39,6 +39,11 @@
     UITestError, \
     UITestFailure, \
     wait_until
+from _stbt.transition import \
+    TransitionResult, \
```
wmanley: I don't think `TransitionResult` belongs here.  Users should not be constructing them.

stbt/__init__.py
```diff
@@ -74,10 +79,14 @@
     "Region",
     "save_frame",
     "TextMatchResult",
+    "TransitionResult",
```
wmanley: I don't think `TransitionResult` belongs here. Users should not be constructing them.

_stbt/transition.py
```diff
@@ -58,7 +57,7 @@ def press_and_wait(
 
     t = _Transition(region, mask, timeout_secs, stable_secs, _dut)
     result = t.press_and_wait(key)
-    sys.stderr.write("%s\n" % result)
+    stbt.debug(str(result))
```
wmanley: This logging could be better.  Perhaps:

    stbt.debug("press_and_wait(%r) -> %r" % (key, result))

tests/test_android.py
```diff
@@ -92,8 +93,8 @@ def test_to_native_coordinates(
         screenshot_x, screenshot_y, coordinate_system,
         _Dimensions(*device_resolution))
     print(native_x, native_y)
-    assert isclose(native_x, expected_coordinates[0], absolute_tolerance=1)
-    assert isclose(native_y, expected_coordinates[1], absolute_tolerance=1)
+    assert isclose(native_x, expected_coordinates[0], atol=1)
+    assert isclose(native_y, expected_coordinates[1], atol=1)
```
wmanley: `pytest.Approx` is even nicer.

tests/test_transition.py
```diff
@@ -0,0 +1,140 @@
...
+    elif state == "white":
+        array = numpy.ones((720, 1280, 3), dtype=numpy.uint8) * 255
+    elif state in ["fade-to-black", "fade-to-white"]:
+        array = numpy.ones((720, 1280, 3), dtype=numpy.uint8) * 127
+    elif state == "ball":
+        # black background, white ball that moves by 1 pixel ever 10ms
+        # in the left half of the frame.
+        array = numpy.zeros((720, 1280, 3), dtype=numpy.uint8)
+        cv2.circle(array, (int(t * 100) % 625, 360), 15, (255, 255, 255))
+    return array
```
wmanley: This is neat.

_stbt/transition.py
```diff
@@ -127,6 +127,7 @@ def press_and_wait(self, key):
         original_frame, _ = next(self.frames)
         self.dut.press(key)
         press_time = time.time()
+        stbt.debug("transition: %.3f: Pressed %s" % (press_time, key))
```
wmanley: Perhaps this log belongs in `DeviceUnderTest.press` if we want to be sure that we're logging every press.

tests/test_android.py
drothlis: Yeah, we'd have to upgrade pytest in .travis.yml.

_stbt/transition.py
drothlis: We can add that later (and have `stbt.press` return the time of the keypress for `press_and_wait` to use).

PR #447, opened 2017-07-19
stbt.load_image: Add errno to IOError exceptions

drothlis:
We change the exception type In 68c0d92 ("stbt.load_image: Raise IOError
instead of ValueError") but it turns out that IOErrors can (should?)
have an errno & filename attribute:

> When exceptions of this type are created with a 2-tuple, the first
> item is available on the instance’s errno attribute (it is assumed to
> be an error number), and the second item is available on the strerror
> attribute (it is usually the associated error message). The tuple
> itself is also available on the args attribute.
>
> When an EnvironmentError exception is instantiated with a 3-tuple, the
> first two items are available as above, while the third item is
> available on the filename attribute. However, for backwards
> compatibility, the args attribute contains only a 2-tuple of the first
> two constructor arguments.

https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError

(Note that IOError inherits from EnvironmentError.)

---

TODO:

- [ ] Fix `stbt run` output "FAIL: test.py: IOError: 2"

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -538,10 +538,11 @@ def _load_template(template):
         relative_filename = template
         absolute_filename = _find_user_file(relative_filename)
         if not absolute_filename:
-            raise IOError("No such template file: %s" % relative_filename)
+            raise IOError(errno.ENOENT, "No such template file",
+                          relative_filename)
         image = cv2.imread(absolute_filename, cv2.CV_LOAD_IMAGE_COLOR)
         if image is None:
-            raise IOError("Failed to load template file: %s" %
+            raise IOError(errno.EIO, "Failed to load template file",
                           absolute_filename)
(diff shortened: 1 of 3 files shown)


drothlis:
This was raised by @wmanley in a post-merge review of #442.

drothlis:
With this change `stbt run` prints:

    FAIL: test.py: IOError: 2'

instead of:

    FAIL: test.py: IOError: No such template file: idontexist.png

drothlis:
Given that the 3-argument form of `IOError` doesn't seem to be mandatory (from a generous interpretation of the [EnvironmentError] documentation) and the above regression in behaviour (originating from `kitchen.text.converters.exception_to_bytes` so it isn't easily fixable) I'm tempted to say that we leave it as is.

[EnvironmentError]: https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError

Issue #266, opened 2015-03-03
Multi-threading issue with Python Threading Lib (0.21)


mxd-rglas:
Whenever I use mutli-threading in my stbt scripts (for detecting random errors that cannot be expected by the script, like a popup error after disconnecting the network cable manually), from time to time there seems to be race conditions about the video frame, if I got the error message right. The error is not reproducible 100% and happens only intermittent. Most probably this could be fixed by Lock, but I would assume the stbt internals themselves need to respect the Lock as well.

Is there a plan to support multi-threading and fix this?

_Exception_:

```
UnboundLocalError: local variable 'res' referenced before assignment
Traceback (most recent call last):
 File "/usr/lib/stbt/stbt-run", line 49, in <module>
    execfile(args.script)
 File "/usr/lib/stbt/stbt.py", line 613, in wait_for_match
    raise MatchTimeout(res.frame, _template_name(image), timeout_secs)  # pylint: disable=W0631,C0301
```

_Code example_:

```
import threading
self._stop = threading.Event()
self._thread = threading.Thread(name='threadname', 
        target=do_something_in_parallel, 
        args=(self._stop,1))
self._thread.daemon = True
self._thread.start()
 # wait for something special to happen
self._stop.set() # Fire the stop event to stop the parallel action

def do_something_in_parallel(stop):
  while not stop.isSet():
    # do your parallel stuff
```

wmanley:
This sounds like a reasonable approach, but I'm not surprised that it doesn't work 100% as this is not a use case that we have any automated tests for in stb-tester itself.

drothlis:
This should be fixed by #449.

PR #449, opened 2017-08-02
Display: Allow multiple threads to call `get_frame()` and `frames()`

wmanley:
Previously `get_frame()` would have a side-effect that we'd never get the
same frame twice.  Instead it would block waiting for a new frame.  This is
useful behaviour for simple loops like:

    wait_until(lambda: match("example.png"))

but it's less desirable for:

    wait_until(lambda: match("example.png") or match("alternative.png"))

as waiting for a different frame to match "alternative.png" is unnecessary
and slows the process down.

It goes particularly wrong for cases like:

    for x, y in izip(detect_motion(),
                     iter(lambda: stbt.match("error-dialog.png"))):
        ...

Which frames will be analysed by `detect_motion` and which matched against?
It's unclear.

With this change a call to `get_frame()` will return immediately with the
latest frame and multiple `stbt.frames()` iterators can be used
concurrently (even from separate threads).  This will make it easier to
implement some more complex performance measurement tests.

The effects of this change could be rather subtle.  I would expect there to
be some loops out there that become busy-waits.  Whether that will actually
cause problems or not is a seperate question.

This is built on top of #420 and should provide a solution to #266 and #246

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -740,8 +740,11 @@ def new_device_under_test_from_config(
         display[0].tell_user_thread(exception)
     mainloop = _mainloop()
 
-    sink_pipeline = SinkPipeline(
-        gst_sink_pipeline, raise_in_user_thread, save_video)
+    if not gst_sink_pipeline and not save_video:
+        sink_pipeline = NoSinkPipeline()
+    else:
+        sink_pipeline = SinkPipeline(  # pylint: disable=redefined-variable-type
+            gst_sink_pipeline, raise_in_user_thread, save_video)
     display[0] = Display(
         gst_source_pipeline, sink_pipeline, restart_source,
         transformation_pipeline)
@@ -1786,6 +1789,25 @@ class SinkPipeline(object):
             else:
                 raise TypeError(
                     "Can't draw object of type '%s'" % type(obj).__name__)
+
+
+class NoSinkPipeline(object):
+    """
+    Used in place of a SinkPipeline when no video output is required.  Is a lot
+    faster because it doesn't do anything.  It especially doesn't do any copying
+    nor video encoding :).
+    """
+    def startup(self):
+        pass
+
+    def teardown(self):
+        pass
+
+    def push_sample(self, _sample):
+        pass
+
+    def draw(self, _obj, _duration_secs=None, label=""):
+        pass
 
 
 class Display(object):
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1641,6 +1641,135 @@ class _Annotation(namedtuple("_Annotation", "time region label colour")):
         _draw_text(img, self.label, label_loc, (255, 255, 255), font_scale=0.5)
 
 
+class SinkPipeline(object):
+    def __init__(self, user_sink_pipeline, raise_in_user_thread, save_video=""):
+        self.annotations_lock = threading.Lock()
+        self.text_annotations = []
+        self.annotations = []
+        self._raise_in_user_thread = raise_in_user_thread
+        self.received_eos = threading.Event()
+
+        if save_video:
+            if not save_video.endswith(".webm"):
+                save_video += ".webm"
+            debug("Saving video to '%s'" % save_video)
+            video_pipeline = (
+                "t. ! queue leaky=downstream ! videoconvert ! "
+                "vp8enc cpu-used=6 min_quantizer=32 max_quantizer=32 ! "
+                "webmmux ! filesink location=%s" % save_video)
+        else:
+            video_pipeline = ""
+
+        sink_pipeline_description = " ".join([
+            "appsrc name=appsrc format=time " +
+            "caps=video/x-raw,format=(string)BGR !",
+            "tee name=t",
+            video_pipeline,
+            "t. ! queue leaky=downstream ! videoconvert !",
+            user_sink_pipeline
+        ])
+
+        self.sink_pipeline = Gst.parse_launch(sink_pipeline_description)
+        sink_bus = self.sink_pipeline.get_bus()
+        sink_bus.connect("message::error", self._on_error)
+        sink_bus.connect("message::warning", self._on_warning)
+        sink_bus.connect("message::eos", self._on_eos_from_sink_pipeline)
+        sink_bus.add_signal_watch()
+        self.appsrc = self.sink_pipeline.get_by_name("appsrc")
+
+        debug("sink pipeline: %s" % sink_pipeline_description)
+
+    def _on_eos_from_sink_pipeline(self, _bus, _message):
+        debug("Got EOS from sink pipeline")
+        self.received_eos.set()
+
+    def _on_warning(self, _bus, message):
+        assert message.type == Gst.MessageType.WARNING
+        Gst.debug_bin_to_dot_file_with_ts(
+            self.sink_pipeline, Gst.DebugGraphDetails.ALL, "WARNING")
+        err, dbg = message.parse_warning()
+        warn("Warning: %s: %s\n%s\n" % (err, err.message, dbg))
+
+    def _on_error(self, _bus, message):
+        assert message.type == Gst.MessageType.ERROR
+        if self.sink_pipeline is not None:
+            Gst.debug_bin_to_dot_file_with_ts(
+                self.sink_pipeline, Gst.DebugGraphDetails.ALL, "ERROR")
+        err, dbg = message.parse_error()
+        self._raise_in_user_thread(
+            UITestError("%s: %s\n%s\n" % (err, err.message, dbg)))
+
+    def startup(self):
+        self.received_eos.clear()
+        self.sink_pipeline.set_state(Gst.State.PLAYING)
+
+    def teardown(self):
+        debug("teardown: Sending eos on sink pipeline")
+        if self.appsrc.emit("end-of-stream") == Gst.FlowReturn.OK:
+            if not self.received_eos.wait(10):
+                debug("Timeout waiting for sink EOS")
+        else:
+            debug("Sending EOS to sink pipeline failed")
+
+        self.sink_pipeline.set_state(Gst.State.NULL)
+
+        # Don't want to cause the Display object to hang around on our account,
+        # we won't be raising any errors from now on anyway:
+        self._raise_in_user_thread = None
+
+    def push_sample(self, sample):
+        # Calculate whether we need to draw any annotations on the output video.
+        now = sample.time
+        texts = []
+        annotations = []
+        with self.annotations_lock:
+            for x in self.text_annotations:
+                x.setdefault('start_time', now)
+                if now < x['start_time'] + x['duration']:
+                    texts.append(x)
+            self.text_annotations = texts[:]
+            for annotation in list(self.annotations):
+                if annotation.time == now:
+                    annotations.append(annotation)
+                if now >= annotation.time:
+                    self.annotations.remove(annotation)
+
+        sample = gst_sample_make_writable(sample)
+        img = array_from_sample(sample, readwrite=True)
+        # Text:
+        _draw_text(
+            img, datetime.datetime.now().strftime("%H:%M:%S.%f")[:-4],
+            (10, 30), (255, 255, 255))
+        for i, x in enumerate(reversed(texts)):
+            origin = (10, (i + 2) * 30)
+            age = float(now - x['start_time']) / 3
+            color = (int(255 * max([1 - age, 0.5])),) * 3
+            _draw_text(img, x['text'], origin, color)
+
+        # Regions:
+        for annotation in annotations:
+            annotation.draw(img)
+
+        self.appsrc.props.caps = sample.get_caps()
+        self.appsrc.emit("push-buffer", sample.get_buffer())
+
+    def draw(self, obj, duration_secs=None, label=""):
+        with self.annotations_lock:
+            if isinstance(obj, str) or isinstance(obj, unicode):
+                obj = (
+                    datetime.datetime.now().strftime("%H:%M:%S.%f")[:-4] +
+                    ' ' + obj)
+                self.text_annotations.append(
+                    {"text": obj, "duration": duration_secs})
+            elif hasattr(obj, "region") and hasattr(obj, "timestamp"):
+                annotation = _Annotation.from_result(obj, label=label)
+                if annotation.time:
+                    self.annotations.append(annotation)
+            else:
+                raise TypeError(
+                    "Can't draw object of type '%s'" % type(obj).__name__)
+
+
 class Display(object):
     def __init__(self, user_source_pipeline, user_sink_pipeline,
                  mainloop, save_video="",
@@ -1658,11 +1787,6 @@ class Display(object):
         self.start_timestamp = None
         self.underrun_timeout = None
         self.tearing_down = False
-        self.received_eos = threading.Event()
-
-        self.annotations_lock = threading.Lock()
-        self.text_annotations = []
-        self.annotations = []
 
         self.restart_source_enabled = restart_source
 
@@ -1691,47 +1815,17 @@ class Display(object):
             appsink])
         self.create_source_pipeline()
 
-        if save_video:
-            if not save_video.endswith(".webm"):
-                save_video += ".webm"
-            debug("Saving video to '%s'" % save_video)
-            video_pipeline = (
-                "t. ! queue leaky=downstream ! videoconvert ! "
-                "vp8enc cpu-used=6 min_quantizer=32 max_quantizer=32 ! "
-                "webmmux ! filesink location=%s" % save_video)
-        else:
-            video_pipeline = ""
-
-        sink_pipeline_description = " ".join([
-            "appsrc name=appsrc format=time " +
-            "caps=video/x-raw,format=(string)BGR !",
-            "tee name=t",
-            video_pipeline,
-            "t. ! queue leaky=downstream ! videoconvert !",
-            user_sink_pipeline
-        ])
-
-        self.sink_pipeline = Gst.parse_launch(sink_pipeline_description)
-        sink_bus = self.sink_pipeline.get_bus()
-        sink_bus.connect(
-            "message::error",
-            lambda bus, msg: self.on_error(self.sink_pipeline, bus, msg))
-        sink_bus.connect("message::warning", self.on_warning)
-        sink_bus.connect("message::eos", self.on_eos_from_sink_pipeline)
-        sink_bus.add_signal_watch()
-        self.appsrc = self.sink_pipeline.get_by_name("appsrc")
+        self._sink_pipeline = SinkPipeline(
+            user_sink_pipeline, self.tell_user_thread, save_video)
 
         debug("source pipeline: %s" % self.source_pipeline_description)
-        debug("sink pipeline: %s" % sink_pipeline_description)
         self.mainloop = mainloop
 
     def create_source_pipeline(self):
         self.source_pipeline = Gst.parse_launch(
             self.source_pipeline_description)
         source_bus = self.source_pipeline.get_bus()
-        source_bus.connect(
-            "message::error",
-            lambda bus, msg: self.on_error(self.source_pipeline, bus, msg))
+        source_bus.connect("message::error", self.on_error)
         source_bus.connect("message::warning", self.on_warning)
         source_bus.connect("message::eos", self.on_eos_from_source_pipeline)
         source_bus.add_signal_watch()
@@ -1808,7 +1902,7 @@ class Display(object):
                 try:
                     yield array_from_sample(sample)
                 finally:
-                    self.push_sample(sample)
+                    self._sink_pipeline.push_sample(sample)
 
     def on_new_sample(self, appsink):
         sample = appsink.emit("pull-sample")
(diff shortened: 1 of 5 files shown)


wmanley:
@drothlis: This is a change with rather significant impact, but it has allowed me to implement tests that I otherwise wouldn't be able to do.  How do you feel about it?

drothlis:
I haven't reviewed the implementation yet, but this is great. It allows writing tests like this:

```
pool = multiprocessing.ThreadPool(processes=3)
stbt.press("KEY_OK")
pool.apply_async(motion_watcher)
pool.apply_async(black_screen_watcher)
pool.apply_async(spinner_watcher)
pool.close()
pool.join()
```

...which could log output like this:

```
Measured 'play_content' at 2017-08-03T12:16:57.208350+00:00
    time to press	0.0
    time to spinner	0.13856959343
    time to black	0.178586959839
    time to not spinner	1.3785636425
    time to not black	2.45867109299
    time to motion	3.41855049133
```

drothlis:
As discussed with @wmanley: We should make `stbt.get_frame()` have an optional configuration flag (global, not a parameter to the function) that makes it use the original behaviour (blocking until the next frame is available). Tell users that if they have any problems, set that flag *and tell us* otherwise we'll remove it in a future release.

wmanley:
Whew, I've completed all the TODO items.  I do like the look of the [example test](https://github.com/stb-tester/stb-tester/pull/449/files#diff-aae2f0e80f5f6d6829c32afd358f813bR582).  We should take some inspiration from this when defining a better threading API #463.  Specifically I like how `imap_unordered` gives us an iterator of things that have happened, and will pass through exceptions, etc too.

@drothlis: Please review in detail.

We should watch out for any test-unreliability.  I've seen `test_detect_motion_times_out_during_yield` fail.  See https://travis-ci.org/stb-tester/stb-tester/builds/298045248 .  I had put a fix in for this.  See https://github.com/wmanley/stb-tester/commit/1bde34547810ddbf33de5fbdd7f9588d2abf245f, but perhaps the cure is worse than the disease in that instance.

drothlis:
`test_detect_match_times_out_during_yield` and `test_detect_motion_times_out_during_yield` are still failing. Is 1bde345 not actually on this branch? (it's on branch `one-past-end`). Clearly we can't merge if it has flaky tests.

_stbt/core.py
```diff
@@ -1907,7 +1906,6 @@ def create_source_pipeline(self):
             # Hauppauge HDPVR capture device.
             source_queue = self.source_pipeline.get_by_name(
                 "_stbt_user_data_queue")
-            self.start_timestamp = None
```
drothlis: `start_timestamp` was changed from a stack variable to a member in cf57a4c (First `wait_for_match` after restarting pipeline obeys `timeout_secs`) because the HD-PVR video-capture device would start its timestamps at 0 after restarting the source pipeline.

I would hope that any other video-capture device in common use has sensible timestamps, and I don't care about supporting the HD-PVR... but this certainly isn't a pure refactoring.

_stbt/core.py
```diff
@@ -1948,28 +1946,25 @@ def pull_frame(self, timeout_secs=10):
 
     def frames(self, timeout_secs=None):
         import time
-        self.start_timestamp = None
+        if timeout_secs is not None:
+            end_time = time.time() + timeout_secs
```
drothlis: This is also a change in behaviour in that the timeout is based on wall clock time instead of the time of the first frame.

It will break video-capture devices that don't use wall-clock time. I think that's OK -- as far as I know that's only the HD-PVR which is obsolete.

It will change test-script behaviour for video-capture devices with significant latency --  but in a safe way, i.e. it will result in more frames being processed, not less. So if you ask for timeout_secs=10, you'll actually get 10 seconds' worth since your `press` or whatever.

_stbt/core.py
```diff
@@ -1948,28 +1946,25 @@ def pull_frame(self, timeout_secs=10):
...
 
         with self.lock:
             while True:
                 ddebug("user thread: Getting sample at %s" % time.time())
-                sample = self._get_sample(max(10, timeout_secs))
+                sample = self._get_sample(max(10, timeout_secs or 0))
```
drothlis: This is an unrelated change. It's also unnecessary:

```
>>> max(10, None)
10
```

But it doesn't hurt.

_stbt/core.py
drothlis: To clarify I don't object to this change but the commit message needs to change and it could possibly be split into several commits and possibly merits a release note.

_stbt/core.py
```diff
@@ -1948,28 +1946,25 @@ def pull_frame(self, timeout_secs=10):
...
-
                 try:
                     yield array_from_sample(sample)
                 finally:
                     self._sink_pipeline.push_sample(sample)
 
+                if timeout_secs is not None and timestamp > end_time:
+                    debug("timed out: %.3f > %.3f" % (timestamp, end_time))
+                    return
+
```
drothlis: So now we're checking for timeout *after* yielding the frame -- also a change in behaviour. It's kind of reverted by 1bde345 on branch `one-past-end`.

_stbt/core.py
```diff
@@ -1698,6 +1698,7 @@ def __init__(self, user_sink_pipeline, raise_in_user_thread, save_video=""):
         self.annotations = []
         self._raise_in_user_thread = raise_in_user_thread
         self.received_eos = threading.Event()
+        self._frames = deque(maxlen=35)
```
drothlis: Does this assume a particular video-capture frame rate?

_stbt/core.py
```diff
@@ -1767,7 +1772,26 @@ def teardown(self):
...
 
-    def push_sample(self, sample):
+    def on_sample(self, sample):
+        """
+        Called from `Display` for each frame.
+        """
+        # The test script can draw on the video, but this happens in a different
+        # thread.  We don't know when they're finished drawing so we just give
+        # them 0.5s instead.
+        SINK_LATENCY_SECS = 0.5
```
drothlis: This says 0.5s but the commit message says 1s.

_stbt/core.py
```diff
@@ -933,9 +933,6 @@ def _match_all(self, image, frame, match_parameters, region):
 
         region = Region.intersect(_image_region(frame), region)
 
-        # Remove this copy once we've confirmed that doing so won't cause
-        # push_sample to doodle on the frame later.
-        frame = frame.copy()
         frame.flags.writeable = False
```
drothlis: Presumably we can remove this `frame.flags.writeable = False` too, because we already set it in `Display.on_new_sample`.

_stbt/core.py
```diff
@@ -933,9 +933,6 @@ def _match_all(self, image, frame, match_parameters, region):
 
         region = Region.intersect(_image_region(frame), region)
 
-        # Remove this copy once we've confirmed that doing so won't cause
-        # push_sample to doodle on the frame later.
-        frame = frame.copy()
```
drothlis: Nice, earlier in the pull request I was wondering if you were going to get around to removing this copy. Should we add a test that we don't write on the frame? Something like:

```
get_frame
sleep
look at the frame
```

_stbt/core.py
```diff
@@ -1979,11 +1998,7 @@ def on_new_sample(self, appsink):
                  "source-pipeline configuration." % sample.time)
 
         self.tell_user_thread(sample)
-        if self.lock.acquire(False):  # non-blocking
-            try:
-                self._sink_pipeline.push_sample(sample)
-            finally:
-                self.lock.release()
+        self._sink_pipeline.on_sample(sample)
```
drothlis: Before, `SinkPipeline.push_sample` would sometimes get called from the GLib main loop (i.e. the thread handling the GStreamer `new-sample` signal) and sometimes from the user thread that's running the user script (if using `stbt.frames()`, `stbt.wait_for_match()`, `stbt.wait_for_motion()`, or similar -- but *not* `stbt.get_frame()`).

N.B. `SinkPipeline.push_sample` is what draws on the frames. But previously if called from the GLib main loop there would be nothing to draw. Whereas now it will always do the drawing from the GLib main loop. Will that slow down the main loop from processing other events? Probably not. Just wondering out loud.

This way is certainly simpler & more consistent. I feel like this whole PR is making things simpler.

docs/release-notes.md
```diff
@@ -112,6 +112,11 @@ UNRELEASED
   working properly with OpenCV 3. OpenCV 2.4 is still our primary supported
   target version of OpenCV.
 
+* Output video now runs a the full frame-rate of the input video rather than
+  slowing down during `wait_for_match`. As a side-effect the latency of the
+  video has increased by 0.5s and if the image processing is particularly slow
+  the annotations won't appear on the output video.
```
drothlis: Also: Annotations added manually when using `stbt.match` + `stbt.get_frame` intsead of `stbt.wait_for_match` will now show up in the video, as long as your processing took < 0.5s.

drothlis: P.S. The long term goal is to have these annotations rendered at playback time (e.g. when looking at a video for an old test result) instead of hard-coding them into the video at test-run time.

_stbt/core.py
```diff
@@ -1962,7 +1962,7 @@ def _get_sample(self, timeout_secs=10):
         return gst_sample
```
drothlis: Looks like you missed this spot -- you should be returning the result of `array_from_sample` from the previous line?

_stbt/core.py
```diff
@@ -903,7 +902,7 @@ def _match_all(self, image, frame, match_parameters, region):
 
         grabbed_from_live = (frame is None)
         if grabbed_from_live:
-            frame = self._display.pull_frame()
+            frame = self._display.get_frame()
```
drothlis: Yay I was always confused about the proliferation of `get_frame` / `_get_sample` / `pull_frame`. Now we're rid of the latter 2.

_stbt/core.py
```diff
@@ -1999,25 +2006,20 @@ def on_new_sample(self, appsink):
         return Gst.FlowReturn.OK
 
     def tell_user_thread(self, frame_or_exception):
-        # `self.last_sample` (a synchronised Queue) is how we communicate from
-        # this thread (the GLib main loop) to the main application thread
-        # running the user's script. Note that only this thread writes to the
-        # Queue.
+        # `self.last_frame` is how we communicate from this thread (the GLib
+        # main loop) to the main application thread running the user's script.
+        # Note that only this thread writes to the Queue.
```
drothlis: only this thread writes to ~~the Queue~~ \`self.last_frame`.

_stbt/core.py
```diff
@@ -1942,36 +1942,43 @@ def set_source_pipeline_playing(self):
...
-
-        if isinstance(gst_sample, Gst.Sample):
-            self.last_used_frame = array_from_sample(gst_sample)
-
-        return gst_sample
-
-    def pull_frame(self, timeout_secs=10):
-        return self._get_frame(timeout_secs)
+    def get_frame(self, timeout_secs=10, since=None):
+        import time
```
drothlis: How expensive is `import time`?

```
In [1]: import time

In [2]: %timeit import time
The slowest run took 43.52 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 367 ns per loop
```

Worst case: 367ns x 43.52 = 0.016ms = 1/1,000th of a frame at 60fps. That's the worst case (the slowest run). Maybe %timeit starts from a clean environment and imports it afresh?

Best case: 367ns = 1/45,000th of a frame at 60fps.

docs/release-notes.md
```diff
@@ -77,8 +77,28 @@ UNRELEASED
   of your stbt config file, and let us know on the mailing list as we may
   remove the workaround completely in a future release.
 
+* A call to `stbt.get_frame()` is no-longer guaranteed to return a new frame, it
+  may return the same frame that the previous call to `stbt.get_frame()`
+  returned. This may have subtle effects on the timing of existing test-scripts.
+  Functions that depend on this behaviour should be refactored to use the
+  `stbt.frames()` iterator method instead.
```
drothlis: "no longer blocks until a new frame is available."

The benefits are that you can call `stbt.get_frame` from multiple threads.

tests/test-stbt-py.sh
```diff
@@ -578,3 +578,42 @@ test_that_transformation_pipeline_transforms_video() {
...
+
+    stbt run -v test.py checkers-8 >out.log
+    grep -q "Checkers" out.log || fail "Expected checkers pattern"
+
+    stbt run -v test.py smpte >out.log
+    grep -q "Motion" out.log || fail "Expected motion"
+
+    stbt run -v test.py black >out.log
+    grep -q "Timeout" out.log || fail "Expected timeout"
+}
```
drothlis: Nice test.

_stbt/core.py
```diff
@@ -1189,8 +1191,22 @@ def match_text(self, text, frame=None, region=Region.ALL,
...
+        timestamp = None
+
+        while True:
+            ddebug("user thread: Getting sample at %s" % self._time.time())
+            frame = self._display.get_frame(
+                max(10, timeout_secs or 0), since=timestamp)
+            ddebug("user thread: Got sample at %s" % self._time.time())
+            timestamp = frame.time
+
+            yield frame.copy(), int(f.time * 1e9)
```
drothlis: `f.time` should be `frame.time`; or better, `timestamp`.

_stbt/core.py
```diff
@@ -2991,25 +2989,41 @@ def test_wait_for_motion_half_motion_int():
...
+
+        def time(self):
+            return self._now
+
+        def sleep(self, duration):
+            self._now += duration
+
+    dut = DeviceUnderTest(display=FakeDisplay(), sink_pipeline=NoSinkPipeline(),
+                          _time=FakeTime(FRAMES[0].time))
     dut.get_frame = lambda: None
```
drothlis: Do we still need `dut.get_frame = lambda: None`?

tests/test-match.sh
```diff
@@ -555,3 +555,32 @@ test_that_matchtimeout_screenshot_doesnt_include_visualisation() {
...
+
+test_global_use_old_threading_behaviour() {
+    set_config global.use_old_threading_behaviour true
+
+    cat > test.py <<-EOF &&
+	ts = set()
+	for _ in range(10):
+	    ts.add(stbt.get_frame().time)
+	assert len(ts) == 10
+	print "Saw %i unique frames" % len(ts)
```
drothlis: This print should go before the assert, to help debugging test failures.

tests/test-match.sh
```diff
@@ -555,3 +555,32 @@ test_that_matchtimeout_screenshot_doesnt_include_visualisation() {
...
+        || fail "use_old_threading_behaviour Warning not printed"
+
+    set_config global.use_old_threading_behaviour false
+
+    cat > test.py <<-EOF &&
+	ts = set()
+	for _ in range(10):
+	    ts.add(stbt.get_frame().time)
+	assert len(ts) < 5
+	print "Saw %i unique frames" % len(ts)
```
drothlis: This print should go before the assert, to help debugging test failures.

tests/test-match.sh
```diff
@@ -555,3 +555,32 @@ test_that_matchtimeout_screenshot_doesnt_include_visualisation() {
     stbt match screenshot.png "$testdir"/black-full-frame.png \
         match_method=ccoeff-normed
 }
+
+test_global_use_old_threading_behaviour() {
```
drothlis: This test probably belongs in `test-stbt-py.sh`.

drothlis: We should also test `stbt.frames()`.

_stbt/core.py
wmanley: I see.  I hadn't realised that `self.start_timestamp` is reset here: https://github.com/stb-tester/stb-tester/blob/50caf83fa4406970c4690ad0f1f9d887e1db7b45/_stbt/core.py#L1905-L1912

So it can be modified from a separate thread.  Hmm this is racy as well.  Quite a narrow race though.  Hmmmm...

wmanley: This commit does cause the breaking change that now `stbt.frames` can yield 0 frames, whereas before it was guaranteed to yield at least one.  I'll fix this with an explicit check.

_stbt/core.py
wmanley: I'll undo it in a fixup

_stbt/core.py
wmanley: I've fixed this in a fixup.

_stbt/core.py
wmanley: Good, fixing this explicitly has made CI pass.

_stbt/core.py
wmanley: I suppose so.  This is just over 0.5s at 60fps.  I wanted to put a limit in to prevent runaway memory usage.  This will be 96MB of video frame data.

_stbt/core.py
wmanley: > Before, SinkPipeline.push_sample would sometimes get called from the GLib main loop (i.e. the thread handling the GStreamer new-sample signal)

I think it would be called in the context of the "new-sample" signal from GStreamer.  This will be called from the GStreamer stream thread.  Bus messages are posted to the GLib main thread.  Signals are dispatched in-place (usually I think).

That's an interesting point about performance.  on-new-sample can't be dispatched again until we return from here.  Arguably the sink pipeline should have its own thread for this stuff.  I doubt that it would hurt too much though.  This won't be an issue if we ever get round to ditching the sink pipeline entirely. (one day...).

docs/release-notes.md
wmanley: I'm having difficulty wording this.  I've added a TODO item.

_stbt/core.py
wmanley: Fixed

docs/release-notes.md
wmanley: Fixed.

tests/test-stbt-py.sh
wmanley: Yes, I like how `imap_unordered` takes care of a lot of the tricky threading details.

_stbt/core.py
wmanley: I think I fix this later in the series.  I'll consider fixing it in the rebase.

_stbt/core.py
wmanley: Probably not.  Removed.

tests/test-match.sh
wmanley: > This test probably belongs in test-stbt-py.sh.

fixed.

tests/test-match.sh
wmanley: Good idea, fixed

_stbt/core.py
wmanley: Good idea.

_stbt/core.py
wmanley: Yes, good idea.  I see you've done this in a fixup.

tests/test-stbt-py.sh
```diff
@@ -432,7 +432,7 @@ test_that_visualisation_doesnt_write_to_user_frame() {
 	
 	f = stbt.get_frame()
 	stbt.draw_text("Hello")
-	time.sleep(3)  # wait for sink pipeline buffer to process the frame
+	time.sleep(1)  # wait for sink pipeline buffer to process the frame
```
wmanley: This seems unrelated.  Perhaps it should be squashed into the previous commit?

tests/test-stbt-py.sh
```diff
@@ -441,6 +441,18 @@ test_that_visualisation_doesnt_write_to_user_frame() {
...
+test_that_frames_are_read_only() {
+    cat > test.py <<-EOF
+	import stbt
+	
+	f = stbt.get_frame()
+	f[0][0][0] = 255
+	EOF
+    ! stbt run -v \
+        --source-pipeline 'videotestsrc pattern=black is-live=true' \
+        test.py
```
wmanley: What does the error message look like?

_stbt/core.py
wmanley: Done

_stbt/core.py
wmanley: Done

_stbt/core.py
wmanley: Fixed in rebase

tests/test-stbt-py.sh
```diff
@@ -617,3 +617,70 @@ test_multithreaded() {
...
+	        break
+	    sa.add(a.time)
+	    sb.add(b.time)
+	assert len(sa) == 10
+	assert len(sb) == 10
+	# sa and sb contain the same frames:
+	assert sa == sb
+	EOF
+    stbt run -vv test.py || fail "Incorrect frames() behaviour"
+}
```
wmanley: Here's the new test I added for `stbt.frames()` behaviour.  It's good because it doesn't use threads so it's deterministic.

wmanley:
Rebased.  The only outstanding TODO item is possibly documenting the fact that `stbt.match` annotations are more likely to turn up in the output video.  If noone can come up with decent wording we should merge tomorrow anyway.

drothlis:
Merged. I'll worry about the release notes wording at release time when I usually tidy up the release notes anyway.

drothlis:
With the test reliability fix, `test_global_use_old_threading_behaviour_frames` has passed 3060 consecutive times.

PR #450, opened 2017-08-10
Use NVidia hardware accelerated VP8 encoder

drothlis:
For encoding the test-run video, if NVidia hardware encoder is present.

TODO:

- [ ] Make it configurable.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1678,10 +1678,14 @@ class SinkPipeline(object):
             if not save_video.endswith(".webm"):
                 save_video += ".webm"
             debug("Saving video to '%s'" % save_video)
+            if os.path.exists("/dev/nvhost-msenc"):
+                vp8enc = "omxvp8enc bitrate=200000"
+            else:
+                vp8enc = "vp8enc cpu-used=6 min_quantizer=32 max_quantizer=32"
             video_pipeline = (
                 "t. ! queue leaky=downstream ! videoconvert ! "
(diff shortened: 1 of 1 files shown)


drothlis:
With one of my testcases this yields a recording of 620KB for a 24 second video. The text quality is slightly worse but the framerate is totally smooth (matching the capture framerate of 25fps though if you use `wait_for_match` or the `stbt.frames` iterator it gets limited to the rate you process frames, but that limitation is in the test-script not in the encoder -- #449 addresses that).

This is 2-5 times larger filesize than the current videos (depending on the CPU available).

With the same bitrate, setting `quality-level=0` or `=1` makes the quality much worse without reducing file size (even increasing it). `quality-level=2` is the default (valid values are 0, 1 or 2).

Setting `iframeinterval=200` doesn't seem to make any difference in quality or file size or seekability. The default is 30 or 60 according to 2 different NVidia forum posts which I can't find now (gst-inspect says "0").

None of the other properties seem relevant or worth playing with.

drothlis:
It would be worth making the video-encoding pipeline configurable, it would certainly make it easier to experiment with different settings. I think it's still worth providing smart defaults though.

It would be nice to also use the VAAPI VP8 hardware encoder if available, but I'm not sure how to determine if it's available. It's only in recent Intel hardware (Broadwell and newer).

drothlis:
Superseded by #457.

PR #453, opened 2017-10-05
Fix loading images with unicode filename

drothlis:


_stbt/core.py
```diff
@@ -2614,6 +2615,9 @@ def _find_user_file(filename):
     :returns: Absolute filename, or None if it can't find the file.
     """
 
+    if isinstance(filename, unicode):
+        filename = filename.encode(sys.getfilesystemencoding(), "replace")
```
wmanley: I feel like there's less to go wrong if we just hard-code utf-8 here.  It's Linux only afterall.

wmanley:
Other than that +1

_stbt/core.py
drothlis: Fixed.

PR #455, opened 2017-10-05
stbt.ocr: Make the default value of `lang` configurable

drothlis:
To avoid specifying it in every single call to `stbt.ocr`.

### DIFF
--- a/stbt/__init__.py
+++ b/stbt/__init__.py
@@ -342,7 +342,7 @@ def wait_for_motion(
 
 def ocr(frame=None, region=Region.ALL,
         mode=OcrMode.PAGE_SEGMENTATION_WITHOUT_OSD,
-        lang="eng", tesseract_config=None, tesseract_user_words=None,
+        lang=None, tesseract_config=None, tesseract_user_words=None,
         tesseract_user_patterns=None, text_color=None):
     r"""Return the text present in the video frame as a Unicode string.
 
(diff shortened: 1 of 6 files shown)


stbt/__init__.py
```diff
@@ -501,7 +501,9 @@ def is_screen_black(frame=None, mask=None, threshold=None):
       binary threshold is applied to the frame. The ``threshold`` value is in
       the range 0 (black) to 255 (white). The global default can be changed by
       setting ``threshold`` in the ``[is_screen_black]`` section of
-      :ref:`.stbt.conf`.
+      :ref:`.stbt.conf`. You may need to install the tesseract language pack;
+      see installation instructions `here
+      <https://stb-tester.com/manual/troubleshooting#install-ocr-language-pack>`_.
```
wmanley: This is useful documentation.  We could even have a lint for it, but it probably wouldn't be worth the effort.

stbt/__init__.py
```diff
@@ -108,7 +108,7 @@ def press(key, interpress_delay_secs=None):
 
         This defaults to 0.3. You can override the global default value by
         setting ``interpress_delay_secs`` in the ``[press]`` section of
-        stbt.conf.
+        :ref:`.stbt.conf`.
```
wmanley: A good usability improvement.

stbt/__init__.py
wmanley: This is in the wrong docstring.

drothlis: Oops moved that to `stbt.ocr` docstring.

PR #457, opened 2017-10-06
Make video-recording pipeline configurable

drothlis:


_stbt/core.py
```diff
@@ -743,13 +743,22 @@ def raise_in_user_thread(exception):
...
 
-    if not gst_sink_pipeline and not save_video:
+    save_video_pipeline = get_config("global", "save_video_pipeline", "")
+    if not save_video_pipeline and save_video:
+        if not save_video.endswith(".webm"):
+            save_video += ".webm"
+        debug("Saving video to '%s'" % save_video)
+        save_video_pipeline = (
+            "vp8enc cpu-used=6 min_quantizer=32 max_quantizer=32 ! "
+            "webmmux ! filesink location=%s" % save_video)
```
drothlis: This will make `stbt record` and `stbt camera calibrate` use the `global.save_video_pipeline` configuration (previously those 2 tools never save a video).

I suppose I could move the config item to `run.save_video_pipeline` (same section as `run.save_video`) and move this logic to `stbt-run`.

drothlis:
Actually this isn't necessary because we already have `sink_pipeline` and `stbt batch run --no-save-video`.

PR #458, opened 2017-10-13
stbt.MotionResult: Add frame attribute

drothlis:
The `MotionResult` returned by `stbt.wait_for_motion` and
`stbt.detect_motion` includes the video-frame that was analysed.
This allows you to perform additional analysis, for example if there was
no motion is the frame black?


wmanley:
Yep, this seems reasonable.

PR #459, opened 2017-10-13
core.py: EOS on source pipeline teardown is configurable (off by default)

drothlis:
We added this in 52579fa to work around a deadlock bug in decklinksrc.
This caused additional problems:

* That commit itself had a race, fixed in cc125fb;
* Non-live sources (e.g. filesrc for self-tests) now would take 10
  seconds to teardown, presumably because the EOS was caught behind a
  long queue of buffers. We tried to fix this in fadfe6e and e051ba8
  but we reverted those two commits (in 8f1b48c and 8ee2ab5) because it
  caused intermittent long teardown in live v4l sources.

Now I've just made it configurable, and we don't send the EOS at all by
default.

### DIFF
--- a/_stbt/config.py
+++ b/_stbt/config.py
@@ -23,7 +23,10 @@ def get_config(section, key, default=None, type_=str):
     config = _config_init()
 
     try:
-        return type_(config.get(section, key))
+        if type_ is bool:
+            return config.getboolean(section, key)
+        else:
+            return type_(config.get(section, key))
     except ConfigParser.Error as e:
         if default is None:
             raise ConfigurationError(e.message)
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -723,6 +723,12 @@ class TextMatchResult(object):
 def new_device_under_test_from_config(
         gst_source_pipeline=None, gst_sink_pipeline=None, control_uri=None,
         save_video=False, restart_source=None, transformation_pipeline=None):
+
+    # Note that all callers (stbt-run etc, via stbt.init_run) pass in all
+    # parameters explicitly; but the behaviour of reading the default values
+    # from config is useful for integrating with a different test-runner like
+    # pytest, or from a Jupyter Notebook.
+
     from _stbt.control import uri_to_remote
 
     if gst_source_pipeline is None:
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -722,7 +722,8 @@ class TextMatchResult(object):
 
 def new_device_under_test_from_config(
         gst_source_pipeline=None, gst_sink_pipeline=None, control_uri=None,
-        save_video=False, restart_source=None, transformation_pipeline=None):
+        save_video=False, restart_source=None, transformation_pipeline=None,
+        source_teardown_eos=None):
 
     # Note that all callers (stbt-run etc, via stbt.init_run) pass in all
     # parameters explicitly; but the behaviour of reading the default values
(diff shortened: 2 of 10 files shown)


drothlis:
With the teardown change, `make check-integrationtests parallel=xargs` takes 297s vs 443s (a 33% reduction, 2m26s time saved -- though in practice it won't be as much because we use `parallel -j4`).

I don't know how much of that was caused by the source-pipeline failing to tear down and us giving up on waiting for EOS on the appsink after 10 seconds, or if it is merely a second or two waiting for the EOS that does actually arrive (multiplied by many tests).

drothlis:
> I don't know how much of that was caused by the source-pipeline failing to tear down and us giving up on waiting for EOS on the appsink after 10 seconds, or if it is merely a second or two waiting for the EOS that does actually arrive (multiplied by many tests).

It was the former -- I ran `tests/run-tests.sh -v` to see the output from even the passing tests, and saw 13 instances of "Source pipeline did not teardown gracefully".

_stbt/core.py
```diff
@@ -720,22 +720,11 @@ def timestamp(self):
...
-        gst_source_pipeline = get_config('global', 'source_pipeline')
-    if gst_sink_pipeline is None:
-        gst_sink_pipeline = get_config('global', 'sink_pipeline')
-    if control_uri is None:
-        control_uri = get_config('global', 'control')
-    if restart_source is None:
-        restart_source = get_config('global', 'restart_source')
-    if transformation_pipeline is None:
-        transformation_pipeline = get_config('global',
-                                             'transformation_pipeline')
```
wmanley: I do make use of this sometimes from Jupyter notebooks and the like when I'm experimenting.  If possible I'd prefer the  defaults to be removed from stbt-run rather than here.

_stbt/config.py
```diff
@@ -20,6 +20,9 @@ def get_config(section, key, default=None, type_=str):
     found, unless `default` is specified (in which case `default` is returned).
     """
 
+    if type_ == bool:
+        type_ = lambda s: s.lower() in ("1", "yes", "true", "on")
+
```
wmanley: Hmm, this looks familiar.  Ah yes see 939300a on #449.

wmanley:
The [GStreamer release notes for 1.6](https://gstreamer.freedesktop.org/releases/1.6/), released on 2015-09-25 say:

> Blackmagic Design DeckLink capture and playback card support was rewritten from scratch; 2k/4k support; mode sensing

and @sdroege had a [blog post](https://coaxion.net/blog/2014/12/improved-gstreamer-support-for-blackmagic-decklink-cards/) about it on 2014-12-22.

GStreamer 1.8 was included in Ubuntu 16.04.

The point is: I wonder if we can just drop this work-around.  I think not sending EOS is technically correct GStreamer usage in this case and I'm becoming less keen on maintaining these workarounds for buggy video capture hardware/software.

drothlis:
> I wonder if we can just drop this work-around.

Let's leave it in (disabled by default, with the flag to turn it back on) for one release. I've added a note to the release notes: "Removed workaround for 3½ year old deadlock bug in decklinksrc. If this is still necessary, set `source_teardown_eos = True` in the `[global]` section of your stbt config file, and let us know on the mailing list as we may remove the workaround completely in a future release."

_stbt/config.py
drothlis: Cool, that implementation is nicer, I've stolen it.

_stbt/core.py
drothlis: OK, I've dropped this change, it wasn't really related anyway.

PR #460, opened 2017-10-15
Add threshold parameters in match_text

sc-stbt:
I added a d fuzzy_match based on difflib to tolerate wrong matching specially in special character.
I add unidecode on _hocr_find_phrase function.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1133,7 +1133,7 @@ class DeviceUnderTest(object):
         if frame is None:
             frame = self.get_frame()
         if threshold is None:
-            threshold = get_config ("match_text", "threshold")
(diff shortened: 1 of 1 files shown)


drothlis:
> _stbt/core.py:2895:4: [E0401(import-error), _hocr_find_phrase] Unable to import 'unidecode'

The unit tests fail because you've added a new dependency `unidecode`.

What problem does this new dependency solve? Would it be possible to add the fuzzy_match functionality without adding a new dependency?

sc-stbt:
Sorry for the delay, I was busy with other urgent task.
We used unidecode lib to avoid fail due to some special character, for instance (ç will be read as 5). to go ahead on our test we passed unidecode word to be matched. but of course we can skip this lib.

PR #461, opened 2017-10-18
Stbt control relay: Allow lirc remote-control names with hyphens; return error for invalid commands

drothlis:


wmanley:
Yep, nice.

PR #462, opened 2017-10-18
stbt.conf: Change default interpress_delay_secs to 0.3

drothlis:
This matches what the documentation says, and it's the best practice to
avoid confusing the device-under-test with infrared signals that are too
close together.


wmanley:
+1.  I think it would be good to credit @rinaldomerlo in the commit message too.

drothlis:
> I think it would be good to credit @rinaldomerlo in the commit message too.

Done (and in the release note).

PR #464, opened 2017-10-25
CEC control: Allow specifying CEC numerical codes directly

drothlis:
This allows you to get to the reserved numbers (maybe some manufacturers
have implemented some of them) or to implement your own string->keycode
mapping.

### DIFF
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -1,3 +1,4 @@
+import re
 from contextlib import contextmanager
 from textwrap import dedent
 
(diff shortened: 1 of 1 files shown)


_stbt/control_gpl.py
```diff
@@ -162,6 +163,14 @@ def press(self, key):
         from .control import UnknownKeyError
         keycode = self._KEYNAMES.get(key)
         if keycode is None:
+            if isinstance(key, int):
+                keycode = key
+            elif re.match(r"^[0-9]+$", key):
+                keycode = int(key, base=10)
+            elif re.match(r"^0[xX][0-9a-fA-F]+$", key):
+                keycode = int(key, base=16)
```
wmanley: Python can do this working out itself.  Perhaps something like:

    try:
        keycode = int(key, 0)
    except ValueError:
        pass

would be better.  But maybe not, `015 == 13`  may be a little confusing.

PR #466, opened 2017-11-02
OpenCV 3 Compatibility

wmanley:
Debian testing now has OpenCV 3 so I need this change to make stbt development convenient on my PC.

This isn't necessarily complete, but it does help to get a bunch of more tests to pass.

TODO:

- [x] Clarify what the changes are for `findContours` between 2 and 3.

Further Work:

- Set up Travis to build a configuration with OpenCV 3 too.  We can't call it supported until we've got regular testing happening. Perhaps this can wait until Ubuntu 18.04 is released?

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -34,6 +34,7 @@ from enum import IntEnum
 from kitchen.text.converters import to_bytes
 from numpy import inf
 
+import _stbt.cv2_compat as cv2_compat
 from _stbt import imgproc_cache, logging, utils
 from _stbt.config import ConfigurationError, get_config
 from _stbt.gst_utils import (array_from_sample, Frame, gst_iterate,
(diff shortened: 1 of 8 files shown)


_stbt/core.py
```diff
@@ -2283,12 +2296,11 @@ def _match_template(image, template, method, roi_mask, level, imwrite):
...
     else:
-        contours, _ = cv2.findContours(
-            roi_mask, cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_NONE)
+        contours = _find_contours(
+            image=roi_mask, mode=cv2.RETR_EXTERNAL,
+            method=cv2.CHAIN_APPROX_NONE)
         rois = [
             _Rect(*cv2.boundingRect(x))
-            # findContours ignores 1-pixel border of the image
-            .shift(Position(-1, -1)).expand(_Size(2, 2))
```
wmanley: TODO: I think I've preserved the `.shift(Position(-1, -1)` with the new implementation by adding `offset=(1, 1)`, but I've thrown away the expand bit.  I'll need to put it back to preserve the same behaviour, but first I need to figure out how it's changed between opencv 2 and 3.

LewisHaley:
https://github.com/opencv/opencv/commit/56c133d is the offending commit that changes the border handling in `findCountours`.

wmanley:
@LewisHaley: It sounds suspiciously like you've done this work before.  It would be much better if you were to push these changes upstream as you do them.  It would have saved me the bother of re-implementing this.

wmanley:
@LewisHaley: Please review.  I think you're the most qualified to tell whether this does what it's supposed to.

LewisHaley:
I've raised https://github.com/stb-tester/stb-tester/pull/467 with our version of OpenCV3 support.

wmanley:
I've diffed this against #467 and can't see any differences that will affect the results.

wmanley:
Inspired by @LewisHaley's implementation I've moved the compatibility code into a separate file.  This is a little bit uglier regarding `_Rect` but I don't think that's a big deal compared to the advantages of putting this in a separate file.

PR #467, opened 2017-11-03
Support opencv3

LewisHaley:


wmanley:
I like the fact that the compatibility code is in a separate file, but I'm not a fan of the shim approach.  In general I like to avoid layers if possible.  The shim approach has meant that there is still a compatibility `if` statement in core.py too.

It's useful to know that there are 3 versions of `findContours` with the return type change in v3.0 and the behavioural change came in v3.2.  I think I'll update #466 to use a separate file as you have and merge that.  I'll check the diff against this PR to make sure I've not missed anything.

Really for this support to be complete we should have a travis configuration that tests against opencv 3.  This should be possible with [Travis's docker support](https://docs.travis-ci.com/user/docker/) and it's [Build Matrix support](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix).  This isn't something I'm intending to do right now though, but if you have an interest in OpenCV 3 support continuing to work you might consider it?

drothlis:
I've merged #466 (@wmanley's implementation) since he'd had to implement it already anyway. Nevertheless seeing your code @LewisHaley was useful so that we could diff against it to ensure we hadn't missed anything, and I do appreciate that you took the time to share it.

PR #468, opened 2017-11-06
Add ability to interrupt other threads

wmanley:
As long as they are blocked on `stbt.frames()` or `time.sleep()`

We add the ability to interrupt a thread that we've previously spawned - as long as they are blocked on `get_frame` or `time.sleep`.  This takes care of tearing threads down once we're done with them without requiring the user to add any complicated logic for this purpose.  This is the invasive part that requires changes to `_stbt.core`

This same mechanism could be used to implement a timeout `contextmanager`.

Example:

    def look_for_logo():
        return wait_until(lambda: match("logo.png"))

    t = threading.Thread(target=look_for_logo)
    t.start()

    interrupt(t)
    t.join()

This is broken out of #463.  Hopefully it should be enough to be able to experiment with threading APIs outside of stbt before we bake something into it.

# TODO

- [ ] Merge #449 
- [ ] Add test
- [ ] Close-up review


drothlis:
Rebased onto master to get Travis fixes from #474.

drothlis:
Somehow this breaks [test_press_visualisation] consistently, both on Travis and my laptop.

[test_press_visualisation]: https://github.com/stb-tester/stb-tester/blob/18e18dd63f7dc226f39d15f009ab3f1af8239a55/tests/test-stbt-py.sh#L339-L372

wmanley:
See also: [Timeouts and cancellation for humans](https://vorpus.org/blog/timeouts-and-cancellation-for-humans/)

PR #469, opened 2017-11-06
[WIP] {detect,wait_for}_motion: Set default noise_threshold to 1.0

drothlis:
This means that any differences will be considered motion. This is more
suitable for lossless video capture.

TODO:

- [ ] What about lossless video capture but watching broadcast TV which will have compression artifacts that change even for a still image? Maybe it's a *good* thing to count that as motion?

### DIFF
--- a/stbt.conf
+++ b/stbt.conf
@@ -52,7 +52,7 @@ interval_secs = 3
 max_presses = 10
 
 [motion]
-noise_threshold=0.84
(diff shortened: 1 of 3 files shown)


drothlis:
Closing. We've seen moire-style artifacts on the Apple TV that are imperceptible to humans but would fail a totally strict threshold — for example see commit 5c8e8c51.

PR #472, opened 2017-11-15
stbt.match debug html: Fix "didn't match" in title

drothlis:
Also add some characterization tests to avoid similar regressions in future.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -2653,6 +2653,7 @@ def _log_match_image_debug(imglog):
         f.write(template.render(
             link=link,
             match_parameters=imglog.data["match_parameters"],
+            matched=any(imglog.data["matches"]),
             matches=imglog.data["matches"],
             min=min,
             pyramid_levels=imglog.data["pyramid_levels"],


drothlis:
Regarding commit 427f5cd (_stbt.core._AnnotatedTemplate.friendly_name: Return relative path):

I considered making this return a path relative to the test-pack root (i.e. the git root) if it's within a git repo, or an absolute path (as before) if not within a git repo. But then we'd be subprocessing out to `git rev-parse --show-toplevel` for every frame in a `wait_until(lambda: stbt.match(...))` loop. The solution to *that* is to memoize `_load_template`. I haven't done that (for expediency) but maybe I really should...

drothlis:
Hmmm the exact output of `cv2.matchTemplate` isn't deterministic across different versions of OpenCV:

```
           <td><a href='match4-heatmap.png'><img src='match4-heatmap.png'></a></td>
           <td></td>
           <td>Matched</td>
-          <td>Region(x=177, y=163, right=312, bottom=207)</td>
+          <td>Region(x=28, y=1, right=163, bottom=45)</td>
           <td><a href='match4-source_with_match.png'><img src='match4-source_with_match.png'></a></td>
-          <td>0.9999</td>
+          <td>0.9998</td>
```

drothlis:
For the sake of merging this fix I've moved the tests to #473.

wmanley:
> I considered making this return a path relative to the test-pack root (i.e. the git root) if it's within a git repo, or an absolute path (as before) if not within a git repo. But then we'd be subprocessing out to `git rev-parse --show-toplevel` for every frame in a `wait_until(lambda: stbt.match(...))` loop.

I think memoizing `_load_template` would be a very good idea, but not for git relative paths.  I think what matters is if the test is in a git repo, so you'd look it up once at the beginning of the test (maybe in stbt-run) possibly storing it in a global (yuck) and then do relpath from then on.  Ideally you wouldn't subprocess either - looking for `.git` recursively upwards would be easy enough to implement

Perhaps a better idea would be to define an environment variable(?) called `STBT_TEST_PACK_ROOT`.  We could set it in stbt-batch and then all the paths would be relative to this?  Or maybe instead of looking for `.git` we should be looking fot `.stbt.conf` insetad

memoizing `_load_template` may have a few subtleties as it currently looks up the stack to find the paths to look in.  You'd need to resolve to a path (or maybe a path, mtime tuple) outside the cache.  After that an application of `functools32.memoize` should be straightforward.

PR #474, opened 2017-12-09
stbt.ocr: Add type checking for tesseract_user_patterns (list)

drothlis:
To avoid the common mistake of passing a string (instead of a list).
Since a string is iterable, this used to accept the string but it would
treat it as a list of patterns of a single character each, which is not
the effect you wanted.

While I'm here I might as well do the same check for
tesseract_user_words.

Note that stbt.match_text doesn't take either of these parameters so I
don't need to touch that.


drothlis:
I could also make it accept a string and convert it to a list with a single string. That would be more "Do What I Mean" but less explicit -- and Python is supposed to be all "Explicit is better than implicit" so I chose to raise a TypeError.

_stbt/core.py
```diff
@@ -1154,6 +1154,16 @@ def ocr(self, frame=None, region=Region.ALL,
...
 
+        if tesseract_user_words is not None and not isinstance(
+                tesseract_user_words, list):
+            raise TypeError(
+                "stbt.ocr: tesseract_user_words must be a list of strings")
+
+        if tesseract_user_patterns is not None and not isinstance(
+                tesseract_user_patterns, list):
+            raise TypeError(
+                "stbt.ocr: tesseract_user_patterns must be a list of strings")
```
wmanley: I think being able to pass in a set or a tuple is useful.  Perhaps specifically blacklisting strings would be better rather than whitelisting lists?

wmanley:
> I could also make it accept a string and convert it to a list with a single string. That would be more "Do What I Mean" but less explicit -- and Python is supposed to be all "Explicit is better than implicit" so I chose to raise a `TypeError`.

There is prior art in the Python standard library for this behaviour: `subprocess.Popen` converts strings: https://github.com/python/cpython/blob/2.7/Lib/subprocess.py#L908-L911

drothlis:
OK, I've made it accept a string.

wmanley:
I think this is worth a test.  Should be easy enough.

wmanley:
And an update to the documentation

drothlis:
CI failure is:

>     _stbt/control_gpl.py:121:8: [E0401(import-error), HdmiCecControl.__init__] Unable to import 'cec'

Probably caused by new Travis images deployed yesterday:
https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch

drothlis:
> I think this is worth a test.

We don't really have a way to test this. See 8b47c35.

drothlis:
I suppose we do have a way to test it after all -- I've found an image that does require `tesseract_user_{words,patterns}` with Ubuntu 16.04's version of tesseract.

drothlis:
I've added tests & updated the stbt.ocr docstring.

I fixed the Travis failure by telling it to use the previous version of the Trusty images, giving us some time to investigate.

PR #478, opened 2018-01-09
Python API: Add stbt.crop to crop a region from a video-frame 

drothlis:


_stbt/core.py
```diff
@@ -585,6 +585,23 @@ def load_image(filename, flags=cv2.IMREAD_COLOR):
...
+    :type frame: `stbt.Frame` or `numpy.ndarray`
+    :param frame: An image in OpenCV format (for example as returned by
+      `frames`, `get_frame` and `load_image`, or the ``frame`` parameter of
+      `MatchResult`).
+
+    :type Region region: The region to crop.
+
+    :returns: An OpenCV image (`numpy.ndarray`) containing the specified region
+      of the source frame. This is a copy, so you can change the data in the
+      image without changing the original video frame.
```
wmanley: Why copy?

_stbt/core.py
```diff
@@ -585,6 +585,23 @@ def load_image(filename, flags=cv2.IMREAD_COLOR):
...
+    """
+    return _crop(frame, region).copy()
+
+
 def _crop(frame, region):
     if not _image_region(frame).contains(region):
         raise ValueError("'frame' doesn't contain 'region'")
```
wmanley: Maybe spruce up this exception to give the dimensions of the frame and the region

_stbt/core.py
drothlis: I thought it would be less confusing. But given that frames are read-only now, there isn't much chance for confusion. I've removed the copy.

_stbt/core.py

PR #479, opened 2018-01-09
stbt.ocr: Add text_color_threshold parameter

drothlis:
~~I've wanted this in the past, though now I don't remember the exact use case so I don't have any images to use for a self-test.~~

I have added a testcase (see [comment](#issuecomment-378564281) below).


stbt/__init__.py
```diff
@@ -408,7 +408,9 @@ def ocr(frame=None, region=Region.ALL,
     :param text_color:
         Color of the text. Specifying this can improve OCR results when
         tesseract's default thresholding algorithm doesn't detect the text,
-        for example for white text on a light-colored background.
+        for example for white text on a light-colored background. The threshold
+        can be adjusted further by passing a 4th number in the tuple, also
+        between 0 and 255.
```
wmanley: I'm not sure of this as an API.  Perhaps a `text_color_threshold` would be better.  I've been thinking of adding a `Color` class much like we have `Region` and I don't see how it would fit in here.

wmanley:
> I don't remember the exact use case so I don't have any images to use for a self-test.

I think we'll have to wait until we've got some examples where this helps before merging it.

drothlis:
I've rebased onto master and added a testcase that currently fails at the default text_color threshold of 25.

Source image:
blue-search-white-guide

text_color | threshold=25 | threshold=50
--- | --- | ---
(200,200,200) |  Reads `'GUidG'` |
(220,220,220) |  Reads `'GUIde'` |  Reads `'Guide'`
(240,240,240) |  Reads `"(Lamb '*"` |

drothlis:
Rebased onto master, and changed to use a separate `text_color_threshold` parameter instead of shoving the threshold into the `text_color` tuple.

_stbt/core.py
```diff
@@ -1220,7 +1220,8 @@ def wait_for_motion(
     def ocr(self, frame=None, region=Region.ALL,
             mode=OcrMode.PAGE_SEGMENTATION_WITHOUT_OSD,
             lang=None, tesseract_config=None, tesseract_user_words=None,
-            tesseract_user_patterns=None, upsample=True, text_color=None):
+            tesseract_user_patterns=None, upsample=True, text_color=None,
+            text_color_threshold=25):
```
wmanley: If we're thinking that a different default might be a good idea perhaps this should be `=None` and we can look up the default with `get_config`?

drothlis: Good idea.

PR #480, opened 2018-01-09
stbt.wait_for_motion/detect_motion/is_screen_black: Masks can be a numpy image

drothlis:
This allows you to construct a mask programmatically, for example using `numpy.zeros` and `numpy.ones`. In future we may add helper functions like "mask_in" and "mask_out".


wmanley:
These functions would really benefit from having a `region` parameter too.  Dealing with regions is faster than masks.

wmanley:
Yep, I can't find anything wrong with this.  I'll be interested to see where this functionality actually comes in useful.

PR #482, opened 2018-01-11
stbt.ocr & stbt.match_text: Add upsample parameter to disable upsampling

drothlis:
Stb-tester upsamples the image before passing it to tesseract (since v0.20 in 2014). For details see https://stb-tester.com/blog/2014/04/14/improving-ocr-accuracy
    
With this parameter you can *disable* that behaviour. This is useful if you want to do your own pre-processing before calling `stbt.ocr`. For example this would have made it possible to implement `text_color` in your test-pack without modifying stb-tester.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1207,7 +1207,7 @@ class DeviceUnderTest(object):
         rts = getattr(frame, "time", None)
 
         xml, region = _tesseract(frame, region, mode, lang, _config,
-                                 user_words=text.split(), text_color=text_color)
+                                 None, text.split(), text_color)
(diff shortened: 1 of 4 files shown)


wmanley:
There's also:

https://github.com/stb-tester/stb-tester/blob/c9c3261e6a437c94d322f6214bbea3e4f1c19491/_stbt/core.py#L1214-L1218

drothlis:
Ah yes. `region` is from the un-upsampled coordinates, but I missed that `box` is from the upsampled coordinates.

Added a test & fixed.

PR #483, opened 2018-01-12
stbt.FrameObject: `is_visible` can call other public properties

drothlis:
Previously `is_visible` could only call private properties of the
FrameObject. If you tried to call a *public* property from `is_visible`,
you would get "RuntimeError: maximum recursion depth exceeded". This was
a side-effect of the FrameObject design that public properties should
return `None` if the FrameObject isn't visible. To work around this you
had to add a private property that was used by both `is_visible` and the
corresponding public property.

Now each FrameObject instance knows if it's inside a call to
`is_visible`, so that you can call public properties.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1636,13 +1636,12 @@ def _memoize_property_fn(fn):
 def _mark_in_is_visible(fn):
     @functools.wraps(fn)
     def inner(self):
-        # This increment isn't threadsafe but I can't imagine using the same
-        # FrameObject instance in multiple threads.
-        self._FrameObject__in_is_visible += 1
+        # pylint: disable=protected-access
+        self._FrameObject__local.in_is_visible += 1
         try:
             return fn(self)
         finally:
-            self._FrameObject__in_is_visible -= 1
+            self._FrameObject__local.in_is_visible -= 1
     return inner
 
 
@@ -1650,7 +1649,7 @@ def _noneify_property_fn(fn):
     @functools.wraps(fn)
     def inner(self):
         # pylint: disable=protected-access
-        if self._FrameObject__in_is_visible or self.is_visible:
+        if self._FrameObject__local.in_is_visible or self.is_visible:
             return fn(self)
         else:
             return None
@@ -1700,7 +1699,8 @@ class FrameObject(object):
         if frame is None:
             raise ValueError("FrameObject: frame must not be None")
         self.__frame_object_cache = {}
-        self.__in_is_visible = 0
+        self.__local = threading.local()
+        self.__local.in_is_visible = 0
         self._frame = frame
 
     def __repr__(self):
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1639,7 +1639,7 @@ def _mark_in_is_visible(fn):
         # pylint: disable=protected-access
         self._FrameObject__local.in_is_visible += 1
         try:
-            return fn(self)
+            return bool(fn(self))
         finally:
             self._FrameObject__local.in_is_visible -= 1
     return inner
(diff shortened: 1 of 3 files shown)


drothlis:
This behaviour has potential to be surprising if a private property calls a public property. If the private property is used by `is_visible` it will behave one way; if it isn't called by `is_visible` it will behave a different way. Seems unlikely to matter in practice, and this behaviour is far more useful for the common case.

_stbt/core.py
```diff
@@ -1633,10 +1633,24 @@ def inner(self):
...
+def _mark_in_is_visible(fn):
+    @functools.wraps(fn)
+    def inner(self):
+        # This increment isn't threadsafe but I can't imagine using the same
+        # FrameObject instance in multiple threads.
+        self._FrameObject__in_is_visible += 1
+        try:
+            return fn(self)
+        finally:
+            self._FrameObject__in_is_visible -= 1
```
wmanley: This isn't threadsafe.  If thread A is calling `.is_visible` and thread B is calling `.text` then thread B may get a result back other than `None`.  Making this threadsafe while preserving the lazy evaluation of `is_visible` is going to be rather tricky. You could make this counter thread local, or add locking - both of which might affect performance, and would definitely make the code more complex.

I think it would be better to change the semantics of `FrameObject` and make `is_visible` eagerly evaluated.  We can call it in the constructor so we have the guarantee that `is_visible` will be called before any other method or property.  That way we won't need locking, we can just check a simple flag.

Hopefully there aren't too many instances in the wild where we create `FrameObject` instances but don't ever check any of the properties on them.

drothlis:
I've fixed the race condition by using thread-local storage for the `in_is_visible` counter.

Evaluating `is_visible` eagerly in the constructor means that if you define members in your subclass's constructor, and those are used by `is_visible`, you have to define them before you call the `super().__init__()` -- which is counter-intuitive. I think one less surprise for the user is worth the additional implementation complexity.

drothlis:
To execute `is_visible` eagerly in the constructor I'd need to use something like this:

```
 1    def __init__(self, frame):
 2        if frame is None:
 3            raise ValueError("FrameObject: frame must not be None")
 4        self.__frame_object_cache = {}
 5        self._frame = frame
 6
 7        self.__in_is_visible = True
 8        value = bool(self.is_visible)
 9        self.is_visible = value  # fails with AttributeError: can't set attribute
10        self.__in_is_visible = False
```

Line 9 isn't necessary for caching the return value, because our `_memoize_property_fn` decorator already does that. It (line 9) is only necessary if we want to implement the functionality "is_visible is always a bool". However it fails because you can't replace a property with a plain value, so I don't know how to implement that.

To implement "is_visible is always a bool" we could use a *different* decorator for the `is_visible` property, but then the implementation complexity is similar to the solution as currently implemented on this branch (at least measured by number of decorators). At least it wouldn't need thread-local storage. I've implemented this in 46feb16 on branch `is-visible-properties-2` (which is branched from this branch). @wmanley it's up to you.

wmanley:
I've made some modifications to your [is-visible-properties-2](https://github.com/stb-tester/stb-tester/commits/is-visible-properties-2) hopefully to simplify.  See [is-visible-properties-3](https://github.com/stb-tester/stb-tester/commits/is-visible-properties-3).  I think it's a bit more straightforward.

wmanley:
I think I'll merge this as is.  The costs of having a `threading.local()` per `FrameObject` instance isn't clear to me but we could always change this in the future.  Perhaps we could use a module level `set` instead if it turns out to be a problem, or move to eager evaluation then.

wmanley:
> However it fails because you can't replace a property with a plain value, so I don't know how to implement that.

The way I did it was to rename `is_visible` in the metaclass.

PR #487, opened 2018-02-08
Python API: Add `stbt.OcrPatterns.IP_ADDRESS`

wmanley:
The OCR system that stb-tester uses was trained against long-form text so
it expects to read normal words.  This can confuse it when it comes across
text that aren't words - like IP addresses.

In these instances you can use `tesseract_user_patterns` to tell tesseract
to expect a string of a certain form, but it can be unclear what to pass to
it.  Now you can write:

    stbt.ocr(tesseract_user_patterns=stbt.OcrPatterns.IP_ADDRESS)

This tells tesseract to expect something of the form `x.x.x.x` where x is
a number fixing the misread.

TODO:

- [ ] Release note


_stbt/core.py
```diff
@@ -689,6 +689,18 @@ def __repr__(self):
...
 
 
+class OcrPatterns(object):
+    """
+    Common tesseract_user_patterns arguments you might want to pass when calling
+    ocr.
+
+    >>> import stbt
+    >>> stbt.ocr(frame=load_image("tests/ip_address.png"))
+    u'192.168.1.64'
```
drothlis: The doctest doesn't use OcrPatterns / tesseract_user_patterns!

_stbt/core.py
```diff
@@ -689,6 +689,18 @@ def __repr__(self):
...
+    """
+    IP_ADDRESS = (r"\d\*.\d\*.\d\*.\d\*",)
```
drothlis: If this was just a string (not a tuple with 1 string) it would be easier to combine with other patterns, e.g. 

    tesseract_user_patterns=[OcrPatterns.IP_ADDRESS, OcrPatterns.SOMETHING_ELSE]

Since #474 it's safe to pass a single string if you only have 1 pattern.

_stbt/core.py
drothlis: The doctest code is potentially confusing to people who aren't expecting to pass in a frame explicitly? Might be better suited as a test in `tests/test_ocr.py`.

drothlis:
Rebased onto master, added a testcase where this PR's original pattern (`r"\d\*.\d\*.\d\*.\d\*"`) isn't sufficient, and changed the patterns to make the testcase pass.

Tested on tesseract 3.04.01 (Ubuntu 16.04) and 3.03 (Ubuntu 14.04). In both cases the old pattern wasn't sufficient.

wmanley:
What did it misread as?

drothlis:
192.168.1231

wmanley:
It would be nice to have a mechanism to set `tessedit_whitelist_chars` at the same time.

drothlis:
Which is funny because `.` is one of the characters that is unambiguously present in the pattern (it isn't clear to me if `\*` means one-or-more or zero-or-more). Yet that is the one that it failed to read.

drothlis:
Maybe it's because of this (from tesseract's trie.h):

```
  // Note: In choosing which patterns to include please be aware of the fact
  // providing very generic patterns will make tesseract run slower.
  // For example \n\* at the beginning of the pattern will make Tesseract
  // consider all the combinations of proposed character choices for each
  // of the segmentations, which will be unacceptably slow.
  // Because of potential problems with speed that could be difficult to
  // identify, each user pattern has to have at least kSaneNumConcreteChars
  // concrete characters from the unicharset at the beginning.
```

...so the `\d\*` pattern was being ignored? No, can't be because using that pattern fixes the ocr result for `ocr/192.168.10.1.png`.

drothlis:
Tesseract 4.00~git2288-10f4998a-2 from Ubuntu 18.04 reads it OK, unaided by patterns or up-scaling.

PR #490, opened 2018-03-08
Improvements to running from srcdir

wmanley:


drothlis:
I've squashed and pushed some fixups:

```
diff --git a/bin/stbt b/bin/stbt
index 96feb41..9f05d0c 100755
--- a/bin/stbt
+++ b/bin/stbt
@@ -56,11 +56,11 @@ check_experimental() {
 exec_stbt() {
     IFS=':'
     for x in $STBT_EXEC_PATH; do
-        if [ -e "$x/$1" ]; then
+        if [ -x "$x/$1" ]; then
             exec "$x/$@"
         fi
     done
-    printf "Couldn't find $1 in PATH $STBT_EXEC_PATH" 2>&1
+    echo "Couldn't find $1 in PATH $STBT_EXEC_PATH" 1>&2
     return 1
 }
 
@@ -76,9 +76,9 @@ case "$cmd" in
     templatematch)  # for backwards compatibility
         exec_stbt stbt-match "$@";;
     virtual-stb)
-        export PYTHONPATH=$PYTHONPATH:$LIBEXECDIR/stbt/
         exec_stbt stbt_virtual_stb.py "$@"
-        echo "stbt virtual-stb is not installed." >&2;;
+        echo "stbt virtual-stb is not installed." >&2
+        exit 1;;
     # Experimental sub-commands:
     camera)
         check_experimental
```

PR #491, opened 2018-03-09
stbt.press: Add adb control

drothlis:
We already had `stbt.android.AdbDevice.press` but you had to create an instance and use it explicitly, rather than calling `stbt.press` like you do for the other control mechanisms.

TODO:

- [x] Integrate `stbt.android` with `stbt.press`.
- [x] Man page.
- [x] Release notes.
- [x] Map standard stb-tester names (`KEY_xxx`) to [Android KeyEvents](https://developer.android.com/reference/android/view/KeyEvent.html).


wmanley:
Good new feature and nice tidy-up too

PR #492, opened 2018-03-16
FrameObject: Fix AttributeError with multiple threads

wmanley:
See 78478e02a1 and the discussion on #483 for the background behind this
thread-local.

Traceback:

      ...
      File ".../stb-tester/_stbt/core.py", line 1727, in __repr__
        args = ", ".join(("%s=%r" % x) for x in self._iter_attrs())
      File ".../stb-tester/_stbt/core.py", line 1727, in <genexpr>
        args = ", ".join(("%s=%r" % x) for x in self._iter_attrs())
      File ".../stb-tester/_stbt/core.py", line 1731, in _iter_attrs
        if self:
      File ".../stb-tester/_stbt/core.py", line 1739, in __nonzero__
        return bool(self.is_visible)
      File ".../stb-tester/_stbt/core.py", line 1660, in inner
        self._FrameObject__local.in_is_visible += 1
    AttributeError: 'thread._local' object has no attribute 'in_is_visible'

TODO:

- [x] Add a test

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1657,7 +1657,10 @@ def _mark_in_is_visible(fn):
     @functools.wraps(fn)
     def inner(self):
         # pylint: disable=protected-access
-        self._FrameObject__local.in_is_visible += 1
+        try:
+            self._FrameObject__local.in_is_visible += 1
+        except AttributeError:
+            self._FrameObject__local.in_is_visible = 1
         try:
             return bool(fn(self))
         finally:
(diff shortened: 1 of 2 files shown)


drothlis:
I don't understand how the existing test `test_that_is_visible_and_properties_arent_racy` could have worked?

wmanley:
> I don't understand how the existing test test_that_is_visible_and_properties_arent_racy could have worked?

It was collecting the results from each thread in a `dict` and then checking that "all" the results in the dict were fine.  Each thread would fail before writing to the dict so the results dict was empty at the end of the tests.  As a result "all" of the results were correct - there were no results.

I've fixed the test.

PR #494, opened 2018-04-04
NoVideo is a test *error*, not test *failure*

drothlis:
Most video-capture hardware we have seen will continue to deliver video
(usually black frames) when the device-under-test is powered off. In
this case NoVideo is an error in your test infrastructure, not of your
device-under-test.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1545,7 +1545,7 @@ class UITestFailure(Exception):
     pass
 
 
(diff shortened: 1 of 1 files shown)


drothlis:
According to d12b429b177fc474bf0ac26db94ba7335223aaa8 the only video-capture hardware we're aware of that stops delivering frames in the Hauppauge HDPVR which is tremendously obsolete.

PR #495, opened 2018-04-16
Pythonize stbt batch run

wmanley:
For better maintainability and speed.  With this change we will be spawning many fewer processes.  In particular all the (fairly slow) calls to `stbt config` from bash are removed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,6 +90,7 @@ INSTALL_CORE_FILES = \
     stbt-batch \
     stbt-batch.d/instaweb \
     stbt-batch.d/post-run.sh \
+    stbt-batch.d/print_backtrace.gdb \
     stbt-batch.d/report \
     stbt-batch.d/report.py \
     stbt-batch.d/run.py \
--- a/Makefile
+++ b/Makefile
@@ -89,10 +89,10 @@ INSTALL_CORE_FILES = \
     stbt_auto_selftest.py \
     stbt-batch \
     stbt-batch.d/instaweb \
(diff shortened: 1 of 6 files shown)


drothlis:
`ts` is still used once in `check_capture_hardware` here: https://github.com/wmanley/stb-tester/blob/a237a2f4cc851cfe453e0bc0bd2f73cbef526476/stbt-batch.d/post-run.sh#L41-L45.
Otherwise we could have removed it from the apt package's dependencies.

drothlis:
In future we could also remove `template` here:
https://github.com/wmanley/stb-tester/blob/a237a2f4cc851cfe453e0bc0bd2f73cbef526476/stbt-batch.d/post-run.sh#L24-L29

...And move it into stbt-run itself (`stbt.match` et al. can record the template used for the last match operation).

drothlis:
I've raised #496 with the fixes.

drothlis:
To see the effect of using Python instead of `ts` I plotted the timestamps in the log file for a test-script that does `wait_for_match` on the following source pipeline: 

    videotestsrc is-live=true ! video/x-raw,width=1280,height=720,framerate=25/1

so we should be getting a line logged every 40ms. This plot shows the offset of the timestamp added by our tee/logger, i.e. mod 40,000μs. "Python" (blue) is #496 (i.e. this pull request + fixes), "ts" (red) is 990778e (i.e. master before this pull request).

screenshot from 2018-04-16 16-25-30

Looks close enough to me.

PR #498, opened 2018-04-24
Install `_stbt` library into global python directory

wmanley:
It means less fiddling with `$PYTHONPATH`.  It's called `_stbt` so that should be enough to prevent people from using it.


Makefile
```diff
@@ -86,6 +81,7 @@ INSTALL_CORE_FILES = \
     _stbt/x11.py \
     _stbt/xorg.conf.in \
     _stbt/xxhash.py \
+    stbt.conf \
```
drothlis: I see, so now we install `stbt.conf` to `/etc/stbt/stbt.conf` *and* to `$libexecdir/stbt/stbt.conf` (the latter location used to be a modified version of `stbt.conf`, now it's just the identical file).

stbt/__init__.py
```diff
@@ -15,8 +15,8 @@
 from contextlib import contextmanager
 
 try:
-    from stbt._vars import _libexecdir  # Generated by 'make install'. pylint:disable=import-error,no-name-in-module
-    sys.path.insert(0, _libexecdir + "/stbt")
+    from _stbt.vars import libexecdir  # Generated by 'make install'. pylint:disable=import-error,no-name-in-module
```
drothlis: This won't be able to find `_stbt` but that's fixed by the next commit.

Makefile
```diff
@@ -135,6 +135,10 @@ install-core: all
 	done
 	printf "libexecdir = '%s'\nsysconfdir = '%s'\n" \
 	    "$(libexecdir)" "$(sysconfdir)" > $(DESTDIR)$(pythondir)/_stbt/vars.py
+	for filename in $(INSTALL_PYLIB_FILES); do \
+	    [ -x "$$filename" ] && mode=0755 || mode=0644; \
+	    $(INSTALL) -m $$mode $$filename $(DESTDIR)$(pythondir)/$$filename; \
+	done
 	chmod 0644 $(DESTDIR)$(pythondir)/_stbt/vars.py
```
drothlis: The new code needs to go 1 line lower -- you've split up the code that deals with `_stbt/vars.py`.

Makefile
```diff
@@ -135,6 +135,10 @@ install-core: all
 	done
 	printf "libexecdir = '%s'\nsysconfdir = '%s'\n" \
 	    "$(libexecdir)" "$(sysconfdir)" > $(DESTDIR)$(pythondir)/_stbt/vars.py
```
drothlis: Do we still need to record `libexecdir`? Yes we do, for `_stbt/control.py::stbt_control_listen`, which imports `$libexecdir/stbt/stbt-control`, and which is used by `stbt record --control-recorder`.

wmanley: The real fix there is to move (at least some of) the contents of `stbt-control` into a library under `_stbt`.  I couldn't be bothered to untangle it right now though.

Makefile
wmanley: The ideal would be to install `/etc/stbt/stbt.conf` as identical to `/usr/.../stbt.conf` but with all the variables commented out.  This way the options and their defaults are documented and it's clear which ones are being overridden.

PR #499, opened 2018-04-25
Make stbt-run headless by default

wmanley:
`stbt run` will no longer show an output video window by default. This is a
better default for headless environments like stbt-docker.  You can re-enable
this by setting `global.sink_pipeline = xvimagesink sync=false` in your
`$HOME/.config/stbt/stbt.conf`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ INSTALL_CORE_FILES = \
     stbt-screenshot \
     stbt-tv
 
-all: $(INSTALL_CORE_FILES) $(INSTALL_PYLIB_FILES)
+all: $(INSTALL_CORE_FILES) $(INSTALL_PYLIB_FILES) etc/stbt.conf
 
 INSTALL_VSTB_FILES = \
     stbt_virtual_stb.py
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -160,6 +162,11 @@ install-gpl: $(INSTALL_GPL_FILES)
 	    $(INSTALL) -m $$mode $$filename $(DESTDIR)$(pythondir)/$$filename; \
 	done
 
+etc/stbt.conf : _stbt/stbt.conf
+	# Comment out defaults in /etc/stbt.conf
```
drothlis: `s,/etc,/etc/stbt,`

_stbt/stbt.conf
```diff
@@ -1,6 +1,10 @@
 [global]
 source_pipeline=videotestsrc is-live=true
-sink_pipeline=xvimagesink sync=false
+sink_pipeline=
```
drothlis: Isn't this missing `fakesink sync=false`?

wmanley: Not since https://github.com/stb-tester/stb-tester/pull/420/commits/5765354927b052e01cfd8379140081924689caba

Makefile
wmanley: Fixed

_stbt/stbt.conf
drothlis: ...plus 91ccfa1.

PR #500, opened 2018-04-25
Fix timing of stbt.press visualisation on the output video

drothlis:
The text would be drawn immediately on the video frame currently being
processed, not taking into account the latency in the video pipeline, so
it would appear a fraction of a second too early in the video.

The difference is quite obvious using videotestsrc + `--control=test` +
a script that does `sleep(2); stbt.press("snow")`. I tried to write an
automated test but it was too flaky.


_stbt/core.py
```diff
@@ -1913,8 +1922,7 @@ def _push_sample(self, sample):
         annotations = []
         with self.annotations_lock:
             for x in self.text_annotations:
-                x.setdefault('start_time', now)
-                if now < x['start_time'] + x['duration']:
+                if now < x.end_time:
                     texts.append(x)
             self.text_annotations = texts[:]
```
wmanley: Now we don't have setdefault this could be "simplified" to a list comprehension:

    # Remove expired annotations
    self.text_annotations = [x for x in self.text_annotations if now < x.end_time]
    current_annotations = [x for x in self.text_annotations if x.time <= now]

Not important

_stbt/core.py
```diff
@@ -1945,11 +1954,14 @@ def _push_sample(self, sample):
...
-                    datetime.datetime.now().strftime("%H:%M:%S.%f")[:-4] +
+                start_time = self._time.time()
+                text = (
+                    datetime.datetime.fromtimestamp(start_time).strftime(
+                        "%H:%M:%S.%f")[:-4] +
                     ' ' + obj)
-                self.text_annotations.append(
-                    {"text": obj, "duration": duration_secs})
+                self.text_annotations.append(_TextAnnotation(
+                    start_time or self._time.time(),
```
wmanley: Don't need `or self._time.time()` here.

_stbt/core.py
```diff
@@ -1929,11 +1937,12 @@ def _push_sample(self, sample):
...
             (10, 30), (255, 255, 255))
-        for i, x in enumerate(reversed(texts)):
+        current_texts = [x for x in texts if x.time <= now]
+        for i, x in enumerate(reversed(current_texts)):
             origin = (10, (i + 2) * 30)
-            age = float(now - x['start_time']) / 3
+            age = float(now - x.time) / 3
             color = (int(255 * max([1 - age, 0.5])),) * 3
-            _draw_text(img, x['text'], origin, color)
+            _draw_text(img, x.text, origin, color)
```
wmanley: Yes, this is much nicer with `namedtuple`

_stbt/core.py
drothlis: Good idea, done.

_stbt/core.py

_stbt/core.py
```diff
@@ -1806,14 +1806,23 @@ def draw(self, img):
...
     def __init__(self, user_sink_pipeline, raise_in_user_thread, save_video=""):
+        import time as _time
+
         self.annotations_lock = threading.Lock()
         self.text_annotations = []
         self.annotations = []
         self._raise_in_user_thread = raise_in_user_thread
         self.received_eos = threading.Event()
         self._frames = deque(maxlen=35)
+        self._time = _time
```
wmanley: I guess you'd originally intended to use this in tests?  Doesn't look necessary any more.

PR #504, opened 2018-05-14
Add stbt.press(hold_secs) parameter, and stbt.pressing context manager

drothlis:
Allows holding down a key on the remote-control for a certain duration, or (for the `stbt.pressing` context manager) while a block of code executes.

Only implemented for LIRC, Roku, and CEC controls. I haven't implemented the other controls because I don't know how, or it's impossible. I specifically know that the IRNetBox doesn't have separate keydown/keyup commands.

[lircd(8)]: http://www.lirc.org/html/lircd.html

---

**TODO**

- [x] Test lircd repeat behaviour with irsend.
- [x] Unit test for exception handling behaviour in the `pressing` context manager?
- [x] Implement for LIRC.
- [x] Implement for CEC.
- [x] Implement for stbt-control-relay.
- [x] Test stbt-run with RedRat3 receiver.
- [x] Compare hold signal vs. real remote control.
- [x] Base class for remote controls.
- [x] Lower CEC sleep to 300ms and calculate a deadline for intervals rather than just sleeping for a set amount of time.
- [x] Test hold with IR: Roku.
- [x] Test hold with RokuHttpControl
    - Does it have a time limit after which it considers the key released? No, I tested keydown "Down" and after 4 minutes it's still happily scrolling down the menu.
- [ ] Test hold with CEC: PS4
- [x] Test hold with CEC: FireTV.

### DIFF
--- a/_stbt/control.py
+++ b/_stbt/control.py
@@ -428,7 +428,7 @@ class IRNetBoxRemote(object):
 class RokuHttpControl(object):
     """Send a key-press via Roku remote control protocol.
 
-    See https://sdkdocs.roku.com/display/sdkdoc/External+Control+Guide
+    See https://sdkdocs.roku.com/display/sdkdoc/External+Control+API
     """
 
     # Map our recommended keynames (from linux input-event-codes.h) to the
--- a/_stbt/control.py
+++ b/_stbt/control.py
@@ -14,7 +14,7 @@ from . import irnetbox, utils
 from .config import ConfigurationError
 from .logging import debug, scoped_debug_level
 
-__all__ = ['uri_to_remote', 'uri_to_remote_recorder']
+__all__ = ['uri_to_control', 'uri_to_control_recorder']
 
 try:
     from .control_gpl import controls as gpl_controls
@@ -26,8 +26,8 @@ class UnknownKeyError(Exception):
     pass
 
 
-def uri_to_remote(uri, display=None):
-    remotes = [
+def uri_to_control(uri, display=None):
+    controls = [
         (r'adb(:(?P<address>.*))?', new_adb_device),
         (r'error(:(?P<message>.*))?', ErrorControl),
         (r'file(:(?P<filename>[^,]+))?', FileControl),
@@ -35,40 +35,40 @@ def uri_to_remote(uri, display=None):
              (?P<hostname>[^:]+)
              (:(?P<port>\d+))?
              :(?P<output>\d+)
-             :(?P<config>[^:]+)''', IRNetBoxRemote),
+             :(?P<config>[^:]+)''', IRNetBoxControl),
         (r'lirc(:(?P<hostname>[^:/]+))?:(?P<port>\d+):(?P<control_name>.+)',
-         new_tcp_lirc_remote),
+         new_tcp_lirc_control),
         (r'lirc:(?P<lircd_socket>[^:]+)?:(?P<control_name>.+)',
-         new_local_lirc_remote),
-        (r'none', NullRemote),
+         new_local_lirc_control),
+        (r'none', NullControl),
         (r'roku:(?P<hostname>[^:]+)', RokuHttpControl),
         (r'samsung:(?P<hostname>[^:/]+)(:(?P<port>\d+))?',
-         _new_samsung_tcp_remote),
+         _new_samsung_tcp_control),
         (r'test', lambda: VideoTestSrcControl(display)),
-        (r'x11:(?P<display>[^,]+)?(,(?P<mapping>.+)?)?', _X11Remote),
+        (r'x11:(?P<display>[^,]+)?(,(?P<mapping>.+)?)?', _X11Control),
         (r'rfb:(?P<hostname>[^:/]+)(:(?P<port>\d+))?', RemoteFrameBuffer),
     ]
     if gpl_controls is not None:
-        remotes += gpl_controls
+        controls += gpl_controls
 
-    for regex, factory in remotes:
+    for regex, factory in controls:
         m = re.match(regex, uri, re.VERBOSE | re.IGNORECASE)
         if m:
             return factory(**m.groupdict())
     raise ConfigurationError('Invalid remote control URI: "%s"' % uri)
 
 
-def uri_to_remote_recorder(uri):
-    remotes = [
-        ('file://(?P<filename>.+)', file_remote_recorder),
+def uri_to_control_recorder(uri):
+    controls = [
+        ('file://(?P<filename>.+)', file_control_recorder),
         (r'lirc(:(?P<hostname>[^:/]+))?:(?P<port>\d+):(?P<control_name>.+)',
-         lirc_remote_listen_tcp),
+         lirc_control_listen_tcp),
         (r'lirc:(?P<lircd_socket>[^:]+)?:(?P<control_name>.+)',
-         lirc_remote_listen),
+         lirc_control_listen),
         (r'stbt-control(:(?P<keymap_file>.+))?', stbt_control_listen),
     ]
 
-    for regex, factory in remotes:
+    for regex, factory in controls:
         m = re.match(regex, uri)
         if m:
             return factory(**m.groupdict())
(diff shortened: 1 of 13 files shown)


_stbt/control.py
```diff
@@ -175,20 +204,35 @@ def __init__(self, control_name, connect_fn):
...
+        s.sendall(command + "\n")
+        _read_lircd_reply(s, command)
+        debug("Holding " + key)
+
+    def keyup(self, key):
+        s = self._connect()
+        command = "SEND_STOP %s %s" % (self.control_name, key)
+        s.sendall(command + "\n")
+        _read_lircd_reply(s, command)
+        debug("Released " + key)
```
drothlis: Commit 5f040b7 from #73 said that we needed to keep the socket open until the keyup, but I don't see how that can be true considering that `irsend SEND_START ... ; irsend SEND_STOP ...` from the command-line works correctly, and it is using 2 entirely separate processes for the start & stop.

_stbt/core.py
```diff
@@ -878,7 +878,49 @@ def __exit__(self, exc_type, exc_value, tb):
...
+
+        def __exit__(self, original_exc_type, exc_value, exc_traceback):
+            try:
+                self._control.keyup(self.key)
+                self.draw_text("Released %s" % self.key, duration_secs=3)
+            except Exception:  # pylint:disable=broad-except
+                # Raise an exception if we fail to release the key, but
+                # not if it would mask an exception from the test script.
+                if original_exc_type is not None:
+                    raise
```
drothlis: This test is backwards.

drothlis:
LIRC does send the repeat signal. Tested with:

* [XBox_One.lircd.conf.txt](https://github.com/stb-tester/stb-tester/files/2004436/XBox_One.lircd.conf.txt), which specifies a repeat signal
* IR blaster pointing at a RedRat3 receiver
* Plotting the receiver with `xmode2`

xbox-one

It's clearly sending the repeat signal (the 2nd and subsequent rows) instead of the original signal.

wmanley:
Python has a `threading.Timer` class already, but that's only for one-shot timers.  I've had a crack at implementing an `IntervalTimer` class for better separation of concerns:

```
import logging
import sys
import threading
import time


class IntervalTimer(object):
    """
    Modelled after threading.Timer but running the function at an interval and
    supporting cancellation and can be stopped and started multiple times.

    Improvements vs threading.Timer:

    * Can call start() after calling stop()
    * Exceptions thrown in the thread are propogated to the caller when you call
      stop.
    * It can be used as a contextmanager.
    """
    def __init__(self, interval, target, args=None, kwargs=None, name=None):
        self.interval = interval
        self.target = target
        self.args = args or ()
        self.kwargs = kwargs or {}

        self._thread = None
        self._thread_name = name

        self._lock = threading.Condition()
        self._stop = False
        self._exception = (None, None, None)
        self._first_time = 0

    class AlreadyRunning(Exception):
        pass

    class AlreadyStopped(Exception):
        pass

    @property
    def is_alive(self):
        with self._lock:
            return self._thread is not None

    def start(self):
        with self._lock:
            if self._thread is not None:
                raise IntervalTimer.AlreadyRunning(
                    "IntervalTimer is already running!")
            self._first_time = time.time() + self.interval
            self._thread = threading.Thread(
                target=self._run, name=self._thread_name)
            self._thread.daemon = True
        self._thread.start()

    def stop(self, raise_=True):
        """
        Guaranteed not to return until the function that is running (if any) has
        completed.

        If the target threw in the timing thread this function will reraise that
        exception here.
        """
        with self._lock:
            if self._thread is None:
                raise IntervalTimer.AlreadyStopped(
                    "Attempt to stop an IntervalTimer that is not running!")
            self._stop = True
            self._lock.notify_all()
        self._thread.join()
        with self._lock:
            exception = self._exception
            self._thread = None
            self._stop = False
            self._exception = (None, None, None)
        if raise_ and exception[0] is not None:
            raise exception[0], exception[1], exception[2]  # pylint:disable=raising-bad-type

    def __enter__(self):
        self.start()
        return self

    def __exit__(self, exc_type, _exc_value, _traceback):
        self.stop(raise_=exc_type is None)

    def _run(self):
        next_time = self._first_time
        while True:
            with self._lock:
                if self._stop:
                    return
                while True:
                    time_remaining = next_time - time.time()
                    if time_remaining <= 0:
                        break
                    self._lock.wait(time_remaining)
                    if self._stop:
                        return
                target, args, kwargs = self.target, self.args, self.kwargs
            next_time += self.interval
            try:
                target(*args, **kwargs)
            except BaseException:
                logging.warning(
                    "Function %s(*%r, **%r) threw in IntervalTimer",
                    target, args, kwargs, exc_info=True)
                self._exception = sys.exc_info()
                break


def test_interval_timer():
    from pytest import raises
    def cb(counter):
        counter[0] += 1
        time.sleep(0.2)

    count = [0]
    t = IntervalTimer(0.5, cb, args=(count,))
    with t:
        time.sleep(1.1)
    assert count[0] == 2

    # Can restart:
    with t:
        time.sleep(1.1)
    assert count[0] == 4

    # Has shutdown:
    time.sleep(1.1)
    assert count[0] == 4

    # Can't be run twice concurrently:
    with t:
        with raises(IntervalTimer.AlreadyRunning):
            with t:
                pass

    with raises(IntervalTimer.AlreadyStopped):
        t.stop()

    # Forwards exceptions:
    def cb_raising():
        raise RuntimeError("Everything is fine")

    t = IntervalTimer(0.1, cb_raising)
    with raises(RuntimeError, message="Everything is fine"):
        with t:
            time.sleep(0.2)

    with raises(RuntimeError, message="Doesn't override other exceptions"):
        with t:
            time.sleep(0.2)
            raise RuntimeError("Doesn't override other exceptions")
```

I did this really just to explore the problem-space.  No action necessary.

drothlis:
> I assume that `self.lock.wait(0)` (if `libcec.Transmit` took longer than
> 300ms) will release the lock if other callers (most importantly
> `keyup`!) are trying to acquire the lock.

Yes it does. I tested it by running the following Python script 400 times:

```
import time, threading

holding = False
lock = threading.Condition()

def send_keydowns():
    deadline = time.time() + 0.3
    with lock:
        while True:
            lock.wait(max(0, deadline - time.time()))
            if not holding:
                return
            deadline = time.time() + 0.3
            time.sleep(0.4)  # simulate what would happen if libcec.Transmit takes > 300ms
            print "send_keydowns"


t = threading.Thread(target=send_keydowns)
holding = True
t.start()

time.sleep(1)
with lock:
    holding = False
```

Most times it printed "send_keydowns" twice, as expected. Occasionally it would run for longer, up to printing "send_keydowns" 11 times. But it never hung indefinitely.

I don't actually know how long `libcec.Transmit` takes. I have no reason to believe that it takes longer than 300ms -- I'm just being careful / paranoid.

drothlis:
I've tested the lirc, roku & cec controls. @wmanley please review the last 3 commits & merge if happy.

_stbt/control_gpl.py
```diff
@@ -217,16 +217,23 @@ def keyup(self, key):
...
+        # if it hasn't seen one within a receiver-determined timeframe which
+        # can't be less than 550ms. For press-and-hold the initiator should
+        # send repeated keydown commands (200 to 450ms apart) followed by the
+        # final keyup.
         end_time = time.time() + 60
+        next_press_deadline = time.time() + 0.3
         with self.lock:
             while True:
-                self.lock.wait(0.450)
+                self.lock.wait(max(0, next_press_deadline - time.time()))
```
wmanley: The only place we call `notify_*` is keyup, so I guess that this is safe with Python's `Condition` implementation.

wmanley:
Hold on, you've still got an outstanding TODO item: "Test hold with CEC: PS4"

drothlis:
I'll fix the commit message & merge. I don't think I need to test on PS4 given that I've tested the CEC implementation wit hthe Fire TV.

wmanley:
It's not worth fixing.  I'll merge.

PR #505, opened 2018-05-18
stbt.is_screen_black: Return IsScreenBlackResult instead of bool

drothlis:
So that you can get the exact frame that was analysed, for more precise
performance measurements.

TODO:

- [x] Tests.
- [x] Release notes.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -1311,26 +1311,29 @@ class DeviceUnderTest(object):
 
         _region = Region.intersect(_image_region(frame), region)
         greyframe = cv2.cvtColor(crop(frame, _region), cv2.COLOR_BGR2GRAY)
-        _, greyframe = cv2.threshold(
-            greyframe, threshold, 255, cv2.THRESH_BINARY)
-        _, maxVal, _, _ = cv2.minMaxLoc(greyframe, mask.image)
+        if mask.image is not None:
+            cv2.bitwise_and(greyframe, mask.image, dst=greyframe)
+        maxVal = greyframe.max()
 
         if logging.get_debug_level() > 1:
             imglog = logging.ImageLogger("is_screen_black")
             imglog.imwrite("source", frame)
             if mask.image is not None:
                 imglog.imwrite('mask', mask.image)
-                imglog.imwrite('non-black-regions-after-masking',
-                               numpy.bitwise_and(greyframe, mask.image))
-            else:
-                imglog.imwrite('non-black-regions-after-masking', greyframe)
+            _, thresholded = cv2.threshold(greyframe, threshold, 255,
+                                           cv2.THRESH_BINARY)
+            imglog.imwrite('non-black-regions-after-masking', thresholded)
 
-        result = IsScreenBlackResult(maxVal == 0, frame)
+        result = IsScreenBlackResult(bool(maxVal <= threshold), frame)
         debug("is_screen_black: {found} black screen using mask={mask}, "
-              "threshold={threshold}, region={region}: {result}".format(
+              "threshold={threshold}, region={region}: "
+              "{result}, maximum_intensity={maxVal}".format(
                   found="Found" if result.black else "Didn't find",
                   mask=mask.friendly_name,
-                  threshold=threshold, region=region, result=result))
+                  threshold=threshold,
+                  region=region,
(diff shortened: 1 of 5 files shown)


_stbt/core.py
```diff
@@ -665,6 +665,27 @@ def test_motionresult_repr():
...
+    """
+    def __init__(self, frame, result):
+        self.frame = frame
+        self.result = result
+
+    def __nonzero__(self):
+        return self.result
+
+    def __repr__(self):
+        return ("IsScreenBlackResult(frame=<%s>, result=%r)" % (
```
wmanley: I think result should be first here - it's what we usually care about.

Perhaps it could be called `is_black` or `black` to be more descriptive too?  `MotionResult` has `.motion`.  `MatchResult` has `.match`.

tests/test_core.py
```diff
@@ -100,7 +100,16 @@ def e(kwargs):
...
-    assert stbt.is_screen_black(frame, mask, threshold, region) == expected
+    assert expected == bool(
+        stbt.is_screen_black(frame, mask, threshold, region))
+
+
+def test_IsScreenBlackResult():
+    frame = stbt.load_image("almost-black.png")
+    result = stbt.is_screen_black(frame)
+    assert result
+    assert numpy.all(result.frame == frame)
```
wmanley: Alternatively `result.frame is frame`?  Maybe that's too strong?

_stbt/core.py
```diff
@@ -1326,7 +1326,12 @@ def is_screen_black(self, frame=None, mask=None, threshold=None,
             else:
                 imglog.imwrite('non-black-regions-after-masking', greyframe)
 
-        return IsScreenBlackResult(frame, maxVal == 0)
+        result = IsScreenBlackResult(frame, maxVal == 0)
+        if result:
+            debug("is_screen_black: Found black screen: %s" % result)
+        else:
+            debug("is_screen_black: Didn't find black screen: %s" % result)
```
wmanley: Log the threshold, region and frame timestamp and mask filename here too.

drothlis:
Note: I didn't add attribute `IsScreenBlackResult.time` (unlike `MatchResult` and `MotionResult`) because you can get it via `IsScreenBlackResult.frame.time`.

_stbt/core.py

_stbt/core.py
drothlis: Done. I've also logged the actual maximum value (intensity) seen, which is useful if you need to adjust your threshold.

_stbt/core.py
```diff
@@ -1311,26 +1311,29 @@ def is_screen_black(self, frame=None, mask=None, threshold=None,
 
         _region = Region.intersect(_image_region(frame), region)
         greyframe = cv2.cvtColor(crop(frame, _region), cv2.COLOR_BGR2GRAY)
-        _, greyframe = cv2.threshold(
-            greyframe, threshold, 255, cv2.THRESH_BINARY)
-        _, maxVal, _, _ = cv2.minMaxLoc(greyframe, mask.image)
+        if mask.image is not None:
+            cv2.bitwise_and(greyframe, mask.image, dst=greyframe)
+        maxVal = greyframe.max()
```
wmanley: I wonder if this has an effect on performance?  In theory it has to do less work.  In practice it all depends on how optimised numpy is vs opencv for this operation.

_stbt/gst_utils.py
```diff
@@ -107,6 +107,14 @@ def __array_finalize__(self, obj):
             return
         self.time = getattr(obj, 'time', None)  # pylint: disable=attribute-defined-outside-init
 
+    def __repr__(self):
+        if len(self.shape) == 3:
+            dimensions = "%dx%dx%d" % (
+                self.shape[1], self.shape[0], self.shape[2])
+        else:
+            dimensions = "%dx%d" % (self.shape[1], self.shape[0])
+        return "<stbt.Frame(time=%r, dimensions=%s)>" % (self.time, dimensions)
```
wmanley: `"shape=%r" % (self.shape,)`?

drothlis: Numpy shapes are (height, width) which is confusing for people who aren't dealing with numpy directly.

PR #510, opened 2018-06-12
stbt auto-selftest enchancements

wmanley:
TODO:

- [ ] Add an associated test for each commit


drothlis:
Extracted `auto-selftest: Fix running against test-packs which are modules` into #513.

PR #511, opened 2018-06-14
Add `frames` parameter to `wait_for_motion` and `detect_motion`

wmanley:
They are no longer members of `DeviceUnderTest` so we no-longer need to
have forwarding implementations in `stbt/__init__.py`.

I think this general model makes some sense.  The different image
processing bits are no longer bound so tightly together.  But we make a
concession to convenience - frames defaults to `stbt.frames()`.  By
introducing this impurity we maintain the convenience of the existing API.
It makes the tests cleaner too.

### DIFF
--- a/_stbt/imgutils.py
+++ b/_stbt/imgutils.py
@@ -1,0 +1,164 @@
+import inspect
+import os
+from collections import namedtuple
+
+import cv2
+import numpy
+
+from .logging import ddebug
+from .types import Region
+
+
+class Frame(numpy.ndarray):
+    """A frame of video.
+
+    A ``Frame`` is what you get from `stbt.get_frame` and `stbt.frames`. It is
+    a subclass of `numpy.ndarray`, which is the type that OpenCV uses to
+    represent images. Data is stored in 8-bit, 3 channel BGR format.
+
+    In addition to the members inherited from `numpy.ndarray`, ``Frame``
+    defines the following attributes:
+
+    :ivar float time: The wall-clock time when this video-frame was captured,
+        as number of seconds since the unix epoch (1970-01-01T00:00:00Z). This
+        is the same format used by the Python standard library function
+        `time.time`.
+    """
+    def __new__(cls, array, dtype=None, order=None, time=None, _draw_sink=None):
+        obj = numpy.asarray(array, dtype=dtype, order=order).view(cls)
+        obj.time = time
+        obj._draw_sink = _draw_sink  # pylint: disable=protected-access
+        return obj
+
+    def __array_finalize__(self, obj):
+        if obj is None:
+            return
+        self.time = getattr(obj, 'time', None)  # pylint: disable=attribute-defined-outside-init
+        self._draw_sink = getattr(obj, '_draw_sink', None)  # pylint: disable=attribute-defined-outside-init
+
+    def __repr__(self):
+        if len(self.shape) == 3:
+            dimensions = "%dx%dx%d" % (
+                self.shape[1], self.shape[0], self.shape[2])
+        else:
+            dimensions = "%dx%d" % (self.shape[1], self.shape[0])
+        return "<stbt.Frame(time=%s, dimensions=%s)>" % (
+            "None" if self.time is None else "%.3f" % self.time,
+            dimensions)
+
+
+def _frame_repr(frame):
+    if frame is None:
+        return "None"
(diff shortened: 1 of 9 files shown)


wmanley:
@drothlis: Please review.  The two user-visible changes are:

1. `wait_for_motion`/`detect_motion` take a `frames` parameter
2. `Region.bounding_box` is public

Apart from that it's all refactorings.

I would pay particular attention to "Add private `logging.draw_on` function".  This is how I've been able to decouple drawing on frames from having access to `DeviceUnderTest.sink_pipeline`.  I think it's good, but the design may be contentious.

_stbt/imgutils.py
```diff
@@ -0,0 +1,164 @@
...
+from collections import namedtuple
+
+import cv2
+import numpy
+
+from .logging import ddebug
+from .types import Region
+
+
+class Frame(numpy.ndarray):
```
drothlis: Doesn't it make more sense for `Frame` to go in `_stbt.types`?

drothlis: I see, it's because `types.py` isn't supposed to have any dependencies outside the Python standard library. We could make an exception for numpy?

_stbt/motion.py
```diff
@@ -0,0 +1,348 @@
...
+
+def limit_time(frames, duration_secs):
+    """
+    Adapts a frame iterator such that it will return EOS after `duration_secs`
+    worth of video has been read.
+    """
+    end_time = None
+    for frame, ts in frames:
+        if end_time is None:
+            end_time = frame.time + duration_secs
```
drothlis: This is a change in behaviour: The timeout is the first frame's time + 10s, instead of `time.time()` + 10s.

This means that the effective timeout will be slightly shorter than before.

_stbt/motion.py
```diff
@@ -0,0 +1,348 @@
...
+        elif frame.time > end_time:
+            break
```
drothlis: Should we add similar debug like `stbt.frame` has?

```
debug("timed out: %.3f > %.3f" % (timestamp, end_time))
```

_stbt/logging.py
```diff
@@ -142,3 +142,13 @@ def test(level):
...
+
+
+def draw_on(frame, *args, **kwargs):
+    draw_sink_ref = getattr(frame, '_draw_sink', None)
+    if not draw_sink_ref:
+        return
+    draw_sink = draw_sink_ref()
+    if not draw_sink:
+        return
+    draw_sink.draw(*args, **kwargs)
```
drothlis: "A try/except block is extremely efficient if no exceptions are raised" [[1]] so it might make sense to use a `weakref.proxy` and catch `ReferenceError`? Just bikeshedding really.

[1]: https://docs.python.org/2/faq/design.html#how-fast-are-exceptions

drothlis: I expect the sink pipeline to outlive the frames, and we manage sink pipeline teardown explicitly (not via `__del__`). I wonder what is the performance impact of strong references vs. weak references. Probably not significant considering we'll only be doing it once every 16ms at most.

_stbt/motion.py
```diff
@@ -0,0 +1,348 @@
...
+        `load_image`) or a single-channel image in OpenCV format.
+
+    :type region: `Region`
+    :param region:
+        Only analyze the specified region of the video frame.
+
+        If you specify both ``region`` and ``mask``, the mask must be the same
+        size as the region.
+
+    :type frames: Iterator[stbt.Frame]
```
drothlis: This is actually `Iterator[(stbt.Frame, int)]`.

wmanley: Fixed

_stbt/motion.py
wmanley: Fixed.

_stbt/logging.py
wmanley: Each frame keeping the sink-pipeline alive would make me nervous, particularly given the teardown issues we've had in the past.

_stbt/motion.py
wmanley: Fixed.

_stbt/imgutils.py
wmanley: I would say leave it be for now and we can revisit the decision in the future.

wmanley:
I've fixed and squashed.  I'll merge once CI is happy.

PR #513, opened 2018-06-18
auto-selftest: Fix running against test-packs which are modules

wmanley:
This updates stbt auto-selftest to be compatible with test-packs that have relative imports.  See also 43fb524ca8.


drothlis:
A thing with subdirectories with `__init__.py` files is called a Python *package* not a *module*:

> Packages are a way of structuring Python’s module namespace by using “dotted module names”

-- https://docs.python.org/2/tutorial/modules.html#packages

I've fixed up the wording in the filenames & docstrings.

PR #515, opened 2018-06-18
Refactor: Move match functions into own file

wmanley:
As a result `detect_match` and `wait_for_match` gained a `frames` parameter.

This is analogous to fdca78f and 21f4440 but for the match functions, which took up a significant proportion of core.py.

`press_until_match` is still in `_stbt.core` because it involves pressing as well.

I've done this to `FrameObject` and there's a couple of other cleanups in here too.


stbt/__init__.py
```diff
@@ -64,18 +68,11 @@
     "get_frame",
     "is_screen_black",
     "load_image",
-    "match",
-    "match_all",
-    "match_text",
-    "MatchParameters",
-    "MatchResult",
-    "MatchTimeout",
```
drothlis: You probably didn't mean to remove any of these from `__all__`.

wmanley: Good spot.  Sloppy on my part.  I wonder why lint didn't object to unused imports?  Fixed

PR #522, opened 2018-07-06
Rittal power support

wmanley:
Like the ATEN this is controlled over SNMP.

I haven't written any tests for it.  I don't really believe in the value
of the existing ATEN tests so I don't want to duplicate them.  A better
approach might be to use [snmpsim] or similar to mock out the PDU at the
SNMP layer.

I also haven't investigated the failure modes of the PDU.

This is a bit rough, but could form the basis for a more mature implementation.

[snmpsim]: http://snmplabs.com/snmpsim/index.html


drothlis:
This hasn't been stress-tested to discover & handle the failure modes of the Rittal hardware, but it's a start. I'll merge it.

PR #524, opened 2018-09-12
stbt auto-selftest: Include full exception traceback

drothlis:
Unfortunately the first parameter (`limit`) limits to the *outermost*
frames, i.e. the code in stbt_auto_selftest.py. Otherwise we could have
set limit to 1 or 2 or 3. Negative numbers don't work.


wmanley:
Excluding the Exception output was deliberate: You don't want your selftests changing every time you make a change higher up in the file changing the line numbers.

PR #528, opened 2018-10-12
Region.bounding_box: Remove limit of just passing two regions

wmanley:
`bounding_box` is associative and commutative so there's no reason
not to.


drothlis:
Rebased onto master.

drothlis:
@wmanley please review my changes to your `erode` implementation & tests (the last fixup commit).

wmanley:
Right: so if we `dilate` with a too-large negative number we get an exception, but if we `erode` with a too-large number we get `None`.  This seems reasonable to me. +1

drothlis:
Squashed & rebased.

PR #530, opened 2018-10-19
stbt.match: Add structure_weighting

wmanley:
We use Sobel edge detection to improve the matching.  This is then
taken into account when constructing the heatmap.  This more closely
matches human expectations at the cost of being slower to run.

The weighting is done by raising the values to a power before
multiplying them together.  This is inspired by SSIM where seperate
brightness, contrast and structure terms are combined with the same
technique.

TODO:

- [ ] Include example in tests where it makes a difference
- [ ] Update debug HTML output

### DIFF
--- a/_stbt/match.py
+++ b/_stbt/match.py
@@ -101,7 +101,7 @@ class MatchParameters(object):
 
     def __init__(self, match_method=None, match_threshold=None,
                  confirm_method=None, confirm_threshold=None,
-                 erode_passes=None):
+                 erode_passes=None, structure_weighting=None):
         if match_method is None:
             match_method = get_config('match', 'match_method')
         if match_threshold is None:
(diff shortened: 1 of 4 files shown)


_stbt/match.py
```diff
@@ -452,10 +458,34 @@ def _find_matches(image, template, match_parameters, imglog):
...
+    weights = []
+
+    if structure_weighting > 0.05:
+        images.append(cv2.Sobel(image, cv2.CV_32F, 1, 0, ksize=3))
+        templates.append(cv2.Sobel(template, cv2.CV_32F, 1, 0, ksize=3))
+        weights.append(structure_weighting * 0.5)
+
+        images.append(cv2.Sobel(image, cv2.CV_32F, 0, 1, ksize=3))
+        templates.append(cv2.Sobel(template, cv2.CV_32F, 0, 1, ksize=3))
+        weights.append(structure_weighting * 0.5)
```
drothlis: Why do the x & y Sobel separately?

drothlis:
Misc. thoughts:

* Sobel operation is a convolution with the Sobel kernel. A convolution in the spatial domain is the same as a multiplication in the frequency domain. `cv2.matchTemplate` already does an FFT. We could do the FFT a single time, then do the Sobel edge-detection and `matchTemplate` search. This would require un-picking `matchTemplate` (OpenCV's license is 3-clause BSD).

_stbt/match.py
drothlis: I see. Specifying `xorder=1, yorder=1` gives diagonal edges, which isn't what we want.

PR #532, opened 2018-10-25
Add navigation function `navigate_fixed_layout`

wmanley:
This is the most basic navigator.  For now it works for pages:

* With a fixed layout i.e. no scrolling
* With no PIP or transparency showing live TV
* With no other ongoing animation

We can look at generalising this in future commits.

TODO:

- [ ] Tests

Possible future extensions:

- [ ] Track which regions have been selected before and bomb-out if we're not making forward progress
- [ ] Allow using a function other than `press_and_wait` on a frameobject-by-frameobject basis.
- [ ] Timeouts


drothlis:
Merged #531 and rebased onto master.

PR #533, opened 2018-10-26
make check: Don't run virtual-stb tests by default

drothlis:
You need to do `make check enable_virtual_stb=yes`.

This is because the tests screw up my X11 display when I run them
locally. Note that stbt-virtual-stb is already a separate debian
package because of all its dependencies on X11.


wmanley:
It's still run in travis, so this is safe.  Merging.

PR #534, opened 2018-11-22
Remove nose dependency from make check

wmanley:
This completes the work started in 03b7049dfd7.  We no longer need
`nose` to be installed to run `make check`.


Makefile
```diff
@@ -236,9 +234,7 @@ check-pytest: all tests/buttons.png tests/ocr/menu.png
 	              -e tests/auto-selftest-example-test-pack/tests/empty_dir/subdir/example_with_no_tests.py \
 	              -e tests/vstb-example-html5/ \
 	              -e tests/webminspector/ \
-	              -e vendor/) \
-	    nosetest-issue-49-workaround-stbt-control.py && \
-	rm nosetest-issue-49-workaround-stbt-control.py
```
drothlis: So now we aren't running `stbt-control::test_main`. I'm not sure what the point of that test is. Might as well remove it.

wmanley: Done

PR #535, opened 2018-11-22
Use CircleCI for Continuous Integration testing

drothlis:
Unlike Travis, CircleCI makes it easy to run tests inside a docker
container. So we'll build docker images for Ubuntu 16.04 and Ubuntu
18.04 (the latter in a future commit), push them to Docker
Hub (manually, needs to be done by the Stb-tester project owners but
only when we need to install new dependencies), and then CircleCI will
pull the image from Docker Hub each time it runs the tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,8 @@ commands:
     steps:
       - run:
           name: make check
+          environment:
+            LANG: en_GB.UTF-8
           command: |
             tesseract --version
             pylint --version
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,44 @@
+version: 2.1
+
+jobs:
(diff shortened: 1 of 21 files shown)


drothlis:
Hmmm it looks like CircleCI needs its config.yml to be on master branch, so I'll merge now and fix any problems as I find them. I've tested `make check` running in the new docker image, locally.

drothlis:
You can see it run here: https://circleci.com/gh/stb-tester/stb-tester/2
It took 2m31s, of which 35s was downloading the 1.6GB docker image from Docker Hub.
cf. the corresponding Travis job which took 6m38s.

A couple of tests failed because $TERM wasn't set but I fixed that and it passed here: https://circleci.com/gh/stb-tester/stb-tester/3

I'm going to remove the .travis.yml now.

PR #537, opened 2018-11-23
stbt.ocr: Only binarize the image if text_color_threshold > 0

drothlis:
In v28 we added `text_color` which calculates each pixel's distance from
the desired color, and then also binarizes the resulting image before
passing it to tesseract. But why do we do the binarization at all --
maybe the color-distance alone is enough, and tesseract can choose its
own binarization. Getting the binarization threshold right has proven
to be a trial-and-error process on a case-by-case basis.

UPDATE: It turns out that (for white text at least) doing color-distance
without our own thresholding, gives exactly the same result as no
color-distance at all. Where "the same result" means identical output
from tesseract's binarizer. See [comment below](#issuecomment-442833424).

TODO:

- [ ] We saw that text_color=white + text_color_threshold=0 makes no difference. Does it help if the text color is not white -- specifically if it isn't close to 255 in any channel?
- [ ] Try implementing an adaptive thresholding ourselves.

### DIFF
--- a/_stbt/ocr.py
+++ b/_stbt/ocr.py
@@ -213,9 +213,11 @@ def ocr(frame=None, region=Region.ALL,
         translucent overlay.
 
     :param int text_color_threshold:
-        The threshold to use with ``text_color``, between 0 and 255. Defaults
-        to 25. You can override the global default value by setting
-        ``text_color_threshold`` in the ``[ocr]`` section of :ref:`.stbt.conf`.
+        The threshold to use with ``text_color``, between 0 and 255. If 0, no
+        binarization is done (tesseract will choose its own binarization
+        threshold). If None, the default value will be read from
+        ``ocr.text_color_threshold`` in :ref:`.stbt.conf`, and if that's not
+        specified the default value is 25.
 
     :param engine:
         The OCR engine to use. Defaults to ``OcrEngine.TESSERACT``. You can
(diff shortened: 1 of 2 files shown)


drothlis:
Tesseract's segmentation operates on the binarized image:

1. A comment on [SegmentPage] says: "pix_binary_ is used as the source image".

2. The "PageSegInput" image you get when enabling tessedit_dump_pageseg_images:
pageseginput

N.B. The only other images you get from `tessedit_dump_pageseg_images` are "NoLines" (find and remove rule/separator lines -- see [FindAndRemoveLines]) and "NoImages" (mask out photo regions in the input -- see [FindImages]). But both of them are identical to the "PageSegInput" image, so it seems that `FindAndRemoveLines` and `FindImages` aren't having any effect.

All of the above are with `--psm=3` (Fully automatic page segmentation, but no Orientation & Script Detection), which is the default.

[SegmentPage]: https://github.com/tesseract-ocr/tesseract/blob/4.0.0/src/ccmain/pagesegmain.cpp#L96-L101
[FindAndRemoveLines]: https://github.com/tesseract-ocr/tesseract/blob/4.0.0/src/textord/linefind.cpp#L229
[FindImages]: https://github.com/tesseract-ocr/tesseract/blob/4.0.0/src/textord/imagefind.cpp#L58

drothlis:
# Tesseract's binarisation

Use Otsu's algorithm to determine a **global** threshold value for each colour channel.

If <25% of pixels are above the threshold, then that's the foreground colour. If <25% of pixels are *below* the threshold, then *that's* the foreground colour.  Either way, you end up with black text on a white background (after binarization).

Otherwise, there's no clear foreground/background split and that colour channel is ignored. If *none* of the channels have a clear foreground, then use the channel that's closest to having 25% either way. You'll always have at least 1 channel that's being considered.

Then, for each pixel: If any of the channels are above/below the threshold (as appropriate), set the output pixel to black.

See [ThresholdRectToPix] and [OtsuThreshold].

[ThresholdRectToPix]: https://github.com/tesseract-ocr/tesseract/blob/4.0.0/src/ccmain/thresholder.cpp#L298
[OtsuThreshold]: https://github.com/tesseract-ocr/tesseract/blob/4.0.0/src/ccstruct/otsuthr.cpp#L30

## Conclusions

Try implementing an adaptive thresholding ourselves.

PR #538, opened 2018-11-29
stbt.ocr: Add debug logging

drothlis:
- [x] Use ImageLogger
- [x] Tests
- [x] Make it work for match_text too
- [x] Show text_color intermediate images
- [x] Show tesseract's own thresholding
- [x] HTML for existing motion & is_screen_black stbt-debug images

Out of scope for this pull request:

- Use hocr mode even for `stbt.ocr`, draw the bounding boxes for each word / line / box.
- Draw the bounding boxes on the recorded video too!


drothlis:
Rebased onto master to get circleci config.

drothlis:
`tests/test_stbt_debug.py::test_match_debug` is failing but only if it's run
after `tests/test_set_config.py`.

The failure is in `scoped_curdir.__enter__`:

```
>               cwd = os.getcwd()
E               OSError: [Errno 2] No such file or directory
```

drothlis:
Hmm the integration tests on ubuntu1804 failed because "no output in 10 minutes" -- in spite of 1a442c2.

It appears to have been running `test_that_stbt_batch_propagates_exit_status_if_running_a_single_test` but that test doesn't involve any signal handling or anything complex like that.

_stbt/logging.py
```diff
@@ -111,7 +111,7 @@ def append(self, **kwargs):
                 self.data[k] = []
             self.data[k].append(v)
 
-    def imwrite(self, name, image, region=None, colour=None):
+    def imwrite(self, name, image, regions=None, colours=None):
```
wmanley: I think regions and colours would be nicer as a list of tuples rather than two lists, but it's not important.

_stbt/match.py
```diff
@@ -776,11 +789,17 @@ def _log_match_image_debug(imglog):
...
+           class="nav pull-left">«prev</a>
+        <a href="../{{ "%05d" % (frame_number + 1) }}/index.html"
+           class="nav pull-right">next»</a>
         <h4>
+            {{"Matched" if matched else "Didn't match"}}
             <i>{{template_name}}</i>
-            {{"matched" if matched else "didn't match"}}
         </h4>
 
+        <img src="source_with_matches.png" />
```
wmanley: We could use divs here like we do with `ocr`, rather than drawing on the image itself.

_stbt/match.py
```diff
@@ -751,6 +753,12 @@ def _log_match_image_debug(imglog):
             result.region, _Annotation.MATCHED if result._first_pass_matched  # pylint:disable=protected-access
             else _Annotation.NO_MATCH)
 
+    imglog.imwrite(
+        "source_with_matches", imglog.images["source"],
+        [x.region for x in imglog.data["matches"]],
+        [_Annotation.MATCHED if x.match else _Annotation.NO_MATCH
+         for x in imglog.data["matches"]])
```
wmanley: We could use `div`s here to draw the regions so we don't need to write `source_with_matches.png`

_stbt/logging.py
```diff
@@ -158,12 +158,40 @@ def html(self, template, **kwargs):
...
+                css_class = "nomatch"
+
+        return jinja2.Template(dedent("""\
+            <div class="region {{css_class}}"
+                 style="left: {{region.x / image.width * 100}}%;
+                        top: {{region.y / image.height * 100}}%;
+                        width: {{region.width / image.width * 100}}%;
+                        height: {{region.height / image.height * 100}}%"
+                 {% if title %}
+                 title="{{ title | escape }}"
```
wmanley: Will this title thing do mouseover?

_stbt/ocr.py
```diff
@@ -631,12 +631,7 @@ def _log_ocr_image_debug(imglog, text):
 
         <div class="annotated_image">
           <img src="source.png" />
```
wmanley: We should probably set `width="{{image.width}}"` and `height="{{image.height}}"` here so the browser can do layout before loading the image.

_stbt/ocr.py
```diff
@@ -631,12 +631,7 @@ def _log_ocr_image_debug(imglog, text):
...
-          <div class="ocr_region"
-               style="left: {{region.x / image.width * 100}}%;
-                      top: {{region.y / image.height * 100}}%;
-                      width: {{region.width / image.width * 100}}%;
-                      height: {{region.height / image.height * 100}}%"
-               title="{{ text | escape }}"></div>
+          {{ draw(roi, image_region, "roi", text) }}
```
wmanley: This is elegant.

_stbt/ocr.py
```diff
@@ -622,29 +629,54 @@ def _find_tessdata_dir():
...
-        <h4>OCR</h4>
+        <h4>{{title}}</h4>
 
         <div class="annotated_image">
           <img src="source.png" />
-          {{ draw(roi, image_region, "roi", text) }}
+          {{ draw(roi, image_region, "roi") }}
+          {% if match_text %}
+          {{ draw(result.region, image_region, result.match,
+                  title=result.text) }}
```
wmanley: It would be interesting to draw all the text that hocr read here, not just the matched region.

_stbt/ocr.py
wmanley: I see, maybe we can't do that or it will break the `max-width` bit.

_stbt/logging.py
drothlis: Yes, but I removed it in a later refactoring.

wmanley:
I think a common case is logging the same image over and over.  We can save images as `stbt-debug/images/<HASH>.png` and avoid the duplication.

PR #540, opened 2018-11-30
API: Add return value from `stbt.press`

wmanley:
A common pattern for benchmarking is:

    stbt.press('KEY_XXX')
    start_time = time.time()
    x = wait_until(xyz)
    duration = x.time - start_time

This moved the `time.time()` bits into `stbt.press` itself where it
can do a better job as it is `interpress_delay_secs` aware.

This was motivated by some changes I want to make to the transition
APIs.  When writing page objects you have to deal with transitions
between different pages.  For example a menu might know that it has
to press `'KEY_OK'` to launch the EPG, but only the EPG knows how
to wait until the EPG has finished loading.  In this case the `key`
argument to `press_and_wait` is defined by the `Menu` page object,
whereas the `mask` would be defined by `EPG`. This API is a step
towards being able to implement this as:

    r = stbt.press('KEY_OK')
    page = EPG.wait_until_loaded(r)

This motivates the `frame_before` member.

TODO:

- [x] Tests
- [x] How do we document return types?

### DIFF
--- a/tests/test-stbt-py.sh
+++ b/tests/test-stbt-py.sh
@@ -614,3 +614,35 @@ test_that_two_frames_iterators_can_return_the_same_frames_as_each_other() {
 	EOF
     stbt run -vv test.py || fail "Incorrect frames() behaviour"
 }
+
+test_that_press_returns_a_pressresult() {
+    cat > test.py <<-EOF &&
+	import time
+	
+	before = time.time()
+	result = stbt.press("KEY_MENU")
+	after = time.time()
+
+	assert before < result.start_time < result.end_time < after
+	assert result.key == "KEY_MENU"
+	assert isinstance(result.frame_before, stbt.Frame)
+
+	before = time.time()
+	result = stbt.press("KEY_HOME", hold_secs=0.001)
+	after = time.time()
+
+	assert before < result.start_time < result.end_time < after
+	assert result.key == "KEY_HOME"
+	assert isinstance(result.frame_before, stbt.Frame)
+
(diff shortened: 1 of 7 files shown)


tests/test-stbt-py.sh
```diff
@@ -614,3 +614,35 @@ test_that_two_frames_iterators_can_return_the_same_frames_as_each_other() {
...
+	before = time.time()
+	result = stbt.press("KEY_MENU")
+	after = time.time()
+
+	assert before < result.start_time < result.end_time < after
+	assert result.key == "KEY_MENU"
+	assert isinstance(result.frame_before, stbt.Frame)
+
+	before = time.time()
+	result = stbt.press("KEY_HOME", hold_secs=3)
```
drothlis: This will add 3 seconds to the tests. It would be 1 second, or even shorter? You could also check that `result.end_time - result.start_time > $hold_secs`.

drothlis:
> We might even want a `stbt.last_keypress` global? (Not for this PR, just idle thought).

Or why not. We wouldn't have to document it as a public API but could experiment with it a bit.

wmanley:
>> We might even want a `stbt.last_keypress` global? (Not for this PR, just idle thought).
>
>Or why not. We wouldn't have to document it as a public API but could experiment with it a bit.

I've done this.  You're right, it will be interesting how it works out.

tests/test-stbt-py.sh
wmanley: Fixed.

PR #542, opened 2018-12-05
stbt.match: Support transparency in reference image

drothlis:
This requires OpenCV 3.

TODO:

- [x] Implement our own sqdiff normalisation ([matchTemplateMask] only supports SQDIFF or CCORR_NORMED). #543.
- [x] First pass.
- [x] Second pass.
- [x] stbt-debug html.
- [x] Tests.
- [x] Benchmarks.
- [x] Documentation.
- [x] Try enabling pyramid levels.
- [x] Test with a completely transparent reference image.
- [x] Fix `wait_for_match`.

[matchTemplateMask]: https://github.com/opencv/opencv/blob/3.2.0/modules/imgproc/src/templmatch.cpp#L840-L917


drothlis:
Tests are failing because buttons.svg scale change caused by newer version of Inkscape.

drothlis:
Hmm I've realised that we can safely enable pyramid optimisation, as long as when we downscale the mask it becomes wider rather than narrower. [cv2.pyrDown] does a 5x5 Gaussian blur before discarding every other row & column, so our mask's hard edges will become blurred. As long as we treat *any* value < 255 as 0, then our downsampled mask is at worst going to be more permissive (masking out more, never masking out less), and thus still suitable.

[cv2.pyrDown]: https://docs.opencv.org/3.2.0/d4/d86/group__imgproc__filter.html#gaf9bba239dfca11654cb7f50f889fc2ff

_stbt/match.py
```diff
@@ -421,7 +471,7 @@ def __str__(self):
 
 
 @memoize_iterator({"version": "25"})
```
wmanley: ~~Bump version?~~  I see it's done later.

_stbt/match.py
```diff
@@ -544,14 +540,18 @@ def imwrite(name, img, scale=1):
...
 
         # Relax the threshold slightly for scaled-down pyramid levels to
         # compensate for scaling artifacts.
-        threshold = max(
-            0,
-            match_parameters.match_threshold - (0.2 if level > 0 else 0))
+        if level == 0:
+            relax = 0
+        elif match_parameters.match_method == MatchMethod.SQDIFF:
+            relax = 0.01
```
wmanley: How did you choose this relax value?  Strictly this isn't a change for transparency, it's general performance with SQDIFF, but it doesn't need to be split out.

_stbt/match.py
```diff
@@ -305,12 +310,57 @@ def _match_all(image, frame, match_parameters, region):
     if match_parameters is None:
         match_parameters = MatchParameters()
 
-    template = _load_image(image)
-
     if frame is None:
         import stbt
         frame = stbt.get_frame()
 
+    template = _load_image(image, cv2.IMREAD_UNCHANGED)
```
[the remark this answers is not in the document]
drothlis: Good idea.

_stbt/match.py
drothlis: It seems to work. I think that, given we're starting "from scratch" with a new match method, we want to avoid repeating the mistake of making it too permissive. This has significant performance benefits because the pyramid optimisation works more often. Ultimately I'd like to even remove the second pass (perhaps only enable it when we detect that the reference image has very low contrast -- see [this comment](https://github.com/stb-tester/stb-tester/pull/544#issuecomment-447799743) on #544).

Before we release v30 I'll do more dogfooding and we can adjust if necessary.

drothlis:
> Also we probably want to remove the same parameter \[`flags`] from our public load_image API for the same reasons.

This broke several of our tests so I'm concerned it'd be a user-breaking change. I've moved that to #547 for the sake of getting this one merged. I *have* fixed the bug in `wait_for_match`.

drothlis:
Hmm the virtual-stb tests (of all things) have found a new problem: Some PNGs can be 16 bits per channel:

```
 cv2.imread("tests/vstb-example-html5/tests/stb-tester-350px.png", cv2.IMREAD_ANYDEPTH)
Out[6]: 
array([[65535, 65535, 65535, ..., 65535, 65535, 65535],
       [65535, 65535, 65535, ..., 65535, 65535, 65535],
       [65535, 65535, 65535, ..., 65535, 65535, 65535],
       ...,
       [40891,   851,     0, ..., 65535, 65535, 65535],
       [48359,   427,  5257, ..., 65535, 65535, 65535],
       [65535, 42236, 63795, ..., 65535, 65535, 65535]], dtype=uint16)
```

This causes a mis-match between the frame & reference image dtypes.

The [wikipedia article on PNG] shows several other modes:

* Indexed colors where pixel values are indexes into a palette (table) rather than RGB values.
* An alpha channel isn't the only way of adding transparency: In RGB, grayscale, and indexed modes you can specify a single pixel value that means 100% transparent.

I don't know how `cv2.IMREAD_UNCHANGED` will treat any of the above.

[PNG]: https://en.wikipedia.org/wiki/Portable_Network_Graphics

drothlis:
I suppose that, when upgrading, you can convert all your reference images with:

```
filenames = subprocess.check_output(["git", "ls-files", "tests/*.png"]).strip().split("\n")
for f in filenames:
    img = cv2.imread(f)
    cv2.imwrite(f, img)
```

drothlis:
Neither `cv2.imwrite` nor gimp 2.8 can write 16-bit-per-channel PNGs (even from a uint16 numpy array).

_stbt/match.py
```diff
@@ -316,6 +316,15 @@ def _match_all(image, frame, match_parameters, region):
     template = _load_image(image, cv2.IMREAD_UNCHANGED)
     t = template.image
     mask = None
+
+    if t.dtype == numpy.uint16:
+        warn("Reference image %s has 16 bits per channel. Converting to 8 bits."
+             % (template.friendly_name))
+        t = cv2.convertScaleAbs(t, alpha=1.0 / 256)
```
wmanley: I'm not sure that this is correct.  Should the scale factor be `255. / 65535.` == `1./257`?

PR #544, opened 2018-12-11
stbt.match: Change default to SQDIFF

drothlis:
Branched from #542.

TODO:

- [x] Merge #542 and rebase.
- [x] Figure out an appropriate threshold (0.98).
- [x] Don't reduce the threshold so much for the smaller pyramid images? (Reduced `relax` to 0.01 in commit 35d0358.)
    - [x] Is this enough? Probably -- the 0.98 is already quite permissive.
    - ~~Should we remove the reduction altogether? I suppose the original motivation was to ensure identical results with pyramid optimisation enabled or disabled. We are confident in it now so we never disable it, and 0.98 is fairly permissive already. We need to keep it for the older match methods for backwards compatibility, but remove it (i.e. set `relax=0`) for `SQDIFF`.~~ No; I've seen the first-pass match be lower by 0.007 at the smaller pyramid levels, so reducing by 0.01 seems appropriate.
- [x] Can we disable the second pass by default? No: `test_match.py::test_that_results_dont_overlap` --which uses [this][1] very dark reference image, which *does* have some structure but it's so dark that it's hard to see-- finds dozens of false positives with 0.999 confidence if we disable the second pass (all the true positives are 1.0000 confidence).
- [x] Fix `test_auto_selftest_caching`.
- [x] Fix `test_template_annotation_labels`.

[1]: https://github.com/stb-tester/stb-tester/blob/576e817/tests/action-panel-template.png


drothlis:
`test_auto_selftest_caching` is failing -- I think because `stbt.match` is so much faster with the more precise matching at the lower pyramid levels.

On CircleCI:

    Run 1 (ubuntu 16.04)                 Run 2 (ubuntu 18.04)
                                        
    ==> without_cache.time <==           ==> without_cache.time <==
    3.19                                 4.72
    ==> cold_cache.time <==              ==> cold_cache.time <==
    2.59                                 4.06
    ==> hot_cache.time <==               ==> hot_cache.time <==
    2.19                                 3.53

On my laptop:

    v29                                  HEAD

    ==> without_cache.time <==           ==> without_cache.time <==
    7.93                                 4.43
    ==> cold_cache.time <==              ==> cold_cache.time <==
    5.34                                 3.72
    ==> hot_cache.time <==               ==> hot_cache.time <==
    3.23                                 3.22

drothlis:
> Can we disable the second pass by default? 

No: test_match.py::test_that_results_dont_overlap --which uses [this][1] very dark reference image, which does have some structure but it's so dark that it's hard to see-- finds dozens of false positives with 0.999 confidence if we disable the second pass (all the true positives are 1.0000 confidence).

[1]: https://github.com/stb-tester/stb-tester/blob/576e817/tests/action-panel-template.png

drothlis:
Rebased onto master and lowered the threshold to 0.98.

drothlis:
Why 0.98:

SQDIFF (our normalisation) is much more consistent than SQDIFF_NORMED (OpenCV's normalisation) -- the latter gives lower certainty score for dark images than for light images, so we've always had to use a very forgiving threshold.

The new SQDIFF gives higher certainty scores overall so we need to increase the threshold a lot. Also, now that we have transparency in our toolkit, users can ignore some pixels by marking them transparent in the reference image instead of lowering the threshold.

0.98 is still quite forgiving. For example `tests/button.png` (a button on a white background, without any transparency) continues to match the button that overlays another button (so the background outside the top-left & bottom-left corners is a bit different) at 0.984:

| reference image (without transparency) | still matches overlapping button |
|---|---|
| button | source_with_matches |

---

0.99 fails to match images that *have* changed a bit due to text rendering or minor redesign, but that will still match at the 2nd pass. For example I saw one case where the squares in the BBC iPlayer logo shrunk by a pixel all around:

| old reference image | new frame captured (cropped to the matching location) |
|---|---|
| iplayer-reference-image | iplayer-frame-cropped |

This matches with 0.986. The second pass still passes because of the erode step:

screenshot from 2019-01-18 18-45-00

wmanley:
```
    ==> without_cache.time <==
    4.43
    ==> cold_cache.time <==
    3.72
    ==> hot_cache.time <==
    3.22
```

Hmm, I wonder where the time is being spent now?

PR #545, opened 2018-12-11
Remove `detect_match` and `use_old_threading_behaviour`

wmanley:


tests/test-stbt-py.sh
```diff
@@ -613,29 +595,6 @@ test_global_use_old_threading_behaviour() {
 }
```
wmanley: I think it's probably worth keeping the parts of these tests that I've kept.  Perhaps the tests need renaming. @drothlis: What do you think?

docs/release-notes.md
```diff
@@ -31,6 +31,11 @@ TODO: Date
   This was introduced in v28, but seems to be unused. See release notes for v28
   below for more information.
 
+* Removed API `stbt.detect_match`. This has been redundant since we introduced
+  `stbt.match` in 0.21 (Dec 2014). It is unlikely there are many uses of it in
+  the wild and it is easily replaced with a combination of `stbt.frames` and
+  `stbt.match`.
+
```
drothlis: Might be worth explaining that it is easily mis-used: if you don't treat the result as an iterator, it is a truthy behaviour so you might think it is saying that a match was found.

drothlis:
I'll rebase & do the fixups I suggested (I want to get the release notes template on master). :-)

PR #547, opened 2018-12-17
stbt.load_image: Change `flags` default value to BGR[A]

drothlis:
This makes transparency work when you call `stbt.load_image` and pass the result to `stbt.match`, instead of passing the filename to `stbt.match`.

TODO:

- [x] Merge #542 & rebase.
- [x] Remove `cv2` from top-level code for the sake of (future) pylint-stubs generation.
- [x] ~~Provide config value for backwards compatibility?~~ I've found a way to provide backward compatibility (see #548).
- [x] Strip alpha channel if all pixels == 255? We use `frame=load_image(...)` a lot in our unit tests, and many of our screenshots have an alpha channel (albeit 100% opaque).

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -30,7 +30,7 @@ from _stbt import logging
 from _stbt.config import get_config
 from _stbt.gst_utils import (array_from_sample, gst_iterate,
                              gst_sample_make_writable)
-from _stbt.imgutils import find_user_file, Frame
+from _stbt.imgutils import find_user_file, Frame, imread
 from _stbt.logging import ddebug, debug, warn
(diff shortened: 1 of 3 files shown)


drothlis:
@wmanley please review. This (plus #548) mostly fixes any (unintentional) incompatible behaviour caused by transparency in #542.

Of course if you use `load_image` to load an image that does have some transparency, the behaviour is different than v29, but that's intentional.

I've left `flags` as an optional parameter so that you can load a greyscale image as a single channel array by passing `cv2.IMGREAD_GRAYSCALE`, for example to load a mask for your custom image-processing function.

PR #555, opened 2019-01-22
transition: Ignore a few scattered low-magnitude pixel differences

drothlis:
I have seen `press_and_wait` fail because a static screen still has a few pixels different (usually 1 to 9 pixels, with the magnitude of the difference between 1 and 16 in a single channel). I don't know if this is caused by the video-source or the video-capture hardware, or something in between.

For a more comprehensive solution see #554: Allow specifying the diffing algorithm.


_stbt/transition.py
```diff
@@ -222,7 +222,7 @@ def strict_diff(f1, f2, region, mask_image):
     if mask_image is not None:
         absdiff = cv2.bitwise_and(absdiff, mask_image, absdiff)
 
-    return absdiff.any()
+    return numpy.count_nonzero(absdiff) > 50 or (absdiff > 20).any()
```
wmanley: These are fixed numbers independent of the size of frame passed in.  They may make sense for 720p (~3,000,000 elements) but would make less sense if we've passed in a region or applied a mask. OTOH 50 elements is a 4x4px region which perhaps is small enough for any image.

PR #557, opened 2019-01-24
Added support for RedRat X

msiden:
Made some small modifications to irnetbox.py making it possible to
use a RedRat X IR blaster with stbt. This basically just involved
adding the RRX PID (13) to class NetBoxTypes and modifying the if/else
in method irsend_raw.

This will treat the RRX same as an IRNetBox allowing for communication
with the device using ethernet. Please note that it will NOT support
the RRX's USB interface.

I have tested this implementation thoroughly on my side and find that
it works well. According to the people at RedRat chances are good it
will also work with IRNetBox 4 but I have been unable to verify this.

TODO:

- [x] Fix CI build failure
- [x] Use mk3 protocol, not mk2 protocol


_stbt/irnetbox.py
```diff
@@ -122,14 +122,14 @@ def indicators_off(self):
...
-        * `port` is a number between 1 and 16.
+        * `port` is a number between 1 and 16 (or 1 and 4 for if you yse a RedRat X).
         * `power` is a number between 1 and 100.
         * `data` is a byte array as exported by the RedRat Signal DB Utility.
 
         """
         if self.irnetbox_model == NetBoxTypes.MK1:
             raise Exception("IRNetBox MK1 not supported")
-        elif self.irnetbox_model == NetBoxTypes.MK2:
+        elif self.irnetbox_model in (NetBoxTypes.MK2, NetBoxTypes.RRX, NetBoxTypes.MK4):
```
drothlis: It's a bit odd that the MK4 uses the protocol of the MK2, not of the MK3. Maybe they all continue to support the old protocol, but it would be better to use the new protocol as it allows finer-grained control of the output power level. The RRX probably supports the new protocol too (would need to be tested).

msiden: This is the message I got from Chris at RedRat:

"We've tried to keep the message protocol backwards compatible, so newer versions of the devices should support the same messages. Also, the RR-X and MK-IV share quite a lot of code, so the MK-IV should work ok.
(edit) 
•	MK-III: Supports the "sync" API, but added asynchronous IR output messages to all simultaneous output of different IR signals on different ports. As far as I'm aware, the stb-tester plugin only supports the sync API? 
•	MK-IV: Same as MK-III, but added additional messages for new features. 
•	RR-X: Same as MK-III, also with additional messages for RR-X specific functionality. 
So, theoretically, if you stick to the core sync message API, then I think it should basically work on all devices."

So, yes, if I understand things correctly, MK3, MK4 and RRX should all support the async protocol but to be sure and to keep things simple we might want to stick to the sync protocol. 

I'm willing to take a quick look into using the async protocol on RRX but please be aware that the solution provided here is the one I've been using for a year now without any issues. I won't be able to test anything on MK3 or MK4 since I don't have access to these devices.

drothlis: To clarify, I'm not asking you to write any new code. I'm suggesting that you use the code that's already there for the MK3, instead of using the code for the MK2. This should do it:

```suggestion
        elif self.irnetbox_model == NetBoxTypes.MK2:
```

... because the MK4 & RRX will be handled by the `else` block later on (you'll need to test it).

You'll also need to make this pull request pass CI. If you click on the red "X" icon it'll tell you why it failed.

msiden: I tested your suggestion, which I think is identical to the current release, and unfortunately it doesn't seem to work. See stack trace below:

  File "/usr/lib/stbt/stbt-control", line 307, in <module>
    main(sys.argv)
  File "/usr/lib/stbt/stbt-control", line 48, in main
    remote.press(args.remote_control_key)
  File "/usr/lib/stbt/_stbt/control.py", line 408, in press
    port=self.output, power=100, data=self.config[key])
  File "/usr/lib/stbt/_stbt/irnetbox.py", line 168, in irsend_raw
    data))
  File "/usr/lib/stbt/_stbt/irnetbox.py", line 198, in _send
    "IRNetBox returned NACK (error code: %d)" % error_code)
Exception: IRNetBox returned NACK (error code: 51)

msiden: I agree it would be preferable to use the async protocol and I'm sure there is a way to accomplish this but I don't feel this is something I can spend more time on at the moment I'm afraid.

As for the failed CI, I have fixed some of the issues but I can't figure out why I'm getting these ImportErrors. I haven't made any changes to this module.

************* Module stbt_camera_calibrate
stbt-camera.d/stbt_camera_calibrate.py:159:4: [E0401(import-error), qrc] Unable to import 'qrcode'
stbt-camera.d/stbt_camera_calibrate.py:160:4: [E0401(import-error), qrc] Unable to import 'qrcode.image.svg'
stbt-camera.d/stbt_camera_calibrate.py:183:8: [E0401(import-error), QRScanner.__init__] Unable to import 'zbar'
stbt-camera.d/stbt_camera_calibrate.py:188:8: [E0401(import-error), QRScanner.read_qr_codes] Unable to import 'zbar'
stbt-camera.d/stbt_camera_calibrate.py:369:4: [E0401(import-error), colour_graph] Unable to import 'matplotlib'
stbt-camera.d/stbt_camera_calibrate.py:505:4: [E0611(no-name-in-module), list_cameras] No name 'GUdev' in module 'gi.repository'

drothlis: I see. According to the statement from RedRat, the RedRatX does support the OUTPUT_IR_ASYNC command, but it's returning an error (NACK). I suppose that's because we're sending an array of 16 ports, but the RedRatX only has 4.

If someone were to fix that, they'd need to change the following lines. But I think we can merge your solution for now.

https://github.com/stb-tester/stb-tester/blob/36be705bb4fe9ae18eb1c5be6fb5ba66b95b1dad/_stbt/irnetbox.py#L156

https://github.com/stb-tester/stb-tester/blob/36be705bb4fe9ae18eb1c5be6fb5ba66b95b1dad/_stbt/irnetbox.py#L166

drothlis: However I do think that we shouldn't include the MK4 here since you haven't tested it. If my theory is correct (in my previous comment) then the MK4 will work with the same code-branch for the MK3.

```suggestion
        elif self.irnetbox_model in (NetBoxTypes.MK2, NetBoxTypes.RRX):
```

drothlis: Thanks for fixing the pylint warnings. The other ones are my fault because I've been changing the CI infrastructure recently -- I'll fix them.

drothlis: If you have time to test it, the patch would look like this:

```
diff --git a/_stbt/irnetbox.py b/_stbt/irnetbox.py
index ff2a20b..c043f3c 100644
--- a/_stbt/irnetbox.py
+++ b/_stbt/irnetbox.py
@@ -79,6 +79,7 @@ class IRNetBox(object):
                     raise
         self._responses = _read_responses(self._socket)
         self.irnetbox_model = 0
+        self.ports = 16
         self._get_version()
 
     def __enter__(self):
@@ -158,7 +159,7 @@ class IRNetBox(object):
             self._send(MessageTypes.OUTPUT_IR_SIGNAL)
             self.reset()
         else:
-            ports = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
+            ports = [0] * self.ports
             ports[port - 1] = power
             sequence_number = random.randint(0, (2 ** 16) - 1)
             delay = 0  # use the default delay of 100ms
@@ -168,7 +169,7 @@ class IRNetBox(object):
                     ">HH16s%ds" % len(data),
                     sequence_number,
                     delay,
-                    struct.pack("16B", *ports),
+                    struct.pack("%dB" % len(self.ports), *ports),
                     data))
 
     def _send(self, message_type, message_data=""):
@@ -206,6 +207,10 @@ class IRNetBox(object):
 
     def _get_version(self):
         self._send(MessageTypes.DEVICE_VERSION)
+        if self.irnetbox_model == NetBoxTypes.RRX:
+            self.ports = 4
+        else:
+            self.ports = 16
 
 
 def RemoteControlConfig(filename):
```

msiden: I tested your suggested changed real quickly and it looks like it might work so let me get back with an update. I want to test it a bit more thoroughly before I commit it. I'll see if I can include the above patch as well.

msiden: Just pushed a new commit that uses your suggested patch. I've tested this on the RRX for about 8 hours and it's been working fine. I'll continue to run my tests on this code to verify it's stability over time but I'm feeling pretty confident there won't be any issues.

msiden: I was wondering what this line does though. More specifically, does "16" refer to the IR ports, i which case this might have to be changed as well?
```
sequence_number` = random.randint(0, (2 ** 16) - 1)
```

drothlis: > `sequence_number = random.randint(0, (2 ** 16) - 1)`

That's unrelated (the sequence_number is a 16-bit integer; that's where the 16 comes from).

drothlis:
Just got back from holiday. Merged. Thanks!

drothlis:
P.S. The CI failure on Ubuntu 16.04 was caused by a configuration change (#582) made after this pull request was branched from master. The 18.04 CI did pass.

msiden:
Thanks David! For the record, I have been using this code daily on two test rigs for the past few weeks and it has been running flawlessly.

PR #558, opened 2019-01-24
stbt.match: Merge neighbouring ROIs from previous pyramid level

drothlis:
If the previous pyramid level (that is, running `cv2.matchTemplate` on a
smaller version of the image) identified Regions Of Interest (ROIs)
that are small and close together, then at the next pyramid level we
have to invoke `cv2.matchTemplate` (now on the full-sized image)
multiple times.

This is still a worthwhile optimisation compared to running
`cv2.matchTemplate` on the entire full-sized image; but if we merge
ROIs that are very close together, it's even faster.

This effect is particularly pronounced, now that we've tightened the
match threshold, if you're looking for an image of a word and the frame
has a full paragraph of text; the word might (tentatively) match at
every word in the paragraph, but not at the spaces between the words.

In the above scenario, I measured 47ms vs. 69ms, on my laptop.

N.B. The morphological operation "Closing" means to dilate and then
erode -- it has the effect of closing holes or gaps. Using a square
kernel (all ones) is 4 times faster than using an elliptical kernel
(250µs for the `c2v.morphologyEx` vs. 1.1ms).

TODO:

- [x] Fix match failures at edge locations.
- [x] Measure the performance impact across a wider range of examples.


drothlis:
|  | master (sqdiff @ 0.98) | this branch |
|---|---|---|
| Looking for this reference image: |  template |
| In this frame: | source |
| ROIs identified at previous pyramid level: | level1-source_matchtemplate_threshold | level1-source_matchtemplate_threshold |
| Becomes ROIs in the full sized image: |  level0-source_with_rois | level0-source_with_rois |
| Number of `cv2.matchTemplate` calls: | 97 | 12 |
| `%timeit stbt.match(template, frame)` takes: | 69ms | 47ms |

drothlis:
Ironically the old sqdiff-normed @ 0.80 takes 54ms: better than the current master, but not as fast as this PR.

Because the old threshold is more generous, the ROIs were more likely to be joined already.

drothlis:
Rebased onto master to get test fixes from #560.

drothlis:
On 1518 template/screenshot pairs from a real test-pack (69 templates x 22 screenshots)

Greatest improvement (absolute): 130ms (37%)
Greatest improvement (relative): 44% (17ms)
Worst deterioration (absolute): 53ms (14%)
Worst deterioration (relative): 24% (22ms)

Average improvement (absolute): 1.758s

Here's the worst deterioration:

| | master (sqdiff @ 0.98) | this branch |
|---|---|---|
| ROIs identified at pyramid level 2: | level2-source_matchtemplate_threshold | level2-source_matchtemplate_threshold |
| Becomes ROIs in pyramid level 1: | level1-source_with_rois | level1-source_with_rois |
| Time of stbt.match: | 91ms | 112ms |

drothlis:
Reducing the kernel size of the closing operation at the smallest pyramid level to (5,5) seems to be an improvement but still isn't unambiguously better than master:

Greatest improvement (absolute): 133ms (37%)
Greatest improvement (relative): 42% (106ms)
Worst deterioration (absolute): 59ms (31%)
Worst deterioration (relative): same one

Average improvement (absolute): 2.876s

All the differences (positive means an improvement):
14a5897

drothlis:
There are almost as many regressions in performance as there are improvements, so I'm giving up on this idea.

drothlis:
Reopening -- I want to revisit this now that we've merged #569.

drothlis:
Average improvement of 7ms per `match` across 1200 screenshot×template pairs.

screenshot from 2019-02-15 14-18-17

PR #561, opened 2019-01-28
stbt.match: Performance optimisation: Don't threshold last matchTemplate heatmap

drothlis:
Here we create `roi_mask` which is used in the next iteration of the
`for` loop. In the last iteration we don't need to create a new
`roi_mask`, because it isn't used after the `for` loop.

This saves up to 1ms on my laptop.

### DIFF
--- a/_stbt/imgproc_cache.py
+++ b/_stbt/imgproc_cache.py
@@ -352,7 +352,7 @@ def test_that_cache_speeds_up_match_all():
     cached_time, uncached_time, cached_result, uncached_result = (
         _check_cache_behaviour(match_all))
 
-    assert uncached_time > (cached_time * 4)
+    assert uncached_time > (cached_time * 2)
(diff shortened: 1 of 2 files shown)


_stbt/imgproc_cache.py
```diff
@@ -352,7 +352,7 @@ def match_all():
     cached_time, uncached_time, cached_result, uncached_result = (
         _check_cache_behaviour(match_all))
 
-    assert uncached_time > (cached_time * 4)
+    assert uncached_time > (cached_time * 2)
```
wmanley: Presumably you had to change this because it got faster,

drothlis: Yeah, when it failed the uncached time was 0.0148 and cached time was 0.0038, so 1 extra millisecond in unchached time would have passed.

PR #565, opened 2019-02-11
match: Change sqdiff normalisation

wmanley:
Previously we'd normalise SQDIFF matching such that a certainty of 1 is
perfect match and a certainty of 0 is the match of a white template against
a black image (or vice-versa).  This normalisation has some strange
properties:

* You can't get a certainty of zero if your template is not extreme
* You can't get a certainty of zero if your frame is not extreme
* For a mid-grey template (128, 128, 128) the *minimum* certainty possible
  is 75%!
* For a mid-grey template comparing against a random noise (i.e. completely
  decorrelated with the template) you expect a certainty of 92%!
* For a black/white template compared against noise you expect a certainty
  of 66%!

This commit changes the normalisation such that:

* Comparing *any* template against noise should give a certainty of ~0.
* The normalisation only depends on the template and not the frame (unlike
  sqdiff-normed).

We do this by calculating the average sqdiff of the template against all
possible frames.  This works out to be:

    ²⁵⁵₀∫(v-k)²dk
    -------------
         255

    = v² - 255v + 255²/3

Summed over the whole template where v is a single pixel/channel value
between 0 - 255.

With this normalisation method we get negative certainties too, but I clamp
them away to avoid confusing users.

TODO:

- [ ] Choose appropriate defaults
- [ ] Check debug html

### DIFF
--- a/_stbt/match.py
+++ b/_stbt/match.py
@@ -377,7 +377,7 @@ def _match_all(image, frame, match_parameters, region):
                                  .translate(input_region.x, input_region.y)
             result = MatchResult(
                 getattr(frame, "time", None), matched, match_region,
-                first_pass_certainty, frame,
+                max(0, first_pass_certainty), frame,
                 (template.relative_filename or template.image),
                 first_pass_matched)
             imglog.append(matches=result)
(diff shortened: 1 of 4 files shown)


drothlis:
What I like about our existing SQDIFF normalisation:

- "Reasonably close" means an error of 0.02 (threshold of 0.98) -- always, regardless of the contents of the template image.
- With the previous SQDIFF_NORMED, we had to relax the threshold significantly to avoid false negatives for some types of images, because for a dark image a threshold of 0.8 is still "reasonably close". This increases false positives for bright images.

This pull request re-introduces the undesirable property that the actual error is stretched/compressed according to the contents of the template image.

> For a mid-grey template (128, 128, 128) the minimum certainty possible is 75%!

For a mid-grey template "A", the minimum certainty is higher than for a completely black template "B", because the most different image from "A" **isn't as different** as "B"'s most different image. Seems fair enough to me.

I don't see why it matters what the lower bound of the certainty is.

drothlis:
We can revisit this if we hit a problem that it would solve. As it is I'm fairly satisfied with the current SQDIFF normalisation.

PR #566, opened 2019-02-12
stbt.match: Add fast-path for the case where image size == template size

wmanley:
This is more likely now we've got transparency - you can match a whole frame excluding the dynamic parts.

I implemented it in C for speed.  On my machine the C implementation is 100x faster than the implementation using numpy and 10x faster than using numba when comparing a 720p template and frame. It takes ~250us for the comparison.

I confirmed that GCC is able to vectorise the functions using the godbolt compiler explorer: https://godbolt.org/z/-3uO1z

**TODO:**

- [x] Update debug html


Makefile
```diff
@@ -74,6 +74,7 @@ INSTALL_PYLIB_FILES = \
     _stbt/imgproc_cache.py \
     _stbt/imgutils.py \
     _stbt/irnetbox.py \
+    _stbt/libstbt-30.so \
```
drothlis: It isn't a public API (we don't expect others to link to it) so we might as well call it `libstbt.so` and save the hassle of bumping the version number every release.

tests/test_match.py
```diff
@@ -280,3 +280,53 @@ def test_png_with_16_bits_per_channel():
     assert stbt.match(
         "tests/uint16.png",
         frame=cv2.imread(_find_file("uint8.png")))
+
+
+@requires_opencv_3
+def test_match_fast_path():
```
drothlis: This only requires OpenCV 3 because we complain if the image has transparency, before we decide to go down the fast path.

tests/test_match.py
```diff
@@ -280,3 +280,53 @@ def test_png_with_16_bits_per_channel():
...
+        reference = _load_image(reference)
+        orig_m = stbt.match(reference, frame=frame)
+        fast_m = stbt.match(reference, frame=frame, region=orig_m.region)
+        assert orig_m.time == fast_m.time
+        assert orig_m.match == fast_m.match
+        assert orig_m.region == fast_m.region
+        assert bool(orig_m) == bool(fast_m)
+        if orig_m:
+            assert orig_m.first_pass_result == pytest.approx(
+                fast_m.first_pass_result, abs=0.1)
```
drothlis: 0.1 seems quite high. What's the maximum difference you saw across the different implementations?

_stbt/match.py
```diff
@@ -1013,6 +1036,7 @@ def link(name, level=None, match=None):  # pylint: disable=redefined-outer-name
         template,
         ConfirmMethod=ConfirmMethod,
         link=link,
+        fast_path=imglog.data.get("fast_path"),
```
drothlis: The keyword args were sorted alphabetically before.

_stbt/sqdiff.c
```diff
@@ -0,0 +1,139 @@
...
+ * color_depth indicates the layout of t in memory:
+ *
+ *                        template layout       frame layout
+ * PIXEL_DEPTH_U8         U8                    U8
+ * PIXEL_DEPTH_BGR        BGR                   BGR
+ * PIXEL_DEPTH_BGRx       BGRx                  BGR
+ * PIXEL_DEPTH_BGRA       BGRA                  BGR
+ *
+ * t_stride and f_stride are the strides between lines measured in bytes for
+ * f and t respectively.
```
drothlis: You got the "f and t respectively" order backwards. :-P

tests/test_match.py
```diff
@@ -280,3 +280,53 @@ def test_png_with_16_bits_per_channel():
...
+        assert (orig_m.frame == fast_m.frame).all()
+        if isinstance(orig_m.image, numpy.ndarray):
+            assert (orig_m.image == fast_m.image).all()
+        else:
+            assert orig_m.image == fast_m.image
```
drothlis: Great test. It covers greyscale, BGR, and BGRA templates and leverages our existing stbt.match implementation as the oracle!

tests/test_match.py
wmanley: This is caused by the pyramid levels.  With pyramid levels disabled they match closely.  In this case `if orig_m` implies that we reached the bottom pyramid level.  Without `if orig_m` and and with `abs=0.01` we get a failure: `0.833 != 0.808` (a difference of 0.025).

I'll change the test to vary its strictness depending on whether it matched or not.

Makefile
wmanley: Fixed.

_stbt/match.py
wmanley: Fixed.

_stbt/sqdiff.c
wmanley: Eagle-eye: fixed.

PR #567, opened 2019-02-13
stbt.match: Invert the meaning of confirm_threshold

drothlis:
Now 0 means "less strict" and 1.0 means "completely strict". i.e. it's measuring the similarity, rather than the difference (error).

This is for consistency with the `match_threshold`. The discrepancy has caused some confusion in the past.

In all the user test-scripts that we've seen, confirm_threshold is never changed from its default values, except to set it to an absurdly high value (where the intent was really to disable the confirm pass; you can do that with `confirm_method=ConfirmMethod.NONE`).

### DIFF
--- a/_stbt/match.py
+++ b/_stbt/match.py
@@ -106,20 +106,20 @@ class MatchParameters(object):
         are a single solid color without any lines or variation).
 
         This is the default method, with a default ``confirm_threshold`` of
-        0.30.
+        0.70.
 
     :param float confirm_threshold:
-      The maximum allowed difference between any given pixel from the reference
-      image and its counterpart from the candidate region in the source video
-      frame, as a fraction of the pixel's total luminance range.
+      The minimum allowed similarity between any given pixel in the reference
+      image and the corresponding pixel in the source video frame, as a
+      fraction of the pixel's total luminance range.
 
       Unlike ``match_threshold``, this threshold applies to each pixel
       individually: Any pixel that exceeds this threshold will cause the match
       to fail (but see ``erode_passes`` below).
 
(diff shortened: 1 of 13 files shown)


stbt-match
```diff
@@ -40,7 +40,7 @@ parser.add_argument(
     "match_parameters", nargs="*",
     help="""Parameters for the image processing algorithm. See
         'MatchParameters' in the stbt API documentation. For example:
-        'confirm_threshold=0.20')""")
+        'confirm_threshold=0.70')""")
```
wmanley: These aren't equivalent, but it's just documentation.

_stbt/match.py
```diff
@@ -106,20 +106,20 @@ class MatchParameters(object):
...
       Unlike ``match_threshold``, this threshold applies to each pixel
       individually: Any pixel that exceeds this threshold will cause the match
       to fail (but see ``erode_passes`` below).
 
-      Valid values range from 0 (more strict) to 1.0 (less strict). Useful
-      values tend to be around 0.16 for ``ABSDIFF``, and 0.30 for
-      ``NORMED_ABSDIFF``. Defaults to 0.30.
+      Valid values range from 0 (less strict) to 1.0 (more strict). Useful
+      values tend to be around 0.84 for ``ABSDIFF``, and 0.70 for
+      ``NORMED_ABSDIFF``. Defaults to 0.70.
```
wmanley: As we don't publish documentation for old versions you probably want a "changed in v30" mention here.

wmanley:
Yep, this is a good usability improvement.

PR #569, opened 2019-02-14
stbt.match: Optimization: Merge regions-of-interest that overlap completely

drothlis:
Currently we're getting rois like this:

20-rois

Note that `cv2.findContours` with `cv2.RETR_EXTERNAL` is working correctly: those internal regions aren't internal contours, they're completely outside the large contour. It's only when we take the *bounding rectangle* of the large contour, that the smaller ones fall inside it. So we have to manually filter them out. I couldn't find a suitable OpenCV function to do it so I've implemented a simple quadratic loop.

After `_merge_regions` it looks like this:

9-rois

This is a quadratic implementation, but it shouldn't matter because we don't expect huge numbers of regions. On my laptop it takes:

| number of regions |   ms |
|-------------------|------|
|                20 | 0.06 |
|               200 | 0.6  |
|              2000 | 9    |

TODO:

- [x] Add unit test screenshot & template from #558?
- [x] Measure total performance impact on `stbt.match`.


_stbt/match.py
```diff
@@ -683,8 +684,8 @@ def _match_template(image, template, mask, method, roi_mask, level, imwrite):
             cv2.rectangle(
                 source_with_rois,
                 (max(0, r.x), max(0, r.y)),
-                (min(s.w - 1, r.x + r.w + t.w - 1),
-                 min(s.h - 1, r.y + r.h + t.h - 1)),
+                (min(s.w - 1, r.right + t.w - 1),
+                 min(s.h - 1, r.bottom + t.h - 1)),
```
wmanley: Right, this is the same as:

```python
d = roi.extend(right=template.shape[1], bottom=template.shape[0])
d = Region.intersect(d, _image_region(source_with_rois))
cv2.rectangle(source_with_rois, (d.x, d.y), (d.right - 1, d.bottom - 1),
              (0, 255, 255), thickness=1)
```

Maybe we should have our own draw_rectangle function that takes a `Region` encompassing the last two lines.

drothlis:
I ran it on one of my test-packs. There are some regressions but overall it seems a worthwhile improvement (average 12ms improvement across all 1100 template×screenshot pairs):

screenshot from 2019-02-14 14-14-55

The worst slowdown (see below) had **350 regions!**

| Reference image | Frame | ROIs identified at previous pyramid level | ROIs in the full sized image |
|---|---|---|---|
| template | source | level1-source_matchtemplate_threshold | level0-source_with_rois |

drothlis:
This was the best improvement: 18 regions that are redundant, because we already need to search the entire frame anyway.

level0-source_with_rois

PR #571, opened 2019-02-15
test_sqdiff_c_numpy_equivalence: Fix generating mask

drothlis:
`mask[mask & 1] == 255` sets *all* elements of the array to 255, unconditionally. Also, it fails if the height or width of the array is 1 (this is how I noticed): "IndexError: index 1 is out of bounds for axis 0 with size 1".

You need to index with an array of bools instead.

### DIFF
--- a/_stbt/sqdiff.py
+++ b/_stbt/sqdiff.py
@@ -96,7 +96,7 @@ def _random_template(size=(1280, 720)):
     tt = numpy.random.randint(0, 256, (tsize[1], tsize[0], 4),
                               dtype=numpy.uint8)
     mask = tt[:, :, 3]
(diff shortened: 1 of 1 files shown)


_stbt/sqdiff.py
```diff
@@ -96,7 +96,7 @@ def _random_template(size=(1280, 720)):
     tt = numpy.random.randint(0, 256, (tsize[1], tsize[0], 4),
                               dtype=numpy.uint8)
     mask = tt[:, :, 3]
-    mask[mask & 1] = 255
+    mask[mask & 1 == 1] = 255
```
wmanley: Good find.

PR #573, opened 2019-02-15
stbt.match optimisation: Don't downscale images < 47px wide

drothlis:
TODO:

- [ ] @wmanley to explain why this is a good idea.
- [ ] Performance measurements.


drothlis:
Not great. Maybe it needs to be only 47 in the x dimension.

screenshot from 2019-02-15 16-24-50

PR #575, opened 2019-02-26
core: Fix teardown error "not all arguments converted during string formatting"

drothlis:
This hides the actual warning message, and it causes the test to error
instead of passing.

Introduced in #563.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -735,8 +735,8 @@ class SinkPipeline(object):
             state = self.sink_pipeline.get_state(0)
             if (state[0] != Gst.StateChangeReturn.SUCCESS or
                     state[1] != Gst.State.PLAYING):
-                debug(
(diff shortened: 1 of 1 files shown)


drothlis:
Seen 32 / 152,000 runs of a short (no-op) test. Unlikely to affect anybody in practice.

_stbt/core.py
```diff
@@ -735,8 +735,8 @@ def __exit__(self, _1, _2, _3):
             state = self.sink_pipeline.get_state(0)
             if (state[0] != Gst.StateChangeReturn.SUCCESS or
                     state[1] != Gst.State.PLAYING):
-                debug(
-                    "teardown: Sink pipeline not in state PLAYING: %r" % state)
+                debug("teardown: Sink pipeline not in state PLAYING: %r"
+                      % (state,))
```
wmanley: Right, it was broken because `state` is a `tuple`.

PR #576, opened 2019-02-26
pylint: Warn on assert True and understand assert False

wmanley:


drothlis:
I've rebased onto master & resolved conflicts. Mainly that we had added other checkers in the meantime, so the new checker needed to use E7008 instead of E7006.

drothlis:
I also tweaked the text of the warning messages.

PR #578, opened 2019-03-04
press_and_wait: Fix regression

wmanley:
Since 75fd9feb65d `press_and_wait` has failed with the exception:

      ...
      File "test.py", line 7, in <module>
        assert stbt.press_and_wait("white")
      File "_stbt/transition.py", line 82, in press_and_wait
        press_result = _dut.press(key)
    AttributeError: 'NoneType' object has no attribute 'press'

This wasn't caught by our existing tests because each of them inject a
custom `dut` for testing and this issue was caused by not defaulting `dut`
correctly.

### DIFF
--- a/tests/test-stbt-py.sh
+++ b/tests/test-stbt-py.sh
@@ -646,3 +646,25 @@ test_that_press_returns_a_pressresult() {
 	EOF
     stbt run -vv --control=none test.py || fail "Incorrect press() behaviour"
 }
+
+test_press_and_wait() {
+    # This is a regression test - dut was incorrectly injected into
+    # press_and_wait causing failures that our pytest tests didn't find.
+    cat > test.py <<-EOF &&
+	import stbt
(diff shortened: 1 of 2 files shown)


tests/test-stbt-py.sh
```diff
@@ -646,3 +646,25 @@ test_that_press_returns_a_pressresult() {
...
+
+	stbt.press("black")
+	stbt.wait_until(stbt.is_screen_black)
+	
+	assert stbt.press_and_wait("white")
+	assert not stbt.press_and_wait("white", timeout_secs=1)
+	
+	stbt.press("ball")
+	stbt.wait_for_motion()
+	threading.Timer(1, lambda: stbt.press("black")).start()
```
drothlis: 0.1 is probably sufficient.

wmanley: Fixed

PR #580, opened 2019-03-14
stbt lint: Check `get_frame()` and `press()` in FrameObject properties

drothlis:


_stbt/pylint_plugin.py
```diff
@@ -87,6 +97,12 @@ def visit_call(self, node):
                     self.add_message('E7003', node=node, args=arg.as_string())
 
         if _in_frameobject(node) and _in_property(node):
+            if re.search(r"\bget_frame$", node.func.as_string()):
+                self.add_message('E7006', node=node)
+
+            if re.search(r"\b(press|press_and_wait)$", node.func.as_string()):
```
wmanley: You can add `press_until_match` here.

_stbt/pylint_plugin.py
```diff
@@ -87,6 +97,12 @@ def visit_call(self, node):
                     self.add_message('E7003', node=node, args=arg.as_string())
 
         if _in_frameobject(node) and _in_property(node):
+            if re.search(r"\bget_frame$", node.func.as_string()):
```
wmanley: We don't want to depend on side-effects at all here, so we can add:

* `stbt.wait_until`
* `stbt.wait_for_match`
* `stbt.wait_for_motion`
* `stbt.detect_motion`
* `stbt.audio.get_rms_volume`
* `stbt.audio.wait_for_volume_change`
* `stbt.audio.audio_chunks`
* `stbt.audio.play_audio_file`
* `stbt.frames`

_stbt/pylint_plugin.py
drothlis: Done (and `pressing` too).

_stbt/pylint_plugin.py
drothlis: I'll do this as a separate checker -- I'll do it in #581 so that I can get this one merged, because it's giving me a really weird test failure that I need to investigate further.

PR #586, opened 2019-05-08
Miscellaneous Python 3 prep

drothlis:
Commits extracted from #583, that should be mergeable now without impacting current Python 2 behaviour.


drothlis:
I have already published the changed circleci docker image to docker hub.

PR #591, opened 2019-05-14
stbt.match with transparency: False negative at lower pyramid levels

drothlis:
TODO:

- [x] Test reference image that's the same size as the frame. *I did this by commenting out the fast-path in the code.*
    - [x] with transparency.
- [x] Test reference image that matches at each edge & corner of the frame. *See the new test_match_region.*
    - [x] with transparency.

---

This reference image (where the white square in the middle is actually transparent):

roku-tile-selection

fails to match this frame:

roku-tile-frame

The match fails at the lowest pyramid level (2) with certainty of only 0.93 -- note the default threshold is 0.98 which is relaxed to 0.97 at the lower pyramid levels.

This bug only applies to MatchMethod.SQDIFF. With the other methods, we relax the threshold *so much* that we're unlikely to hit this bug. Relaxing the threshold by that much with SQDIFF would essentially make the pyramid optimisation a waste of time.


drothlis:
I think this happens because [cv2.pyrDown] blurs the image before downscaling it. With the reference image, the edge pixels stay white because there is nothing outside the image's edge to blur it with. But in the actual frame, the white edge gets blurred with the dark background, becoming a much darker shade of white.

Note that we call [cv2.pyrDown] with the default [borderType], which is `BORDER_REFLECT_101`.

If we change to `borderType=cv2.BORDER_REPLICATE` then the above example matches successfully, **but only if** we crop the reference image 1 or 2 pixels wider so that it includes a pixel of the background colour.

[cv2.pyrDown]: https://docs.opencv.org/3.2.0/d4/d86/group__imgproc__filter.html#gaf9bba239dfca11654cb7f50f889fc2ff
[borderType]: https://docs.opencv.org/3.2.0/d2/de8/group__core__array.html#ga209f2f4869e304c82d07739337eae7c5

drothlis:
In theory this affects images without transparency too; but in practice probably not because less pixels are affected, relative to the total number of pixels that are considered when calculating the average. We already insist that the image be >20px in both dimensions (i.e. >10px after downscaling). This only affects the pixels near the edges, so it's more likely to affect reference images with thin edges where everything else (the centre) is transparent.

drothlis:
NB. `BORDER_TRANSPARENT` and `BORDER_ISOLATED` aren't supported.

drothlis:
With this change (removing 1 row/column of pixels from the the edge of the reference image each time we scale down), the smallest pyramid level is now entirely transparent because we have removed all of the border. So it matches the whole frame. (Before it was 0.93).

At the middle pyramid level the sqdiff is now 0.9995 (vs. 0.9925).

drothlis:
With 0a4e6a1, the MatchResult.region that we report for a non-match is different by a few pixels (if the non-match happened at a smaller pyramid level so we didn't bother looking for a match at the full size).

If you've specified a region to search in, the non-match region we return might even be slightly outside the input region (because it's off by one at the smaller pyramid level, then we upscale by multiplying it by 2). This could be confusing. I don't think it really matters though, because it's only for non-matches. And I don't think there's anything we can do about it.

PR #592, opened 2019-05-17
added support for color recognition

msiden:
This is what I had in mind for color recognition. 

I have added two new functions: stbt.get_color() and stbt.is_color(). Both take a region argument and is_color also needs color_range which is a tuple describing the lowest and highest expected color in that region. I think the implementation is very general.

Some use-cases taken from my project:
* Verifying that the background changes to white when entering the Kids section (The Main section is black-themed)
* Verifying whether a text box is empty or not. I found OCR to be unreliable here since it would often return blank when there were only two characters entered
* Verifying that menu items are selected/highlighted. In this case I found color recognition more reliable and also more maintainable than image recognition since it doesn't require any screenshots

### DIFF
--- a/_stbt/color.py
+++ b/_stbt/color.py
@@ -1,0 +1,41 @@
+# coding: utf-8
+
+
+def get_color(region, frame=None):
+    """Get the average color of a region."""
+
+    import stbt
+
+    # Grab a frame if not provided by caller
+    frame = stbt.get_frame() if frame is None else frame
+
+    # Extract the requested region
+    frame = stbt.crop(frame.copy(), region)
+
+    # Convert the frame to a list and flatten it
+    frame = [item for sublist in frame.tolist() for item in sublist]
+
+    # Get total number of pixels
+    number_of_pixels = len(frame)
+
+    # Calculate the average color
+    blue = green = red = 0
+    for pixel in frame:
+        blue += pixel[0]
+        green += pixel[1]
+        red += pixel[2]
+
+    return blue/number_of_pixels, green/number_of_pixels, red/number_of_pixels
+
+
+def is_color(region, color_range, frame=None):
+    """
+    Check if the color in a certain part of the screen is within a given range.
+
+    color_range is a tuple with the lowest and highest color values in the range. BGR format.
+    Example: ((0, 0, 0), (50, 50, 50))
+
+    """
+    return all([
+        (c >= color_range[0][i] and c <= color_range[1][i])
+        for i, c in enumerate(get_color(region=region, frame=frame))])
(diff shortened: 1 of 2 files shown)


_stbt/color.py
```diff
@@ -0,0 +1,41 @@
...
+def get_color(region, frame=None):
+    """Get the average color of a region."""
+
+    import stbt
+
+    # Grab a frame if not provided by caller
+    frame = stbt.get_frame() if frame is None else frame
+
+    # Extract the requested region
+    frame = stbt.crop(frame.copy(), region)
```
drothlis: There's no need to copy the frame here as we don't modify it.

_stbt/color.py
```diff
@@ -0,0 +1,41 @@
...
+    number_of_pixels = len(frame)
+
+    # Calculate the average color
+    blue = green = red = 0
+    for pixel in frame:
+        blue += pixel[0]
+        green += pixel[1]
+        red += pixel[2]
+
+    return blue/number_of_pixels, green/number_of_pixels, red/number_of_pixels
```
drothlis: Looping over each pixel in Python is going to be (relatively) slow. Using numpy like this will be much faster:

```
frame = stbt.crop(frame, region)
return numpy.mean(frame, axis=(0, 1))
```

_stbt/color.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def is_color(region, color_range, frame=None):
```
drothlis: I like that you give a minimum & maximum color. I think that we made a mistake in `stbt.ocr` by making `text_color` and `text_color_threshold`  parameters (i.e. `min = text_color - text_color_threshold` and `max = text_color + text_color_threshold`). Specifying the min & max directly is much easier; in practice I know the min & max that I want, and I have to calculate the average in order to specify `text_color` and `text_color_threshold`.

_stbt/color.py
drothlis: P.S. `crop` returns a view into the original frame's data, it doesn't actually modify the frame.

msiden:
@drothlis Actually I have found that returning the avarage can be accurate in the scenario you described (the black tab where the selection is  marked by changing text color) if you define a tightly cropped region. In this case a different avarage would be returned depending on whether the text is selected or not.

drothlis:
@msiden quite right: searching for the average colour value can be reliable. What I meant (and I wasn't very clear about this) is that the colour value in your test script is less obviously related to the UI you're testing:

* You can't tell what colour value to put in your test script just by looking at the image with a colour picker.
* If the test fails it's less obvious in what way the colour changed.
* If you need to change the size of the region (for whatever reason) then you need to calculate a new colour value. etc.

msiden:
@drothlis Those are fair points. Now I see how it might be preferable to return the dominant color rather than the avarage. I don't have any experience with that particular solution but I guess it can't be that much more complicated than this one.

PR #593, opened 2019-05-21
Remove "source_teardown_eos" workaround for old decklinksrc teardown hang

drothlis:
Installing gst-python overrides the auto-generated GObject Introspection
bindings, among other things adding the `__iter__` method on top of the
`Gst.Iterator` object to fix native python iteration.[1]

[1]: https://cgit.freedesktop.org/gstreamer/gst-python/tree/gi/overrides/Gst.py#n245

We implemented this ourselves back in March 2014, before gst-python was
widely available. It has been in Ubuntu (as python-gst-1.0) since 16.04
at least.


drothlis:
This passes CI with our old CI docker images (before we install python-gst) -- presumably none of our tests actually exercise the part where we were calling `gst_iterate`, which was only in one place: During teardown, if config is `source_teardown_eos` is set to True. It's to enable a workaround for a decklinksrc bug, apparently:

> #### v28
> 
> 24 January 2018.
>
> * Removed workaround for 4 year old deadlock bug in decklinksrc. If this is still necessary, set `source_teardown_eos = True` in the `[global]` section of your stbt config file, and let us know on the mailing list as we may remove the workaround completely in a future release.

But then, if we do install python-gst (see the second commit) pylint has some weird complaints. We'd need to investigate if those are real problems or if they need to be addressed. It feels risky.

drothlis:
Rather than investigate those pylint warnings, I have removed support for the `source_teardown_eos` config setting. This was the only user of `gst_iterate`. For all we know that decklinksrc bug doesn't exist anymore. We don't have the hardware to test it, and no-one has reported the need for the workaround since v28 was released over 1 year ago.

PR #594, opened 2019-05-21
Python3 support (unit tests)

drothlis:
In this pull request I'm aiming to get all the unit tests passing on both Python 2 & 3. We can leave the integration tests to a later pull request, so that we can "bank" this progress.

TODO:

- [ ] Make `Xxhash.update` only accept bytes (dropping ebc02abc59c0f347fb7b2691813a805bac623fc8)


drothlis:
We're already down to 5 / 267 failing tests (from 26).

drothlis:
Down to 2 test failures (`test_stbt_control_relay` and `test_stbt_control_relay_with_socket_passing`) caused by bugs in LircControl. I think what I'll do is write some integration tests using real lirc hardware, fix LircControl, and then if necessary fix fake-lircd to get the other lirc integration tests passing again.

drothlis:
@wmanley more Python 3 changes for you to review. Only 28 commits this time! :-)

_stbt/gst_hacks.py
```diff
@@ -112,7 +112,7 @@ def test_map_sample_reading_data():
 
     s = Gst.Sample.new(Gst.Buffer.new_wrapped(b"hello"), None, None, None)
     with map_gst_sample(s, Gst.MapFlags.READ) as a:
-        assert b'hello' == b''.join(chr(x) for x in a)
+        assert b'hello' == ''.join(chr(x) for x in a).encode("ascii")
```
wmanley: What is the type of `a` here I wonder?  Ah: It'll be `ctypes.c_byte * 5` probably.

An alternative here would be:

    [ord(x) for x in 'hello] == a[:]

which is just as unreadable.

_stbt/config.py
```diff
@@ -138,13 +138,13 @@ def _to_enum(type_, value, section, key):
 
 
 @contextmanager
-def _sponge(filename):
+def _sponge(filename, mode="w"):
```
wmanley: This makes `sponge` consistent with `open` I guess.

_stbt/frameobject.py
```diff
@@ -197,12 +197,30 @@ def __bool__(self):
...
+    def __le__(self, other):
+        return self.__cmp__(other) <= 0
+
+    def __gt__(self, other):
+        return self.__cmp__(other) > 0
+
+    def __ge__(self, other):
+        return self.__cmp__(other) >= 0
+
+    def __cmp__(self, other):
```
wmanley: `functools.total_ordering` can help with this too.  Then all you need to define is `__eq__` and `__lt__` and it will define the rest.  This is fine though.

Example:

```
@functools.total_ordering
class Test(object):
    def __init__(self, a):
        self.a = a
    def __eq__(self, other):
        return self.a == other.a
    def __lt__(self, other):
        return self.a < other.a
```

stbt_control_relay.py
```diff
@@ -68,47 +69,45 @@ def main(argv):
...
-                elif action == "SEND_START":
+                elif action == b"SEND_START":
                     control.keydown(key)
-                elif action == "SEND_STOP":
+                elif action == b"SEND_STOP":
                     control.keyup(key)
             except Exception as e:  # pylint: disable=broad-except
                 debug("Error pressing key %r: %r" % (key, e))
-                send_response(conn, cmd, success=False, data=str(e))
+                send_response(conn, cmd, success=False, data=to_bytes(str(e)))
```
wmanley: On Python 2 this will call

    e.__unicode__().encode('utf-8')

On Python 3 will be

    e.__str__().encode('utf-8')

Looks fine.

_stbt/utils.py
```diff
@@ -80,7 +80,7 @@ def find_import_name(filename):
 
 def to_bytes(text):
     if isinstance(text, str):
-        return text.encode("utf-8")
+        return text.encode("utf-8", errors="backslashreplace")
```
wmanley: I don't think this can fail?  All unicode sequences should be representable as utf-8 unless I'm missing something?

tests/test-stbt-run.sh
```diff
@@ -218,55 +218,107 @@ test_that_relative_imports_work_when_stbt_run_runs_a_specific_function() {
...
+		AssertionError
+		EOF
+        *) fail "check_unicode_error isn't implemented for $scenario" ;;
+    esac |
+    while IFS='' read line; do
+        grep -q -e "^$line" $output || {
+            cat $output;
+            fail "Didn't find line: «$line»";
+        }
+    done
```
wmanley: Nice tests.

_stbt/ocr.py
```diff
@@ -250,7 +250,7 @@ def ocr(frame=None, region=Region.ALL,
     if isinstance(tesseract_user_patterns, (bytes, str)):
         tesseract_user_patterns = [tesseract_user_patterns]
 
-    imglog = ImageLogger("ocr")
+    imglog = ImageLogger("ocr", result=None)
```
wmanley: Would doing `imglog.set(result=text)` make sense in this instance?

_stbt/ocr.py
```diff
@@ -250,7 +250,7 @@ def ocr(frame=None, region=Region.ALL,
...
 
     text, region = _tesseract(
```
wmanley: Hmm, shouldn't `_tesseract` be returning `bytes`?.  IIRC the xml parser expects to receive bytes because XML can define the encoding inline.

stbt_control_relay.py
```diff
@@ -33,13 +33,13 @@
...
 from builtins import *  # pylint:disable=redefined-builtin,unused-wildcard-import,wildcard-import,wrong-import-order
 import argparse
+import logging
 import os
 import re
 import signal
 import socket
 import sys
 
-import _stbt.logging
```
wmanley: Yes, this is sensible.  We can think about doing this more widely in the future.

tests/test_stbt_control_relay.py
```diff
@@ -104,6 +105,8 @@ def preexec_fn():
         os.environ['LISTEN_PID'] = str(os.getpid())
         if fd != 3:
             os.dup2(fd, 3)
+        if sys.version_info.major > 2:
+            os.set_inheritable(3, True)  # pylint:disable=no-member
```
wmanley: I wonder why they didn't backport this in `future`?  It should just call through to set `O_CLOEXEC`.

_stbt/utils.py
drothlis: I'm not sure. What's the point of having `errors` at all on `encode()`?

_stbt/ocr.py
drothlis: The whole debug template is a bit ad-hoc, it isn't super consistent. The text goes in `output` (including the hocr text from `match_text`), and `result` is only used to draw a region.

_stbt/ocr.py
drothlis: It works, I suppose because the xml library supports taking a str, and tesseract is always outputting UTF-8.

_stbt/xxhash.py
```diff
@@ -64,6 +64,8 @@ def __del__(self):
         _libxxhash.XXH64_freeState(self._state)
 
     def update(self, data):
+        if isinstance(data, str):
+            data = data.encode("utf-8")
```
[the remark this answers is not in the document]
drothlis: Fixed.

_stbt/imgproc_cache.py
```diff
@@ -211,7 +211,7 @@ def default(self, o):  # pylint:disable=method-hidden
             from _stbt.xxhash import Xxhash64
             h = Xxhash64()
             h.update(numpy.ascontiguousarray(o).data)
-            return (o.shape, h.hexdigest())
+            return (o.shape, h.hexdigest().decode("utf-8"))
```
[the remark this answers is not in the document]
drothlis: Fixed.

drothlis:
I've fixed xxhash. I also had to cherry-pick the CI docker image  (and `extra/pylint.sh`) changes from #597 because I have already published those docker images so there's no `pep8` in the python2 image.

_stbt/frameobject.py
drothlis: I think I didn't used `@functools.total_ordering` because I thought I was working on the MetaClass and I wasn't sure if the decorator would work in that case. Looking at it again now it turns out I was changing the actual `FrameObject` class. I'll leave it as is though.

_stbt/utils.py
wmanley: For encodings where the whole of unicode can't be represented.  e.g. ASCII

```
$ ipython3
In [2]: 'Röthlis'.encode('ASCII')
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-2-10281243efbf> in <module>()
----> 1 'Röthlis'.encode('ASCII')

UnicodeEncodeError: 'ascii' codec can't encode character '\xf6' in position 1: ordinal not in range(128)

In [3]: 'Röthlis'.encode('ASCII', errors="backslashreplace")
Out[3]: b'R\\xf6thlis'
```

PR #596, opened 2019-05-22
Remove "restart_source" config setting

drothlis:
This was a workaround for a bug in the Hauppauge HDPVR video-capture
device, which is ancient and unreliable hardware. As far as I know,
nobody uses this setting.

Note that we *do* restart the source pipeline if we receive EOS (always,
regardless of configuration). This setting also restarted the source
pipeline on underrun (no buffers received, but without an EOS either).

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -112,8 +112,6 @@ def new_device_under_test_from_config(parsed_args=None):
         args.control = get_config('global', 'control')
     if args.save_video is None:
         args.save_video = False
-    if args.restart_source is None:
-        args.restart_source = get_config('global', 'restart_source', type_=bool)
 
     display = [None]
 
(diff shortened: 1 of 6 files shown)


wmanley:
> Note that we do restart the source pipeline if we receive EOS (always, regardless of configuration).

Why do we do this?

wmanley:
Yes, it's nice to remove this workaround.  Makes some complex code a little easier to reason about.

drothlis:
> > Note that we do restart the source pipeline if we receive EOS (always, regardless of configuration).
> 
> Why do we do this?

To handle VidiU's H.264 encoder restart:  6b7ba0a313213f7742dbcefcd0564d9b9dc74ca4.

That in itself might be worth revisiting. As far as I know the only VidiU user (YouView) had some kind of repeater / HDMI doctor in between the STB and the VidiU, which provided a continuous signal even when the STB rebooted or changed mode, thus avoiding this behaviour from the VidiU. IIRC.

Of course we've had this behaviour so long that maybe it gets used for other hardware too and we don't even realise it.

drothlis:
> Note that we do restart the source pipeline if we receive EOS (always, regardless of configuration).

Now we have removed this behaviour too: #715.

PR #598, opened 2019-05-28
Python 3 packaging

drothlis:
The main change here is that `make install` modifies the shebang lines upon installation, from `#!/usr/bin/python` to `#!/usr/bin/python2.7` or `#!/usr/bin/python3` depending on the value of `$python_version`.

The rest of the changes are just test infrastructure to cope with that, plus other test fixes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ INSTALL_PYLIB_FILES = \
     stbt/__init__.py \
     stbt/android.py
 
-INSTALL_CORE_FILES = \
+INSTALL_CORE_SCRIPTS = \
     stbt_config.py \
     stbt_control.py \
     stbt_lint.py \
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -123,7 +123,9 @@ install-core: all
 	$(INSTALL) -m 0644 etc/stbt.conf \
 	    $(DESTDIR)$(sysconfdir)/stbt/stbt.conf
 	for filename in $(INSTALL_CORE_SCRIPTS); do \
-	    $(INSTALL) -m 0755 $$filename $(DESTDIR)$(libexecdir)/stbt/$$filename; \
+	    sed '1s,^#!/usr/bin/python\b,#!/usr/bin/python$(python_version),' \
+	        $$filename > $(DESTDIR)$(libexecdir)/stbt/$$filename; \
+	    chmod 0755 $(DESTDIR)$(libexecdir)/stbt/$$filename; \
```
wmanley: An alternative would be to modify the stbt shell script that we already modify on installation.

drothlis: I tried, but we also have to teach `stbt-screenshot` which python to use when it calls `stbt_run.py`.... which isn't unfixable but I've run out of steam.

PR #602, opened 2019-07-05
stbt.FrameObject repr: Limit float property values to 3 decimal places

drothlis:
Otherwise doctest output of floating-point values might vary from run to
run.

60 frames per second is 17ms per frame, so a 1ms precision should be
enough for any measurements related to video-frames. For example we have
been affected by this in a FrameObject that we use in our A/V sync
tests, that has a property showing the lag or lead time from the audio
to the corresponding video frame.

We only apply this to properties that have "time" or "_secs" in the
name, because we don't know what other use-cases people are using
floating-point values for. If you need this, at least now you have the
option of working "time" or "_secs" into your property name.

This won't help properties that are lists of floats, etc. But it's a
start.

### DIFF
--- a/_stbt/frameobject.py
+++ b/_stbt/frameobject.py
@@ -180,8 +180,13 @@ class FrameObject(with_metaclass(_FrameObjectMeta, object)):
         """
         The object's string representation includes all its public properties.
         """
-        args = ", ".join(("%s=%r" % x) for x in self._iter_fields())
-        return "%s(%s)" % (self.__class__.__name__, args)
+        args = []
+        for name, value in self._iter_fields():
+            if isinstance(value, float) and ("time" in name or "_secs" in name):
+                args.append("%s=%.3f" % (name, value))
+            else:
+                args.append("%s=%r" % (name, value))
+        return "%s(%s)" % (self.__class__.__name__, ", ".join(args))
 
     def _iter_fields(self):
         if self:
--- a/tests/test_frameobject.py
+++ b/tests/test_frameobject.py
@@ -55,7 +55,7 @@ class FrameObjectWithProperties(stbt.FrameObject):
 
     >>> frame = _load_frame("with-dialog")
     >>> FrameObjectWithProperties(frame)
-    FrameObjectWithProperties(is_visible=True, public=5)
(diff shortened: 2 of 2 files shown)


drothlis:
I tried solving this problem with [numtest], but:

* You have to put `import numtest` at the top level of the file. This isn't suitable for doctests in production files as you don't want to install numtest in the production environment.
* It doesn't work with pytest:
    ```
        def check_output(self, want, got, optionflags):
            res = doctest.OutputChecker.check_output(self, want, got,
    >                                                optionflags)
    E       TypeError: unbound method check_output() must be called with NumTestOutputChecker instance as first argument (got LiteralsOutputChecker instance instead)
    
    /usr/lib/python2.7/dist-packages/_pytest/doctest.py:279: TypeError
    ```
* It doesn't work with Python 3 (it's available via pip3, but doesn't work). Last commit was 3 years ago.

numtest's [implementation] is < 200 lines of code. It would be nice to implement this in pytest itself, so that you can enable it globally in pytest.ini instead of adding `# doctest: +NUMBER` to each line. numtest's license is MIT (same as pytest).

[numtest]: https://github.com/boisgera/numtest
[implementation]: https://github.com/boisgera/numtest/blob/master/numtest/__init__.py

drothlis:
P.S. numtest only applies to doctests written manually; we'd still need something like this for the doctests generated by `stbt auto-selftest`.

drothlis:
@wmanley what do you think of this idea?

drothlis:
Note: Even if you do something like this in your FrameObject:

```
@property
def xyz(self):
    return round(..., 3)
```

You can still get something like `0.23300000000000001`.

drothlis:
> It would be nice to implement [numtest] in pytest itself

I have implemented https://github.com/pytest-dev/pytest/pull/5576

wmanley:
> Otherwise doctest output of floating-point values might vary from run to run.

I don't think this is true.  All the floating-point operations we perform should be deterministic.  The real issue solved here is that doctest output varies between Python versions.

drothlis:
I'm not going to merge this, it'll be too magical. Just the thought of having to document when it does & doesn't apply makes me think it's a bad idea.

PR #604, opened 2019-07-15
core: Allow changing the `sink_pipeline` on the fly

wmanley:
This is useful as a librification measure.  This allows us to separate the construction of `Display` (the source pipeline) from `SinkPipeline`. Now they can be constructed and initialised separately.


_stbt/core.py
```diff
@@ -159,8 +159,22 @@ def __exit__(self, exc_type, exc_value, tb):
...
+        assert isinstance(self._sink_pipeline, NoSinkPipeline)
+        if self._running:
+            new_sink_pipeline.__enter__()
+        try:
+            self._sink_pipeline = new_sink_pipeline
+            self._display.set_sink_pipeline(new_sink_pipeline)
+        finally:
+            if self._running:
+                old_sink_pipeline.exit_prep()
+                old_sink_pipeline.__exit__(None, None, None)
```
wmanley: Really this is only safe if `old_sink_pipeline` is `NoSinkPipeline`.  `Display.on_new_sample` calls `sink_pipeline.on_sample`, and may call it on the old sink pipeline after or during `old_sink_pipeline.__exit__`.  In practice I don't think this will be a problem.

wmanley:
Closing, I don't plan on working on this further.

PR #605, opened 2019-07-16
stbt.press_and_wait: Ignore pixel intensity differences <= 5

drothlis:
On the Apple TV, successive frames differ by up to 3 (out of 255) on almost every pixel, in a funky multicoloured moire pattern. Here's what it looks like after enhancing the levels (remember the actual differences are <= 3 so you can't see them with the naked eye):

small_diffs_enhanced

### DIFF
--- a/_stbt/transition.py
+++ b/_stbt/transition.py
@@ -246,10 +246,11 @@ def strict_diff(prev, frame, region, mask_image):
         _ddebug("found %s diffs above 20 (max %s) in %r", frame,
                 numpy.count_nonzero(big_diffs), maxdiff, out_region)
     elif maxdiff > 0:
-        small_diffs_count = numpy.count_nonzero(absdiff)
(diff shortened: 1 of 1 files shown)


drothlis:
This will have the same problem that `wait_for_motion` currently has: A slow fade-in or fade-out won't be detected.

With `press_and_wait`, detecting the start of the transition doesn't have a problem because we always check the frame just before the key-press, against the latest frame.
But once the animation has started, when checking for the animation to end we do compare successive frames against each other, so it could be affected by this slow-fade-out problem.

I don't know if this will actually be a problem with `press_and_wait`, as it's typically used for transitions between menu items, which are more rapid / obvious animations.

wmanley:
I wonder if this is caused by some sort of temporal dithering, perhaps in YUV to RGB conversion or vice-versa.

Merging

PR #606, opened 2019-07-16
wait_for_motion: Do better at detecting slow motion

wmanley:
`wait_for_motion` has a `noise_threshold` intended to exclude false
positives when looking for motion.  By default if any pixel changes in
intensity by less than 40 per frame (e.g. takes longer than 250ms to go
from black to white) motion will not be detected.

This commit fixes this issue by not strictly looking at consecutive frames,
instead it will compare the current frame to the last one where motion was
seen.  This allows differences to accumulate over time.  In theory noise
will be decorrelated with time so there will be no accumulation of error,
whereas actual motion will be correlated with time, so we will detect the
motion.

Here's a comparison of the old and new implementations.  In both cases a
difference of above 40 will be considered motion.  The old impl never
fires, the new one fires on every other frame

    pixel value:    0    40    80    120    160    200    240    255
    old impl diffs:   40    40    40     40     40     40     15
    new impl diffs:   40    80*   40     80*    40     80*    15

This may make the implementation more difficult to explain, but it should
make the behaviour of the function behave more in line with expectations.

If this implementation is found to work well in practice it might make
sense to revisit the default parameter values.

I've included a test demonstrating the pathological case for the old
implementation.  It looks like this:

gradient_wipe

This was inspired by reviewing #605.  I haven't finished reviewing it yet though (sorry @drothlis).

### DIFF
--- a/_stbt/motion.py
+++ b/_stbt/motion.py
@@ -22,8 +22,7 @@ def detect_motion(timeout_secs=10, noise_threshold=None, mask=None,
     """Generator that yields a sequence of one `MotionResult` for each frame
     processed from the device-under-test's video stream.
 
-    The `MotionResult` indicates whether any motion was detected -- that is,
-    any difference between two consecutive frames.
+    The `MotionResult` indicates whether any motion was detected.
 
     Use it in a ``for`` loop like this::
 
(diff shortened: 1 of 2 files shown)


wmanley:
An alternative implementation would be to compare frames a set amount of time apart.  Like compare frame 2 to 4 and 3 to 5, etc.  This would also increase sensitivity to real motion without increasing the sensitivity to noise.  There are different trade-offs here that I can go into later.

Issue #590, opened 2019-05-13
Color recognition


msiden:
I always thought it would be great if stbt could identify what color is displayed in a particular region of the screen. I added this functionality to my own stbt-based framework a while back and it's been very useful and a great alternative in scenarios where image recognition can be unreliable in my experience. 

Use cases:
* Verify that the background or a specific region changes color when you enter certain menus or screens
* Verify that menu text changes color when selected

This function works as follow:
* Grab a frame from the device under test
* Extract the specified region of the frame. This can be as small as one pixel. It's recommended to keep regions as small as possible for maximum accuracy and speed
* Check the BGR values for each pixel and divide by the total number of pixels. In other words, it "blends" all pixels and returns the average.
* Return result as a tuple (B, G, R)

This would result in two new methods in stbt
* stbt.get_color - Takes a region argument and returns a tuple as described above
* stbt.is_color - Takes two arguments, region and color_range which is a tuple containing the lowest and highest allowed values for the expected color ((0, 0, 0), (70, 70, 70)). Returns a boolean.

This is only a few lines of code that I would be happy to share with the community if deemed interesting.

drothlis:
That sounds useful. One specific use-case is for UIs where the selection is shown by changing the colour of the button or menu entry (instead of drawing a box around it).

We've implemented similar things in our test-packs -- I think we were looking for the dominant colour (the most frequent colour in the region) rather than the average. Or sometimes the brightest colour. It depends on what your UI looks like and what you're trying to accomplish.

A similar thing we've done is to detect a spinner like this:

![Screenshot from 2019-05-17 10-20-37](https://user-images.githubusercontent.com/12424/57917948-ba21a600-788d-11e9-8d75-b427fce3063f.png)

The 4 blobs spin around so you can't use `stbt.match` to find it. The background can show live TV, not just black. What we did was check that all 4 colours were present in that region, with at least some number of pixels for each particular colour. This worked very reliably (and fast).

If you share some examples of your use-cases, we might be able to find a solution that's general. In any case, please do share your code -- even if it isn't general enough to merge, it can serve as an example for other people to use in their test-packs.

msiden:
I opened a pull request https://github.com/stb-tester/stb-tester/pull/592
I think that should give you a good idea of what I had in mind.

PR #612, opened 2019-08-09
[WIP]Test on UI elements by color range

JNan86:
### Challenges
I would like to share some functions to test on UI elements by color range which I've been using for a long time in our test pack. I implemented the functions due to the following challenges:

I wanted to match a red icon which appears when recording is enabled on a channel from planner. We needed to test the same software on two different STBs, the UI layout were exactly the same but there is slight color difference which is enough to cause the image matching to fail. As I only wanted to quickly check whether the icon pops up when the recording is started, so I wrote the method `get_mask_by_color_range` to filter the frame to get a mask, then wrote another method `get_color_existence` to check whether the color is existed. In this way, I could make my test code generic for both STBs.

Later on in another project, we needed to determine the selected tab on a top level menu. The tabs do not have the same size and the size depends on how long the text is. The color of a tab will change when the cursor is on it. In this case, it was too verbose to match by the tab images, so again I used the method `get_mask_by_color_range` to get a mask highlighting the tab selected, and based on that, I wrote the method `get_colored_box_region` to find the bounding region of the selected tab then use OCR to read the content of the tab.

### List of class and functions implemented
+ `stbt.Color` the class to represent a specific color, defaults to be instantiated by HSV but can be parsed from BGR
+ `stbt.get_mask_by_color_range` the method to get a mask highlighting the area where the range of color exists
+ `stbt.get_color_existence` (I have not yet migrated this function) the method to determine whether the specified color
is existed in a given `stbt.Region`
+ `stbt.get_colored_box_region` the method to find the bounding `stbt.Region` of the colored area

Please check the docstrings for more details.

### Todo
- [ ] To migrate `get_color_existence`
- [ ] To implement self-test
- [ ] To discuss whether there is a need to wrap the matching results into something like `stbt.MatchResult`
- [ ] To pass `make check`

### Things to note
+ HSV was used to easier define a range of color.
+ In terms of performance, I'm not sure whether all methods can finish processing within 1/25 second on all platforms but for functional test requirements, this is not a big deal. Matching by color is generally faster and more reliable.

It seems like color matching/recognition is a common requirement for test development. I'm not too sure whether we are trying to solve the same problem? #592

### DIFF
--- a/_stbt/color.py
+++ b/_stbt/color.py
@@ -100,3 +100,53 @@ def get_mask_by_color_range(hsv_starting_boundary, hsv_ending_boundary, frame=No
         cv2.rectangle(roi, (region.x, region.y), (region.right, region.bottom), color=255, thickness=cv2.FILLED)
         mask = cv2.bitwise_and(mask, mask, mask=roi)
     return mask
+
+
+def get_colored_box_region(hsv_starting_boundary, hsv_ending_boundary, outer_area=stbt.Region.ALL, inner_area=0,
+                           frame=None, region=None):
+    """
+    The method to determine the bounding box of a colored region, very useful when searching for a colored rectangle
+    or round-cornered rectangle (e.g. a tab, text field in a box or keyboard keys) from the screen, especially when
+    you find it hard to match the rectangle by images (e.g. the tab size/content are variable, or there are too many
+    different tabs). After the bounding region is found, you can carry out OCR or further image processing within the
+    region. To ensure the best matching accuracy, outer_area and inner_area must be specified, otherwise any possible
+    contour filled with the color specified will be returned. To leave some allowance between the outer_area and
+    inner_area helps you to better handle a range of possible box size or tab length.
+
+        +--- outer_area ----------------------+
+        | ####################################|
+        | #+--- inner_area -----------------+#|
+        | #|################################|#|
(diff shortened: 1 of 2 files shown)


msiden:
This does indeed sound quite similar to what I proposed in #592. I think my proposal may be more basic. As for the get_colored_box_region method I am doing similar things in my test pack and can confirm that it is a very elegant and powerful feature.

drothlis:
Hi @JNan86! Thanks for your contribution, it looks like useful stuff.

I haven't reviewed the code or the API yet, but could you please attach some screenshots to issue #590, of the kind of things that you need to detect? You can crop them to remove sensitive information. I've already attached one example. We will need these to (1) design an API that meets all (most) colour-related needs, and (2) for unit tests.

@msiden you too if you can.

drothlis:
> The color of a tab will change when the cursor is on it. [...] `get_colored_box_region` to find the bounding region of the selected tab then use OCR to read the content of the tab.

See also `find_boxes` and `find_solid_coloured_boxes` from pull request #519 for a different approach (using edge detection) for finding a box (rectangle) by colour.

wmanley:
Closing, we have this now: https://stb-tester.com/manual/python-api#stbt.find_regions_by_color

PR #613, opened 2019-08-12
Add stbt.Grid: Helper for converting pixel coordinates <-> grid coordinates

drothlis:


stbt/grid.py
```diff
@@ -0,0 +1,146 @@
...
+Appearance
+==========
+
+Regularly spaced elements.
+
+Variable number of items.
+
+Laid out in order.
+
+Empty cells at the end, but no empty cells in the middle.
```
wmanley: I don't think this comment makes sense for the implementation below.  Probably should be stripped out.

drothlis: Done.

drothlis:
I'm finding more & more things I need to fix with Keyboard so I've merged the outstanding related PRs and we can address any review issues post-merge.

PR #614, opened 2019-08-15
stbt.Keyboard: Reverse order of arguments to enter_text & navigate_to

drothlis:
This API is a little bit hard to use: In the example from the docstring, we were passing an out-of-date page object into `Keyboard.navigate_to`. So after typing the text, it was calculating the path to the "SEARCH" button from your initial state, not from the actual state (which should be the last letter in the text you've typed). This could lead it to pressing a sequence of keys that take it out of the keyboard altogether.

This fixes the example in the docstring, and we also reverse the order of arguments to give us a bit more flexibility in how we might solve this in the future without breaking the Keyboard API.

### DIFF
--- a/stbt/keyboard.py
+++ b/stbt/keyboard.py
@@ -155,8 +155,9 @@ class Keyboard(object):
                     ...  # implementation not shown
 
                 def enter_text(self, text):
-                    self._kb.enter_text(page=self, text=text.upper())
-                    self._kb.navigate_to(page=self, target="SEARCH")
+                    page = self
+                    page = self._kb.enter_text(page, text.upper())
+                    self._kb.navigate_to(page, "SEARCH")
                     stbt.press("KEY_OK")
         """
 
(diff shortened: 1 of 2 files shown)


drothlis:
This shows how difficult it is to use the API correctly.

To fix this type of bug, `stbt.Keyboard` would need to call `page.refresh()` at the very beginning of `enter_text` and `navigate_to`.

We could also change it back so that it takes the PageObject in the Keyboard contructor.
* It could take a PageObject class instead of an instance (but that would only work for PageObjects that don't take any parameters in `__init__`).
* If it takes a PageObject instance, then you can't create the `Keyboard` instance at module/class level, you'd have to do it at run-time.

@wmanley your thoughts?

drothlis:
Another idea: (getting more and more magical here...)

`stbt.press` could record the last keypress sent.

Then the `Keyboard` methods can check the given `page`, and if its timestamp is older than the last keypress, call `refresh` on it. (This wouldn't handle other reasons that the page object could be out of date: interacting with the DUT via some other mechanism that isn't `stbt.press`; or the screensaver kicking in; etc).

In fact we already store `DeviceUnderTest._last_keypress`, though we don't do anything with it yet. What where we thinking of when we added it?

drothlis:
I'm going to merge this as the actual change here is fairly uncontroversial. We can discuss the greater issue here post-merge, or in a separate ticket.

PR #617, opened 2019-08-20
Keyboard & Grid fixes

drothlis:
Provides a much easier way of specifying grid-like keyboards. See `test_composing_complex_keyboards` in 1df3d7b for an example.

### DIFF
--- a/_stbt/grid.py
+++ b/_stbt/grid.py
@@ -138,7 +138,7 @@ class Grid(object):
                               else self._position_to_region(position))
                     break
             else:
-                raise IndexError("data '%r' not found" % data)
+                raise IndexError("data %r not found" % (data,))
 
         return Grid.Cell(
             index,
--- a/_stbt/grid.py
+++ b/_stbt/grid.py
@@ -161,7 +161,11 @@ class Grid(object):
             return self.get(data=key)
 
     def __iter__(self):
-        return _GridIter(self)
+        for i in range(len(self)):  # pylint:disable=consider-using-enumerate
+            yield self[i]
+
+    def __len__(self):
+        return self.cols * self.rows
 
     def _index_to_position(self, index):
         area = self.cols * self.rows
@@ -214,25 +218,6 @@ class Grid(object):
         else:
             raise IndexError("Index out of range: position %r in %r" %
                              (position, self))
-
-
-class _GridIter(object):
-    def __init__(self, grid):
-        self.grid = grid
-        self.index = 0
-
-    def __iter__(self):
-        return self
-
-    def __next__(self):
-        i = self.index
-        if i == self.grid.area:
-            raise StopIteration()
-        self.index += 1
-        return self.grid[i]
-
-    def next(self):  # Python 2
-        return self.__next__()
 
 
 def grid_to_navigation_graph(grid):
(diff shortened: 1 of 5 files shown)


_stbt/grid.py
```diff
@@ -140,63 +215,70 @@ def position_to_region(self, position):
...
+        return self
 
-        :param int index: See `Grid.index_to_position`.
-        :returns: See `Grid.position_to_region`.
-        """
-        return self.position_to_region(self.index_to_position(index))
+    def __next__(self):
+        i = self.index
+        if i == self.grid.area:
+            raise StopIteration
```
wmanley:
```suggestion
            raise StopIteration()
```

drothlis: The Python tutorial here uses `StopIteration` without the parentheses, in both the Python 2 & Python 3 versions of the doc:

https://docs.python.org/3/tutorial/classes.html#iterators

drothlis: I did a quick test and calling `next` on a generator function does raise an **instance** of `StopIteration`, so I'll commit your suggestion.

_stbt/grid.py
```diff
@@ -20,106 +22,179 @@ class Grid(object):
...
+        elif data is not None:
+            for i in range(self.cols * self.rows):
+                position = self._index_to_position(i)
+                if data == self.data[position[1]][position[0]]:
+                    index = i
+                    region = (None if self.region is None
+                              else self._position_to_region(position))
+                    break
+            else:
+                raise IndexError("data '%r' not found" % data)
```
wmanley: I'm not sure if you want quoting around `%r`, if it's a str it will end up double quoted.  Also: `data` could be a `tuple` in which case this will fail.

```suggestion
                raise IndexError("data %r not found" % (data,))
```

_stbt/grid.py
```diff
@@ -20,106 +22,173 @@ class Grid(object):
...
+        pass
 
     def __repr__(self):
-        return "Grid(region=%r, cols=%r, rows=%r)" % (
+        s = "Grid(region=%r, cols=%r, rows=%r)" % (
             self.region, self.cols, self.rows)
+        if self.data:
+            return "<" + s + ">"
+        else:
+            return s
```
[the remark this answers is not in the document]
drothlis: I did it this way because the data can get very verbose.

_stbt/grid.py
```diff
@@ -272,3 +230,52 @@ def __next__(self):
...
+    .. _Directed Graph: https://en.wikipedia.org/wiki/Directed_graph
+    .. _hashable: https://docs.python.org/3.6/glossary.html#term-hashable
+    """
+    G = nx.DiGraph()
+
+    def name(c):
+        if c.data is None:
+            return c.index
+        else:
+            return c.data
```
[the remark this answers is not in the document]
drothlis: I'm not too fussed and I want to get this merged.

Grids with `region=None` are only really for our self-tests; in practice when you create a Grid you need to specify a region so that you can map the selection (found with `stbt.match`) to the relevant grid data.

_stbt/grid.py
```diff
@@ -157,6 +157,9 @@ def __getitem__(self, key):
         else:
             return self.get(data=key)
 
+    def __iter__(self):
+        return _GridIter(self)
```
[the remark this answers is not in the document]
drothlis: Done.

PR #618, opened 2019-08-30
Fix matchTemplate error for match on the edge of the previous pyramid level

drothlis:
The error is:

> cv2.error: (-215) _img.size().height <= _templ.size().height &&
> _img.size().width <= _templ.size().width in function matchTemplate

Presumably we are passing a too-small roi into the next pyramid level.

TODO:

- [x] Regression test.
- [x] Fix it.
- [x] Construct an artificial regression test to prove my theory of the cause.
- [x] Update `test_match_region` to include:
    - odd-sized templates at the edge
    - matching 1px from the edge


drothlis:
This appears to be a regression since v30. The regression test passes when I rebase it onto v30.

Here are the changes to `_stbt/match.py` since v30:
https://github.com/stb-tester/stb-tester/compare/v30...master#diff-e0e121b2903d65dac3300623ce4b7c28

drothlis:
Reverting 0a4e6a1dc2494 (#591) fixes this regression test (`test_pyramid_roi_too_small`).

drothlis:
Right. In 0a4e6a1 we crop out the outside 1px of the template, and to compensate we expand the ROI by 1px, but if this ROI is on the edge of the frame, the expanded 1px is outside the frame so it's ignored.

Possible solutions:

1. If the smaller template matches right at the edge, but it wouldn't have matched 1px away from the edge, then the full-sized template isn't going to match anyway. So we can discard ROIs that are smaller than the template.

    match-at-the-edge

2. Crop the edges of the frame by 1px too (only at the smaller pyramid levels), and then adjust the positions of the match before doing the next-higher level.

drothlis:
> Construct an artificial regression test to prove my theory of the cause.

Hmmm I haven't been able to construct an artificial test that fails without the fix. See b34f8bc Add more tests for stbt.match at the edges of the image.

drothlis:
@wmanley I'm not sure what to do next. Merge this fix, even though I haven't been able to reproduce it with an artificial test-case? (I've only reproduced it with the frame & reference images seen in the wild.) Or revert #591?

drothlis:
Yay, I have reproduced this bug with an artificial test-case against master (c1611f2cfe28):

test_match_region2[frame_width=320,frame_height=180,width=20,height=20,offset_x=-2,offset_y=0,x=-20,y=0]

drothlis:
@wmanley this is ready now. Please review.

PR #619, opened 2019-09-02
stbt.Keyboard.Selection: Implement `__bool__`

drothlis:
So that you can return `Selection(None, None)` if no selection is
detected, and callers can say `if selection:` instead of
`if selection.region:`.


drothlis:
Or maybe the Page Object should return `None` instead of `Selection(None, None)`.

wmanley:
> Or maybe the Page Object should return None instead of Selection(None, None).

I'm erring on the side of this.  Of course these two approaches aren't mutually exclusive, and we don't have the walrus operator yet.

Implementation wise it looks fine to me.

PR #621, opened 2019-09-05
Disallow calling region.intersect() on a Region instance

drothlis:
`intersect` is a `@staticmethod`, but you can still call it on an instance -- and the instance will be ignored, not considered among the arguments.

So `r1.intersect(r2)` behaves like `stbt.Region.intersect(r2)`, not like `stbt.Region.intersect(r1, r2)`.

Before v30, calling `intersect` with a single region wouldn't have been allowed because it required exactly 2 arguments; but since v30 (when we allowed an arbitrary number of arguments) this incorrect usage wouldn't cause an assertion and even `stbt lint` wouldn't catch it.

Similarly for stbt.Region.bounding_box.

---

TODO:

- [x] Check if `intersect` and `boundingbox` still show up in the Sphinx documentation.

### DIFF
--- a/_stbt/types.py
+++ b/_stbt/types.py
@@ -6,6 +6,8 @@ from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
 from builtins import *  # pylint:disable=redefined-builtin,unused-wildcard-import,wildcard-import,wrong-import-order
+from future.utils import with_metaclass
+
 from collections import namedtuple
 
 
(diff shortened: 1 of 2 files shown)


wmanley:
This is a neat solution, and less scary than it initially sounds ("metaclass").

drothlis:
You can thank stackoverflow. That reminds me, must upvote that answer.

drothlis:
I can't get these class methods from the MetaClass to show up in the documentation generated with Sphinx. I've tried this:

```
stbt.Region
-----------
.. autoclass:: stbt.Region
   :members:
   :inherited-members: bounding_box, intersect
   :member-order: bysource
   :exclude-members: from_extents
```

Note that giving specific `inherited-members` doesn't have any effect -- all the inherited members show up:

> bottom
>     Alias for field number 3
> 
> count(value) → integer -- return number of occurrences of value
> 
> index(value[, start[, stop]]) → integer -- return first index of value.
>     Raises ValueError if the value is not present.
> 
> right
>     Alias for field number 2
> 
> x
>     Alias for field number 0
> 
> y
>     Alias for field number 1

PR #622, opened 2019-09-05
Region: Allow translating one region by another one

wmanley:
...in fact this will allow translating by any tuple that has x and y as the first two items.  This is helpful when dealing with items that have substructure.  For example: A carousel might have multiple items laid out in a grid:

    +---------------+     +----------------+
    |    Title 1    |     |    Title 2     |
    |               |     |                |
    | Description 1 |     | Description 2  |
    | this TV show  |     | this one isn't |
    | is really     |     | so interesting |
    | good and      |     |                |
    | exciting      |     |                |
    |               |     |                |
    +---------------+     +----------------+

The boxes might be given by `REGION[0]` and `REGION[1]`.  You could then
have `TITLE_REGION` and `DESCRIPTION_REGION` relative to the boxes.  To read
all the information it might look like:

    for r in REGION:
        title = stbt.ocr(region=TITLE_REGION.translate(r))
        description = stbt.ocr(region=DESCRIPTION_REGION.translate(r))

### DIFF
--- a/_stbt/types.py
+++ b/_stbt/types.py
@@ -115,9 +115,6 @@ class Region(with_metaclass(_RegionClsMethods,
     Region(x=1, y=1, right=4, bottom=3)
     >>> Region(2, 3, 2, 1).translate(b)
     Region(x=6, y=7, right=8, bottom=8)
-    >>> Region(2, 3, 2, 1).translate(b, 5)
-    Traceback (most recent call last):
-    TypeError
     >>> Region.intersect(Region.ALL, c) == c
     True
     >>> Region.ALL
--- a/_stbt/types.py
+++ b/_stbt/types.py
@@ -269,13 +269,11 @@ class Region(with_metaclass(_RegionClsMethods,
         :returns: A new region with the position of the region adjusted by the
             given amounts.  The width and height are unaffected.
 
-        `translate` either accepts seperate `int` x and y coordinates or a
-        single `Region` argument:
+        ``translate`` accepts separate x and y arguments, or a single `Region`.
+
+        For example, move the region 1px right and 2px down:
 
         >>> b = Region(4, 4, 9, 6)
-
-        Move the region 1px right and 2px down:
-
         >>> b.translate(1, 2)
         Region(x=5, y=6, right=14, bottom=12)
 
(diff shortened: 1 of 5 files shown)


wmanley:
Hmm, this failed in `test_that_stbt_run_exits_on_ctrl_c`.  I have a hard time believing that the failure is caused by this patch, but it's failed twice in a row, once on [Python 2](https://circleci.com/gh/stb-tester/stb-tester/1040) and once on Python 3 (https://circleci.com/gh/stb-tester/stb-tester/1042).  I'll rerun and investigate this separately.

_stbt/types.py
```diff
@@ -71,6 +71,11 @@ class Region(namedtuple('Region', 'x y right bottom')):
     Region(x=-inf, y=-inf, right=2, bottom=2)
     >>> c.translate(x=-9, y=-3)
     Region(x=1, y=1, right=4, bottom=3)
+    >>> Region(2, 3, 2, 1).translate(b)
+    Region(x=6, y=7, right=8, bottom=8)
+    >>> Region(2, 3, 2, 1).translate(b, 5)
+    Traceback (most recent call last):
+    TypeError
```
drothlis: I don't think this last test should go in the docstring.

_stbt/types.py
```diff
@@ -242,13 +247,20 @@ def norm_coords(name_x, name_width, name_right,
 
         return Region(x=x, y=y, right=right, bottom=bottom)
 
-    def translate(self, x=0, y=0):
+    def translate(self, x=None, y=None):
```
drothlis: ~I think we could leave the defaults as 0 & 0, for documentation purposes.~ I suppose it's so that you can check if you've passed `y` when you've already given a region.

The docstring needs to explain that you can give it x & y ints or a single `Region`.

_stbt/types.py
```diff
@@ -242,13 +247,20 @@ def norm_coords(name_x, name_width, name_right,
...
         """
-        return Region.from_extents(self.x + x, self.y + y,
-                                   self.right + x, self.bottom + y)
+        try:
+            p = x[0], x[1]
+        except TypeError:
+            p = x or 0, y or 0
+        else:
+            if y is not None:
+                raise TypeError()
```
drothlis: The TypeError needs a message explaining what you've done wrong.

wmanley:
@drothlis: What do you think of the docs now?

PR #623, opened 2019-09-16
stbt.Grid: Fix returned region when getting by region 

drothlis:


_stbt/grid.py
```diff
@@ -129,6 +129,8 @@ def get(self, index=None, position=None, region=None, data=None):
         elif region is not None:
```
wmanley: Not related to this PR but this makes me think.  Currently we consider `None` to be a kind of region - intersect will return None if there is no intersection.  This is probably a mistake.  We should define something like `Region.NONE` which has the usual region methods and `isinstance(Region.NONE, Region)`, but isn't `None` to make this unambiguous.

wmanley: See #624

PR #624, opened 2019-09-17
Replace treating `None` as `Region` with `Region.NONE`

wmanley:
This may make it easier to use functions that return a `Region` that might
be `None` because in many cases they may not need to check `region is None`
before performing operations on that region.  It also makes it easier to
implement functions with take a default `Region` because we can follow the
standard `region is not None` to mean "did the caller pass a region
argument".  For an example of this see `Grid`.

I don't think this should go in the next release.  It would be better to let it bake for a while.

**TODO:**

- [ ] Should `Region.NONE == Region.NONE`?  Currently `Region.NONE == Region.NONE`, but `Region.NONE != Region(float('nan'), float('nan'), float('nan'), float('nan'))` because `NAN != NAN` and `NAN != float('nan')` but `(NAN,) == (NAN,)` and `(NAN,) != (float('nan'))`
- [ ] Should `Region.NONE.width == 0` or `nan`?, equally for height
- [ ] Is `NONE` a good name?  Perhaps `EMPTY` would be better?
- [ ] Codify all these weird behaviours in tests so we'll know if we break them
- [ ] Validate that we've converted all the uses of `None` as `Region` in our codebase.
- [ ] Release note

### DIFF
--- a/_stbt/logging.py
+++ b/_stbt/logging.py
@@ -180,7 +180,7 @@ class ImageLogger(object):
     def _draw(self, region, source_size, css_class, title=None):
         import jinja2
 
-        if region is None:
+        if not region:
             return ""
 
         if isinstance(css_class, bool):
--- a/_stbt/match.py
+++ b/_stbt/match.py
@@ -374,7 +374,7 @@ def _match_all(image, frame, match_parameters, region):
                 % (template.relative_filename, match_parameters.match_method))
 
     input_region = Region.intersect(_image_region(frame), region)
-    if input_region is None:
+    if not input_region:
         raise ValueError("frame with dimensions %r doesn't contain %r"
                          % (frame.shape, region))
     if input_region.height < t.shape[0] or input_region.width < t.shape[1]:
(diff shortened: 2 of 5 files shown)


drothlis:
> Should `Region.NONE == Region.NONE`?  Currently `Region.NONE == Region.NONE`, but `Region.NONE != Region(float('nan'), float('nan'), float('nan'), float('nan'))` because `NAN != NAN` and `NAN != float('nan')` but `(NAN,) == (NAN,)` and `(NAN,) != (float('nan'))`

Yes, `Region.NONE == Region.NONE`.

> Should `Region.NONE.width == 0` or `nan`?, equally for height

0, I think.

> Is `NONE` a good name?  Perhaps `EMPTY` would be better?

EMPTY, definitely better.

drothlis:
> but `(NAN,) == (NAN,)`

Interesting. That's still true as of Python 3.10.5.

wmanley:
Rebased on master to resolve conflicts.

PR #625, opened 2019-09-17
stbt.Keyboard: Verify that you've reached the target

drothlis:
Previously the only verification was in the while loop's condition:
`while current != target`. So if your model of the keyboard was wrong,
`navigate_to` would keep on trying to reach the target until
`navigate_timeout` was reached (default 20 seconds). Maybe this would
allow to eventually reach the key, but it can also lead to very
confusing, hard to debug, behaviour.

Now we verify that we've reached each letter in the word you're trying
to type, so we fail earlier and we provide a useful error message.

Optionally we can also verify each state in the journey from the current
selection to the target letter, by specifying
`verify_every_keypress=True`. This is slower and less human-like but
you'd only set while debugging your model.

---

TODO:

- [x] Tests with ~an incorrect model~ a model that doesn't match the actual behaviour.

### DIFF
--- a/stbt/keyboard.py
+++ b/stbt/keyboard.py
@@ -237,22 +237,30 @@ class Keyboard(object):
         if target not in self.G:
             raise ValueError("'%s' isn't in the keyboard" % (target,))
 
+        assert page, "%s page isn't visible" % type(page).__name__
         deadline = time.time() + self.navigate_timeout
         current = _selection_to_text(page.selection)
         while current != target:
-            assert page, "%s page isn't visible" % type(page).__name__
             assert time.time() < deadline, (
                 "Keyboard.navigate_to: Didn't reach %r after %s seconds"
                 % (target, self.navigate_timeout))
             keys = list(_keys_to_press(self.G, current, target))
             log.info("Keyboard: navigating from %s to %s by pressing %r",
                      current, target, keys)
-            for k in keys[:-1]:
+            for k, _ in keys[:-1]:
                 stbt.press(k)
-            assert stbt.press_and_wait(keys[-1], mask=self.mask,
-                                       stable_secs=0.5)
+            key, possible_targets = keys[-1]
+            assert stbt.press_and_wait(key, mask=self.mask, stable_secs=0.5)
             page = page.refresh()
+            assert page, "%s page isn't visible" % type(page).__name__
             current = _selection_to_text(page.selection)
+            assert current in possible_targets, \
+                "Expected to see %s after pressing %s, but saw %r" % (
+                    _join_with_commas(
+                        [repr(x) for x in sorted(possible_targets)],
+                        last_one=" or "),
+                    key,
+                    current)
         return page
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


drothlis:
@wmanley please review. The 1st commit might be sufficient to fix the confusing behaviour you experienced. The 2nd commit adds even more (optional) checking.

stbt/keyboard.py
```diff
@@ -327,3 +337,24 @@ def _add_weights(G):
...
+    ''
+    """
+    if len(items) > 1:
+        return last_one.join([
+            ", ".join(items[:-1]),
+            items[-1]])
+    elif len(items) == 1:
+        return items[0]
+    else:
+        return ""
```
wmanley: A less readable oneliner:

    return " or ".join([", ".join(items[:-1]), items[-1]]) if items else ""

stbt/keyboard.py
```diff
@@ -327,3 +337,24 @@ def _add_weights(G):
                 # take a shortcut through here.
                 for target in targets:
                     G[node][target]["weight"] = 100
+
+
+def _join_with_commas(items, last_one=", "):
```
wmanley: Presumably there will be a `last_one=" and "` in a future commit...

No, I didn't see it.  I don't see why this couldn't default to `" or "` or even have `" or "` hard-coded.

stbt/keyboard.py
```diff
@@ -171,6 +171,19 @@ def enter_text(self, text, page):
...
 
+        :param bool verify_every_keypress:
+            If True, we will read the selected key after every keypress and
+            assert that it matches the model (``graph``). If False (the
+            default) we will only verify the selected key corresponding to each
+            of the characters in ``text``. For example: to get from Q to D on a
+            qwerty keyboard you need to press KEY_DOWN, KEY_RIGHT, KEY_RIGHT.
+            The default behaviour will only verify that the selected key is D
+            after pressing KEY_RIGHT the last time. This is faster, and closer
+            to the way a human uses the on-screen keyboard.
```
wmanley: It's probably closer to how experienced humans use the on-screen keyboard - like manual testing.  It's how I use it, but other members of my family wait after each press before making the next press.

One way this implementation is not like a human, is that if it were to land on the wrong key the human might be surprised, but they'd just carry on from where they ended up and not worry too much.

stbt/keyboard.py
drothlis:
```
>>> items = ['a']
>>> " or ".join([", ".join(items[:-1]), items[-1]]) if items else ""
' or a'
```

stbt/keyboard.py
drothlis: It seemed clearer from the call-site.

stbt/keyboard.py
drothlis: Good point about non-technical users.

I suppose with the original implementation I was kinda hoping it would work like you describe (carry on from where you ended up) but in reality it was just confusing.

stbt/keyboard.py
drothlis: And yes, I made exactly that mistake in my initial implementation. :-)

PR #628, opened 2019-09-30
ocr, match_text: Raise exception if region is outside of frame

drothlis:
For `ocr` and `match_text` this is a change in behaviour: Previously if the region was entirely outside of the frame, they would log a warning and return "" or a Falsey TextMatchResult, respectively (but for `region=None` they would already raise a TypeError). I don't think the old behaviour is anything you'd ever want to do intentionally, it's far more likely to be a bug in your test. And the new behaviour is consistent with all the other stbt APIs.

For all the other functions this is a nicer error message. `press_and_wait` would raise "UnboundLocalError: local variable 'f1' referenced before assignment". The others would raise either "ValueError: frame with dimensions (720, 1280, 3) doesn't contain None" (from stbt.crop) or "AttributeError: 'NoneType' object has no attribute 'x/y/right/bottom'".

### DIFF
--- a/_stbt/black.py
+++ b/_stbt/black.py
@@ -16,8 +16,8 @@ from builtins import *  # pylint:disable=redefined-builtin,unused-wildcard-impor
 import cv2
 
 from .config import get_config
-from .imgutils import (_frame_repr, _image_region, load_image,
-                       pixel_bounding_box, crop)
+from .imgutils import (crop, _frame_repr, load_image, pixel_bounding_box,
+                       _validate_region)
 from .logging import debug, ImageLogger
 from .types import Region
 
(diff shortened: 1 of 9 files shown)


drothlis:
My motivation for this was to fix a very confusing error message in `stbt.press_and_wait` when passing `region=None`.

drothlis:
@wmanley please review.

According to the comments in the selftests (which I have removed), we used to want to allow passing `region=None` into `ocr` and have it return `""`. I can't say I've ever wanted that functionality -- if you pass a calculated region that turns out to be None or outside of the frame, it's far more likely to be a bug in your test script in which case you want a loud exception instead of an easy to miss warning.

drothlis:
Rebased onto master & fixed conflicts.

drothlis:
@wmanley please review.

drothlis:
Rebased onto master (again) to fix conflicts. I re-did the 3rd commit from scratch to make sure I didn't miss anything.

PR #629, opened 2019-10-14
Dont expose future newtypes

wmanley:
This fixes using `stbt.Keyboard` on Python 2.

Don't expose future.types.newxxx to our users
--------------------------------------------------

I've checked all our code for instances of `bytes(` or `str(`.  On Python 2
with the `future` library they can actually return instances of `newbytes`
and `newstr` rather than what you'd expect.  This changes some calls to
`str` to be explicit about what type you'll get, and to avoid these
future newtypes.  future newtypes can behave oddly so I want to avoid
exposing them to our users.

I'm of the opinion that as a library that intends to support Python 2 and 3
`stbt` shouldn't leak that we've implemented this support with
python-future so we should return normal Python types wherever possible.
We need to support both Python 2 and 3, but it doesn't make sense for our
customers' test-packs to be python-version agnostic.

We're still exposing some details of python-future to our users - most notably as we're deriving from `future.object`.  I've not come across any issues caused by this yet.

### DIFF
--- a/_stbt/black.py
+++ b/_stbt/black.py
@@ -11,7 +11,8 @@ from __future__ import unicode_literals
 from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
-from builtins import *  # pylint:disable=redefined-builtin,unused-wildcard-import,wildcard-import,wrong-import-order
+from builtins import (ascii, chr, filter, hex, input, map, next, oct, open, pow,  # pylint:disable=redefined-builtin,unused-import,wildcard-import,wrong-import-order
+                      range, round, super, zip)
 
 import cv2
 
--- a/_stbt/config.py
+++ b/_stbt/config.py
@@ -2,7 +2,8 @@ from __future__ import unicode_literals
 from __future__ import print_function
 from __future__ import division
 from __future__ import absolute_import
-from builtins import *  # pylint:disable=redefined-builtin,unused-wildcard-import,wildcard-import,wrong-import-order
+from builtins import (ascii, chr, filter, hex, input, map, next, oct, open, pow,  # pylint:disable=redefined-builtin,unused-import,wildcard-import,wrong-import-order
+                      range, round, super, zip)
 from future.utils import native_str
 
 import configparser
--- a/_stbt/control.py
+++ b/_stbt/control.py
@@ -2,7 +2,8 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 from __future__ import print_function
 from __future__ import division
-from builtins import *  # pylint:disable=redefined-builtin,unused-wildcard-import,wildcard-import,wrong-import-order
+from builtins import (ascii, chr, filter, hex, input, map, next, oct, open, pow,  # pylint:disable=redefined-builtin,unused-import,wildcard-import,wrong-import-order
+                      range, round, super, zip)
 from future.utils import text_to_native_str
 
 import os
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -405,9 +405,9 @@ def _fake_cec():
     def cec_cmd_get_data(cmd):
         # Ugly, but can't find another way to do it
         import ctypes
-        return bytes(memoryview(ctypes.cast(  # pylint:disable=undefined-variable
-            int(cmd.parameters.data), ctypes.POINTER(ctypes.c_uint8)).contents,
-            0, cmd.parameters.size))
+        return bytearray(ctypes.cast(
+            int(cmd.parameters.data),
+            ctypes.POINTER(ctypes.c_uint8 * cmd.parameters.size)).contents)
 
     def Transmit(_, cmd):
         io.write(b"Transmit(dest: 0x%x, src: 0x%x, op: 0x%x, data: <%s>)\n" % (
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -133,14 +133,7 @@ class HdmiCecControl(object):
     }
 
     def __init__(self, device, source, destination):
-        # On Ubuntu 18.04 `cec/__init__.py` tries to import `_cec` but can't
-        # find it.
-        # https://bugs.launchpad.net/ubuntu/+source/libcec/+bug/1805620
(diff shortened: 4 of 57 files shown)


wmanley:
> Sometimes they would return python-future's newtypes which behave oddly in
certain circumstances.  For example:
>
>    b"Hi " + newbytes("there")
>
> would be `"Hi b'there'"`.  This was causing `press()` to fail as we'd ask
> `lircd` for key `b'KEY_RIGHT'`.

I'm not sure this is actually true.  Let me reproduce.

allanlewis:
[`six`][1] has [`text_type`][2] and [`binary_type`][3] for this purpose, both of which are aliases to the correct types in Python 2 or Python 3, as appropriate. Apologies if `six` in general or this specifically has already been considered - I haven't been following the whole of the Python 3 compatibility project.

[1]: https://six.readthedocs.io/
[2]: https://six.readthedocs.io/#six.text_type
[3]: https://six.readthedocs.io/#six.binary_type

wmanley:
I've checked all the [18 overrides](https://python-future.org/imports.html#explicit-imports) that `future.builtins` contains:

ascii, bytes, chr, dict, filter, hex, input, int, map, next, oct, open, pow, range, round, str, super, zip

We only use `bytes`, `str` and `super` in `utils.py`, and all these uses are in the implementation of `strip_newtypes`, `to_bytes` and `to_unicode`.

wmanley:
I've only audited the `bytes`, `str` and `dict` newtypes.  Most notably I haven't audited `int` which could be an issue:

```python
>>> i = future.builtins.int(5)
>>> isinstance(i, int)
False
>>> isinstance(i, long)
True
```

_stbt/utils.py
```diff
@@ -88,8 +89,51 @@ def find_import_name(filename):
     return import_dir, import_name
 
 
+if sys.version_info.major == 2:  # Python 2
+    text_type = unicode  # pylint: disable=undefined-variable
```
drothlis: python-future already has this as `future.utils.text_type` (it's the native `unicode` type on Python 2, not `newstr`).

_stbt/utils.py
```diff
@@ -88,8 +89,51 @@ def find_import_name(filename):
...
+
+    def strip_newtypes(text):
```
drothlis: This could all probably be replaced by [future.utils.native](https://python-future.org/reference.html#future.utils.native):

> On Py3, this is a no-op: native(obj) -> obj
>
> On Py2, returns the corresponding native Py2 types that are superclasses for backported objects from Py3:
> [...]
> Existing native types on Py2 will be returned unchanged.

_stbt/utils.py
```diff
@@ -88,8 +89,51 @@ def find_import_name(filename):
...
+        check(strip_newtypes(newbytes(b"abc")), b"abc")
+else:
+    text_type = str
+
+    def strip_newtypes(text):
+        # newtypes won't be used on Python 3
+        return text
+
+
+native_str = str
```
drothlis: Instead of this other modules can import [future.utils.native_str](https://python-future.org/reference.html#module-future.utils):

> always equal to the native platform string object (because this may be shadowed by imports from future.builtins)

wmanley:
I've added `text_type` and `native_str` to `_stbt.utils` here rather than
`from future.utils import text_type, native_str` because the implementation
is *trivial* and I'd like to drop the dependency on python-future when
we're using Python 3 (at some point in the future).  I figure that
`python-future` is supposed to let you "just" write in Python 3, and
convert that to Python 2 as necessary, so it doesn't make a lot of sense
to require it when running on Python 3 natively.

This is why I've named these functions after the equivalent in future/six.  I'm not married to this decision, so I'll change it if you disagree.

wmanley:
Ok, I've checked `int` as well.

drothlis:
> I figure that python-future is supposed to let you "just" write in Python 3, and convert that to Python 2 as necessary, so it doesn't make a lot of sense to require it when running on Python 3 natively.

I think the policy should be: Use the "basic" `future` compatibility features (like `native_str`, `string_types` etc) without using the "magic" features (replacing builtins with future types that almost-but-not-quite behave like Python 3 builtins). This is identical to using `six` except that `future` comes with a `futurize` script that can do some of the conversions automatically -- though perhaps that's less important now that the conversion has been done.

I **don't** think we should write our own Python 2/3 compatibility layer for the sake of avoiding an import on Python 3.

wmanley:
I agree, that sounds like good policy.  I'll adjust this code accordingly.

drothlis:
> I don't think we should write our own Python 2/3 compatibility layer for the sake of avoiding an import on Python 3.

I've changed my mind about this.

```
>>> # Python 2
>>> from future.utils import text_to_native_str
>>> text_to_native_str(b"Röthlisberger", encoding="utf-8")
/lib/python2.7/site-packages/future/utils/__init__.pyc in text_to_native_str(t, encoding)
--> 221         return unicode(t).encode(encoding)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 5: ordinal not in range(128)
```

So bytes -> bytes conversion fails. WTF. I suppose "text" means unicode but there's no other function --as far as I can tell-- to convert bytes or unicode to the native str type.

drothlis:
`six` has `ensure_str` which is sensible:

```
def ensure_str(s, encoding='utf-8', errors='strict'):
    """Coerce *s* to `str`.

    For Python 2:
      - `unicode` -> encoded to `str`
      - `str` -> `str`

    For Python 3:
      - `str` -> `str`
      - `bytes` -> decoded to `str`
    """
    if not isinstance(s, (text_type, binary_type)):
        raise TypeError("not expecting type '%s'" % type(s))
    if PY2 and isinstance(s, text_type):
        s = s.encode(encoding, errors)
    elif PY3 and isinstance(s, binary_type):
        s = s.decode(encoding, errors)
    return s
```

but six also has hundreds of lines messing with import paths, "MovedAttribute" and "MovedModule" and I'm feeling a bit burned from that kind of magic.
https://github.com/benjaminp/six/blob/master/six.py

drothlis:
Superseded by #728.

PR #630, opened 2019-10-14
Backport parts of #629 to v31

wmanley:
So `stbt.Keyboard` isn't broken for our customers who use Python 2.


drothlis:
I've merged these changes to master too, seeing #629 isn't merged to master yet.

PR #631, opened 2019-10-15
Add `for_object_repository` decorator

wmanley:
A decorator that marks classes and functions so they appear in the Object
Repository.

Classes that directly derive from `stbt.FrameObject` have this decorator
applied to them automatically.  This can be used to register other classes
and functions.

These classes are extracted by static analysis, so the decorator is a noop,
returning the class unchanged.

### DIFF
--- a/_stbt/frameobject.py
+++ b/_stbt/frameobject.py
@@ -20,6 +20,38 @@ except ImportError:
     from itertools import izip_longest as zip_longest
 
 
+def for_object_repository(cls=None):
+    """A decorator that marks classes and functions so they appear in the Object
+    Repository.
+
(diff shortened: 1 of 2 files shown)


_stbt/frameobject.py
```diff
@@ -93,6 +122,7 @@ def __init__(cls, name, parents, dct):
         super(_FrameObjectMeta, cls).__init__(name, parents, dct)
 
 
+@for_object_repository()
```
drothlis: A bit late but is this decorator appropriate here? You only need it on indirect **subclasses** of FrameObject.

drothlis: Raised #635.

Issue #632, opened 2019-10-19
Makefile generates clang error with xxhash.h file included


eugenetriguba:
**Environment**: Ubuntu 18.04 LTS

**Issue**: I get a clang error whenever I try to build stb-tester from source using the Makefile.

```sh
$ make
virtual-stb support disabled
cc -shared -fPIC -O3 -o _stbt/libstbt.so _stbt/sqdiff.c -O2
git submodule init && \
git submodule sync && \
git submodule update && \
touch vendor/.submodules-checked-out
Synchronizing submodule url for 'vendor/xxHash'
cc -shared -fPIC -O3 -o _stbt/libxxhash.so vendor/xxHash/xxhash.c vendor/xxHash/xxhash.h -O2
clang: error: cannot specify -o when generating multiple output files
Makefile:304: recipe for target '_stbt/libxxhash.so' failed
make: *** [_stbt/libxxhash.so] Error 1
```

**Solution**: Under the "Third-party dependencies" section in the Makefile, removing ```vendor/xxHash/xxhash.h``` from the ```XXHASH_SOURCES``` results in successful compilation.

https://stackoverflow.com/questions/43525486/cannot-specify-o-when-generating-multiple-output-files-c-error

Does the header file need to be there or is it worth considering removing since I wasn't able to get things running with it there?

drothlis:
Thanks for the bug report. I can reproduce this with clang but not with gcc. I suppose clang doesn't like being given .h files on the command line and thinks they're .c files.

We do still need the header file in `XXHASH_SOURCES` so that libxxhash.so gets rebuilt if the header file changes. Please remove the header file from the compilation command line using make's [filter] function, and submit a pull request.

[filter]: https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

PR #633, opened 2019-10-25
Filter out xxhash header file to prevent clang error

eugenetriguba:
If you attempt to run make using clang, you would currently get an error where xxHash is being compiled because it appears clang will treat all the files as .c files. Since the .h file is still necessary in the sources, this change circumvents that error by simply filtering it out from the compilation step. 

```sh
clang: error: cannot specify -o when generating multiple output files
make: *** [_stbt/libxxhash.so] Error 1
```

See issue #632


drothlis:
Merged. Thanks! Fixes #632.

PR #634, opened 2019-11-03
stbt.Keyboard: Automatically refresh page if it's out of date

drothlis:
This bypasses a common error in user code when we forget to update
`page` between calls to `enter_text` and `navigate_to`. For example:

    def enter_text(self, text):
        kb.enter_text("peppa pig", page=self)
        kb.navigate_to("SEARCH", page=self)  # self.selection is outdated!
        stbt.press_and_wait("OK")

We consider the Page Object instance to be out of date if its frame is
older than the last keypress. This won't reflect changes that aren't
triggered by the remote control (particularly changes triggered by the
passage of time, such as the screensaver activating) but in practice
it'll solve a common bug.

Of course it doesn't fix any other (non-Keyboard) bug caused by
out-of-date Page Object instances. Maybe this will only add to the
confusion ("If this code pattern causes a bug, how come the same pattern
does work when it's using stbt.Keyboard?"). Or maybe this is a good
pattern that we can start using elsewhere. Thoughts please.

### DIFF
--- a/stbt/keyboard.py
+++ b/stbt/keyboard.py
@@ -235,6 +235,9 @@ class Keyboard(object):
             if letter not in self.G:
                 raise ValueError("'%s' isn't in the keyboard" % (letter,))
 
+        if page._frame.time < stbt._dut._last_keypress.end_time:  # pylint:disable=protected-access
+            page = page.refresh()
+
         for letter in text:
             page = self.navigate_to(letter, page, verify_every_keypress)
             stbt.press("KEY_OK")
@@ -258,6 +261,9 @@ class Keyboard(object):
 
         if target not in self.G:
             raise ValueError("'%s' isn't in the keyboard" % (target,))
+
+        if page._frame.time < stbt._dut._last_keypress.end_time:  # pylint:disable=protected-access
+            page = page.refresh()
 
         assert page, "%s page isn't visible" % type(page).__name__
         deadline = time.time() + self.navigate_timeout


drothlis:
@wmanley wants a `stbt.page` global attribute that is always set to the latest truthy Page Object that has been constructed. This would provide a more general solution. The `page` argument to `stbt.Keyboard.enter_text` and `.navigate_to` would become optional, defaulting to `stbt.page`.

Requirements:

- Separate base class for full-screen Page Objects, vs. classes that represent smaller components *on* a page.
  - Full screen Page Objects:
    - constructed by user
    - has is_visible
    - Naming: PageObject
  - Parts of the page:
    - constructed by other Page Objects
    - doesn't have is_visible (or it always returns True)
    - Naming: PageComponent? Widget? WidgetObject?
      - I had previously seen a framework based on Selenium that has separate base classes for pages vs. components, but I can't find it now.

Consequences:

- is_visible becomes non-lazy because we need to evaluate it to decide whether to update `stbt.page`. In practice it might not matter -- why would you create a PageObject without evaluating is_visible pretty promptly?

drothlis:
We also discussed the fact that `FrameObject.refresh` doesn't change `self`: Users have found this confusing. We decided not to change this because doing so isn't general enough: You could use `refresh` to update a variable but only if the visible page is of the same type.

`stbt.page` described in the previous comment is a more general solution.

PR #635, opened 2019-11-04
stbt.FrameObject base class: Remove @for_object_repository decorator

drothlis:
This decorator is used for marking indirect subclasses of FrameObject,
that wouldn't be otherwise detected by Stb-tester's [Object Repository].
It doesn't make any sense to apply it to the FrameObject base class
itself.

[Object Repository]: https://stb-tester.com/manual/object-repository


wmanley:
I put this in originally really for documentation purposes, but I agree that it's confusing.  It could cause people to think that inheritance is the mechanism by which objects are selected for the object repository - which isn't true.

PR #637, opened 2019-11-11
stbt.ocr: Add `corrections` parameter

drothlis:
A dict of {bad: good} mappings to correct known OCR mistakes.

TODO:

- [x] Support plain text & regexes.
- [ ] Way of specifying default corrections.

Out of scope for this pull request:

- Also `stbt.match_text`? But it could be tricky if the pattern to replace spans multiple hocr xml elements.

### DIFF
--- a/_stbt/ocr.py
+++ b/_stbt/ocr.py
@@ -23,7 +23,8 @@ from .imgutils import _frame_repr, _image_region, crop
 from .logging import debug, ImageLogger, warn
 from .types import Region
 from .utils import (
-    named_temporary_directory, native_int, native_str, text_type, to_unicode)
+    basestring, named_temporary_directory, native_int, native_str, text_type,
+    to_unicode)
 
 # Tesseract sometimes has a hard job distinguishing certain glyphs such as
 # ligatures and different forms of the same punctuation.  We strip out this
@@ -150,7 +151,8 @@ def ocr(frame=None, region=Region.ALL,
         mode=OcrMode.PAGE_SEGMENTATION_WITHOUT_OSD,
         lang=None, tesseract_config=None, tesseract_user_words=None,
         tesseract_user_patterns=None, upsample=True, text_color=None,
-        text_color_threshold=None, engine=None, char_whitelist=None):
+        text_color_threshold=None, engine=None, char_whitelist=None,
+        corrections=None):
     r"""Return the text present in the video frame as a Unicode string.
 
     Perform OCR (Optical Character Recognition) using the "Tesseract"
@@ -240,10 +242,26 @@ def ocr(frame=None, region=Region.ALL,
         that tesseract won't think that a zero is the letter o.
         Note that Tesseract 4.0's LSTM engine ignores ``char_whitelist``.
 
+    :param dict corrections:
+        Dictionary of corrections to replace known OCR mis-reads. Each key of
+        the dict is the text to search for; the value is the corrected string
+        to replace the matching key. If the key is a string, it is treated as
+        plain text and it will only match at word boundaries (for example the
+        string ``"he saw"`` won't match ``"the saw"`` nor ``"he saws"``). If
+        the key is a regular expression pattern (created with `re.compile`) it
+        can match anywhere, and the replacement string can contain
+        backreferences such as ``"\1"`` which are replaced with the
+        corresponding group in the pattern (same as Python's `re.sub`).
+        Example::
+
+            corrections={'bad': 'good',
+                         re.compile(r'[oO]'): '0'}
+
     | Added in v28: The ``upsample`` and ``text_color`` parameters.
     | Added in v29: The ``text_color_threshold`` parameter.
     | Added in v30: The ``engine`` parameter and support for Tesseract v4.
     | Added in v31: The ``char_whitelist`` parameter.
+    | Added in v32: The ``corrections`` parameter.
     """
     if frame is None:
         import stbt
@@ -269,6 +287,10 @@ def ocr(frame=None, region=Region.ALL,
         tesseract_user_patterns, tesseract_user_words, upsample, text_color,
         text_color_threshold, engine, char_whitelist, imglog)
     text = text.strip().translate(_ocr_transtab)
+
+    if corrections is not None:
+        text = apply_ocr_corrections(text, corrections)
+
     debug(u"OCR in region %s read '%s'." % (region, text))
     _log_ocr_image_debug(imglog, text)
     return text
@@ -361,6 +383,31 @@ def match_text(text, frame=None, region=Region.ALL,
     _log_ocr_image_debug(imglog)
 
     return result
+
+
+# Python 2.7 & 3.6 have `re._pattern_type` but that will be removed in Python
+# 3.7 where they introduce `re.Pattern`.
+PatternType = type(re.compile(""))
+
+
+def apply_ocr_corrections(text, corrections):
+    """Applies the same corrections as `stbt.ocr`'s ``corrections`` parameter.
+
+    This is also available as a separate function, so that you can use it to
+    post-process old test artifacts using new corrections.
+    """
+    # Match plain strings at word boundaries:
+    pattern = "|".join(r"\b(" + re.escape(k) + r")\b"
+                       for k in corrections
+                       if isinstance(k, basestring))
+    if pattern:
+        replace = lambda matchobj: corrections[matchobj.group(0)]
+        text = re.sub(pattern, replace, text)
+    # Match regexes:
+    for k, v in corrections.items():
+        if isinstance(k, PatternType):
+            text = re.sub(k, v, text)
+    return text
 
 
 _memoise_tesseract_version = None
--- a/_stbt/ocr.py
+++ b/_stbt/ocr.py
@@ -257,6 +257,14 @@ def ocr(frame=None, region=Region.ALL,
             corrections={'bad': 'good',
                          re.compile(r'[oO]'): '0'}
 
+        Plain strings are replaced first (in the order they are specified),
+        followed by regular expresions (in the order they are specified).
+
+        The default value for this parameter can be set with
+        `stbt.set_global_ocr_corrections`. If global corrections have been set
+        *and* this ``corrections`` parameter is specified, the corrections in
+        this parameter are applied first.
+
     | Added in v28: The ``upsample`` and ``text_color`` parameters.
     | Added in v29: The ``text_color_threshold`` parameter.
     | Added in v30: The ``engine`` parameter and support for Tesseract v4.
@@ -287,9 +295,7 @@ def ocr(frame=None, region=Region.ALL,
         tesseract_user_patterns, tesseract_user_words, upsample, text_color,
         text_color_threshold, engine, char_whitelist, imglog)
     text = text.strip().translate(_ocr_transtab)
-
-    if corrections is not None:
-        text = apply_ocr_corrections(text, corrections)
+    text = apply_ocr_corrections(text, corrections)
 
     debug(u"OCR in region %s read '%s'." % (region, text))
     _log_ocr_image_debug(imglog, text)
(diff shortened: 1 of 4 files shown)


drothlis:
It'll get tedious to specify this each time you call `ocr()`. Maybe we should support a config file of replacements, that applies globally.

drothlis:
Should we put the global ocr corrections in stbt.conf?

```
[ocr_corrections]
bad = good

[ocr_corrections_regex]
b.*d = g\1d
```

Problems: ConfigParser keys can't have "=" or "#" (nor, presumably, "[" or "]").

drothlis:
Maybe a separate [TOML] file? TOML supports quoted keys so that the keys can have all sorts of characters in them, even equals signs.

TOML double quotes are like python strings; TOML single quotes are like Python `r'raw'` strings.

[TOML]:  https://github.com/toml-lang/toml

Crazy idea: Leave it in stbt.conf, but parse (just) that section as TOML:

    toml.loads("\n".join("%s=%s" % (k,v) for k,v in c.items("ocr_corrections")))

Wouldn't work with TOML multi-line string literals.

drothlis:
stbt.ocr: Apply corrections from ocr_corrections.toml

This looks for files called "ocr_corrections.toml" anywhere there is a
stbt.conf -- that is, alongside (in the same directory as) any files
listed in $STBT_CONFIG_FILE. These files are optional (non-existent
files are not an error).

TOML is a format like this:

    key = "string value"

Unlike yaml, string values must be quoted. This avoids unexpected
behaviour, like "no" looks like a string but is actually a boolean in
yaml.

The "keys" are the bad text that needs replacing, and "values" are the
new text to replace it with. The "bad" text (keys) are treated as
regular expression patterns that can match anywhere in the tesseract
output. Make sure you include enough boundary characters in the pattern
(like "\b" to anchor at the beginning or end of a word), especially if
the pattern is very short, so that it doesn't match in unexpected
places.

TOML Keys can be quoted too, if they contain characters like an equals
sign:

    "R=PLAC=" = "REPLACE"

Dots are special in TOML key names, so you need to escape those:

    "B.d" = "Bad"

Remember that we treat the keys as regexes, so you will need to escape
any literal dots:

    "B\.d" = "Bad"

drothlis:
@wmanley there are still some outstanding todo items, but can you give this a high-level review please?

drothlis:
Make them not regexes? And match at word boundaries. Something like this:

```
ocr_corrections = {
    "Acfion": "Action",
    ...
}
ocr_corrections_regex = re.compile("|".join(r"\b(" + re.escape(x) + r")\b"
                                            for x in ocr_corrections))

def ocr_corrections_replace(matchobj):
    return ocr_corrections[matchobj.group(0)]

text = re.sub(ocr_corrections_regex, ocr_corrections_replace, text)
```

drothlis:
> Make them not regexes? And match at word boundaries.

I've been using this technique in a customer's test-pack (by defining an `ocr` function that wraps `stbt.ocr`) and it works well -- I haven't needed regexes yet.

wmanley:
For globals I'd suggest steering clear of ini or toml, I think Python is fine.  This allows flexibility and has a well known format with well understood escapes.  I would suggest adding a function:

    def set_global_ocr_corrections(corrections: List[Tuple[str, str]]):

Usage:

```python
stbt.set_global_ocr_corrections([
    ("Foo", "Bar"),
    ("Blah", "Bloo"),
])
```

People can add it to their `tests/__init__.py` so it's initialised at startup.

This provides us scope to expand it in the future.  For example we could have:

    (re.compile("F[0oO]{2}"), "Foo"),

and customers could choose to customise the list based on the STB in use:

```python
if stbt.get_config("device_under_test", "device_type") == "roku":
    stbt.set_global_ocr_corrections([("ab", "cd")])
```

_stbt/ocr.py
```diff
@@ -240,10 +241,16 @@ def ocr(frame=None, region=Region.ALL,
         that tesseract won't think that a zero is the letter o.
         Note that Tesseract 4.0's LSTM engine ignores ``char_whitelist``.
 
+    :param dict corrections:
```
wmanley, line 244: A list of tuples may provide more flexibility.  That way the keys needn't be hashable.  I suppose we could always do:

```
if isinstance(corrections, dict):
    corrections = list(corrections.items())
```

wmanley: Interestingly it turns out that `re.compile` objects are hashable:

```python
>>> hash(re.compile("abc"))
8787775815358
```

_stbt/ocr.py
```diff
@@ -269,6 +276,11 @@ def ocr(frame=None, region=Region.ALL,
         tesseract_user_patterns, tesseract_user_words, upsample, text_color,
         text_color_threshold, engine, char_whitelist, imglog)
     text = text.strip().translate(_ocr_transtab)
+
```
wmanley, line 279:
```suggestion

if isinstance(corrections, dict):
    corrections = list(corrections.items())
```

_stbt/ocr.py
```diff
@@ -269,6 +276,11 @@ def ocr(frame=None, region=Region.ALL,
...
+    if corrections is not None:
+        for bad, good in corrections.items():
+            text = re.sub(bad, good, text)
```
wmanley, line 285: This will apply corrections to already corrected strings.

wmanley:
In your experience did you need to apply corrections to OCR at sub-word boundary?  I'm wondering if requiring word boundaries for all keys (or by default) might make the behaviour more predictable?

wmanley:
Breaking on word-boundaries would also help with applying the implementation to `match_text`.

_stbt/ocr.py
wmanley: We could start by sorting the list of keys by length and applying the longest corrections first - excluding things that were replaced from further replacements.

drothlis:
I have force-pushed the following changes:

- Drop the idea of using toml or ini or any other external config file.
    - In the next commit later today I will add `set_global_ocr_corrections` as per @wmanley's suggestion [above](#issuecomment-608456652).
- Treat strings vs. re patterns (from `re.compile`) differently: With strings we enforce matching at word boundaries.

> In your experience did you need to apply corrections to OCR at sub-word boundary? I'm wondering if requiring word boundaries for all keys (or by default) might make the behaviour more predictable?

In the test-pack I was working on, I used word boundaries (in fact I used literally this new implementation I've pushed for the strings with word boundaries; it just didn't support regexes).

In the test-pack I was working on, I also used this single regex (I had hard-coded it into my private `ocr` function):

    text = re.sub(r"([a-zA-Z]) 8\. ([a-zA-Z])", r"\1 & \2", text)

_stbt/ocr.py
drothlis: I prefer dicts because there's far less syntax characters involved. With Python 3 they're ordered already, and in Python 2 you can use an OrderedDict if you care about the order.

_stbt/ocr.py
drothlis: The new implementation doesn't re-do the plain-text corrections (the ones that add word boundaries). It does for regexes, but it does them in the order they're specified (similar to nginx location patterns, except that nginx stops at the first match).

drothlis:
> For globals I'd suggest steering clear of ini or toml, I think Python is fine. This allows flexibility and has a well known format with well understood escapes. I would suggest adding a function:

Thanks for the suggestion. Imagine having to deal with toml's escaping, on top of regexes' own escaping!

> People can add it to their `tests/__init__.py` so it's initialised at startup.

I have documented this in `set_global_ocr_corrections` docstring.

tests/test_ocr.py
```diff
@@ -185,6 +185,38 @@ def test_char_whitelist():
...
+    ({'O': '0'},                                     'OO'),
+    ({'OO': '00'},                                   '00'),
+    # Strings are case-sensitive, and they aren't regexes:
+    ({'oo': '00', '[oO]': '0'},                      'OO'),
+    # Regexes do match anywhere:
+    ({re.compile('[oO]'): '0'},                      '00'),
+    # Make sure it tries all the patterns:
+    ({'AA': 'BB', 'OO': '00'},                       '00'),
+    ({re.compile('^O'): '1', re.compile('O$'): '2'}, '12'),
+])
```
wmanley, line 203: I'd suggest adding a test that checks the not applying corrections twice behaviour.

Something like:

{"abc 123": "456", "456": "789"}
"abc 123 456"
becomes "456 789"

A test case that demonstrates the order-dependence would be useful too, both of strings and regexes

Also a test for the regex running after the words behaviour.

_stbt/ocr.py
```diff
@@ -363,6 +391,57 @@ def match_text(text, frame=None, region=Region.ALL,
     return result
 
 
+# Python 2.7 & 3.6 have `re._pattern_type` but that will be removed in Python
+# 3.7 where they introduce `re.Pattern`.
```
wmanley, line 395: I'm bemused by Python's attitude towards backward compatibility.  How high can the maintenance burden of adding `_pattern_type = Pattern` be?

_stbt/ocr.py
```diff
@@ -363,6 +391,57 @@ def match_text(text, frame=None, region=Region.ALL,
...
+
+
+def set_global_ocr_corrections(corrections):
+    """Specify default OCR corrections that apply to all calls to `stbt.ocr`
+    and `stbt.apply_ocr_corrections`.
+
+    See the ``corrections`` parameter of `stbt.ocr` for more details.
+
+    We recommend calling this function from ``tests/__init__.py`` to ensure it
+    is called before any test script is executed.
```
wmanley, line 438: I think the recommendation is good, but I'm not sure about the reasoning. I think the advantage of `__init__.py` is not when it's run, it's the fact that it's will be run for every test-case in your test-pack.  Maybe that's what's intended and I'm misreading putting too much emphasis on the word "before".

_stbt/ocr.py
```diff
@@ -363,6 +391,57 @@ def match_text(text, frame=None, region=Region.ALL,
...
+
+
+def apply_ocr_corrections(text, corrections=None):
+    """Applies the same corrections as `stbt.ocr`'s ``corrections`` parameter.
+
+    This is also available as a separate function, so that you can use it to
+    post-process old test artifacts using new corrections. See also
+    `stbt.set_global_ocr_corrections`.
+    """
+    if corrections:
```
wmanley, line 406: I think this should be `corrections is not None`.  That way you can pass `corrections={}` to disable corrections.

_stbt/ocr.py
```diff
@@ -240,10 +242,34 @@ def ocr(frame=None, region=Region.ALL,
         that tesseract won't think that a zero is the letter o.
         Note that Tesseract 4.0's LSTM engine ignores ``char_whitelist``.
 
+    :param dict corrections:
+        Dictionary of corrections to replace known OCR mis-reads. Each key of
+        the dict is the text to search for; the value is the corrected string
+        to replace the matching key. If the key is a string, it is treated as
+        plain text and it will only match at word boundaries (for example the
+        string ``"he saw"`` won't match ``"the saw"`` nor ``"he saws"``). If
```
wmanley, line 250: I like this example.

_stbt/ocr.py
wmanley: If Python 2 users are going to need to use `OrderedDict` I think this should be explicitly called out in the documentation.  It can be tricky to get right.  For example:

Looks right at first glance, but is not:

```py
>>> collections.OrderedDict({"a": 1, "b": 2, "c": 3})
OrderedDict([('a', 1), ('c', 3), ('b', 2)])
```

Correct:

```py
>>> collections.OrderedDict([("a", 1), ("b", 2), ("c", 3)])
OrderedDict([('a', 1), ('b', 2), ('c', 3)])
```

tests/test_ocr.py
```diff
@@ -185,6 +185,38 @@ def test_char_whitelist():
...
+def test_corrections(corrections, expected):
+    f = load_image('ocr/00.png')
+    print(corrections)
+    assert expected == stbt.ocr(frame=f, mode=stbt.OcrMode.SINGLE_WORD,
```
wmanley, line 207: I think you definitely want some tests with multiple words.  Perhaps test `apply_ocr_corrections` directly.  You only need integration tests involving `ocr` (i.e. just to make sure `apply_ocr_corrections` is being called), but it's easy to add more thorough tests by calling `apply_ocr_corrections`.

_stbt/ocr.py
```diff
@@ -363,6 +391,57 @@ def match_text(text, frame=None, region=Region.ALL,
...
+    for k, v in corrections.items():
+        if isinstance(k, PatternType):
+            text = re.sub(k, v, text)
+    return text
+
+
+global_ocr_corrections = {}
+
+
+def set_global_ocr_corrections(corrections):
```
wmanley, line 431: An alternative to this function is  (and to config variables affecting `stbt` in general) is a module/object like `stbt.defaults`.  Then instead of configuring say match parameters in the config file you could do it in Python.  Something like:

```py
# Setup stbt defaults:
stbt.defaults.ocr_corrections = ...
stbt.defaults.match_parameters = ...
```

This would have the advantage that we could apply it to our other knobs more uniformly.  On the down-side it feels like there's more coupling here.

We can cross that bridge if/when we come to it though.

PR #638, opened 2019-11-14
Add properties to Position for compatibility with Region

wmanley:
This will allow treating `Position` and `Region` polymorphically in
some cases.

### DIFF
--- a/_stbt/types.py
+++ b/_stbt/types.py
@@ -12,8 +12,38 @@ from collections import namedtuple
 
 
 class Position(namedtuple('Position', 'x y')):
-    """A point with ``x`` and ``y`` coordinates."""
-    pass
+    """A point with ``x`` and ``y`` coordinates.
+
+    >>> Region.bounding_box(Position(1, 8), Position(7, 3))
+    Region(x=1, y=3, right=7, bottom=8)
+    >>> b = Region(4, 4, right=13, bottom=10)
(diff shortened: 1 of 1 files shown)


_stbt/types.py
```diff
@@ -12,8 +12,38 @@
...
+
+    >>> Region.bounding_box(Position(1, 8), Position(7, 3))
+    Region(x=1, y=3, right=7, bottom=8)
+    >>> b = Region(4, 4, right=13, bottom=10)
+    >>> b.contains(Position(1, 3))
+    False
+    >>> b.contains(Position(7, 8))
+    True
+    >>> b.contains(Position(13, 10))
+    True
```
drothlis: This should be False, because Region is a half-open interval `[x, right)` and `[y, bottom)`.

drothlis:
Does it make sense for any other Region APIs to interact with Positions? Here are all the stbt.Region APIs:

| API | Makes sense |
|---|---|
| Region.intersect(region, position) | ❌ |
| Region.bounding_box(region, position) | ❌ |
| region.contains(position) | ✔ - alternate implementation (by making `Region.contains` handle Positions explicitly) in #651 |
| region.translate(position) | ✔ - already supported but no tests |

drothlis:
Closing - I've merged #651 instead.

PR #639, opened 2019-11-20
Allow utf-8 paths in $STBT_CONFIG_FILE

drothlis:


drothlis:
This was a regression in v31.

PR #640, opened 2019-11-27
Add setup.py to build pure-Python package for PyPI

drothlis:
I don't know much about Python packaging but I followed these guides
by the official Python Packaging Authority:

https://packaging.python.org/tutorials/packaging-projects/
https://packaging.python.org/guides/distributing-packages-using-setuptools

This package doesn't include the command-line tools (like `stbt run`)
but it allows importing stbt as a library from any Python code. The
primary motivation is to make the stbt library easy to install for IDE
linting & autocompletion.

OpenCV & numpy dependencies are pulled in via `pip install`'s dependency
resolution. The official opencv-python & numpy packages on PyPI take
care of building/packaging the corresponding C extensions, so this
should work on Linux, MacOS, and Windows.

This doesn't include GStreamer & related dependencies, so
`stbt.get_frame()` won't work, but you'll be able to call `stbt.match()`
if you specify a frame explicitly (e.g. loaded from disk with
`stbt.load_image()`).

It doesn't include LIRC or CEC integrations; `stbt.press()` doesn't
work.

It doesn't include any C extensions, so the following are disabled:

- `stbt.match` fast path.
- `imgproc_cache` cache (requires LMDB & xxHash).

`stbt.ocr()` won't work, unless you install tesseract manually. You have
to ensure the right tesseract version and the right tesseract data files
are installed, if you want consistent behaviour between your PC and the
Stb-tester Node.


drothlis:
I have published it to PyPI (from commit 1b5035e6) if you want to test it:
https://pypi.org/project/stb-tester/

drothlis:
We can call the next version "32", it doesn't need to have a minor version. But if we need to do packaging fixes then we need to add a minor version -- hope it doesn't confuse our users.

drothlis:
Future TODOs:

- `*.pyi` files?
- Type annotations

drothlis:
## Testing on MacOS

I have MacOS 10.14 (just over a year old but still getting point releases).
Latest is MacOS 10.15 (released June 2019).

### Python3

```
$ pip3 install opencv-python~=3.2.0
ERROR: Could not find a version that satisfies the requirement opencv-python~=3.2.0 (from versions: 3.4.2.16, 3.4.2.17, 3.4.3.18, 3.4.4.19, 3.4.5.20, 3.4.6.27, 3.4.7.28, 3.4.8.29, 4.0.0.21, 4.0.1.24, 4.1.0.25, 4.1.1.26, 4.1.2.30)
ERROR: No matching distribution found for opencv-python~=3.2.0

$ pip3 install opencv-python~=3.2
Collecting opencv-python~=3.2
  Downloading https://files.pythonhosted.org/packages/d7/f5/80ad99888a640b99ba1192ce34eff68781561844c6a2fc3a8d4acd975771/opencv_python-3.4.8.29-cp37-cp37m-macosx_10_9_x86_64.whl (45.1MB)
     |████████████████████████████████| 45.1MB 836kB/s 
Collecting numpy>=1.14.5
  Downloading https://files.pythonhosted.org/packages/60/9a/a6b3168f2194fb468dcc4cf54c8344d1f514935006c3347ede198e968cb0/numpy-1.17.4-cp37-cp37m-macosx_10_9_x86_64.whl (15.1MB)
     |████████████████████████████████| 15.1MB 16.0MB/s 
Installing collected packages: numpy, opencv-python
Successfully installed numpy-1.17.4 opencv-python-3.4.8.29
```

The latest 3.2 release on PyPI is 3.2.0.8 from July 2018 (a year before MacOS 10.14).

All the other packages in setup.py installed ok with pip3 (I didn't try numpy because it's pulled in by opencv-python).

Conclusion: We need CI for the pure-python package on MacOS (and Windows too).

drothlis:
I've dropped "Windows & MacOS CI" from the todo list for now, in the interest of getting this merged. I think at least the package will install on both those platforms now. Any breakages are more likely to be caused by new MacOS versions and/or new versions of the dependencies, and we'd need a different approach to CI to catch such issues.

drothlis:
I have published 31.1.1 (see 12fdb97) to PyPI.

drothlis:
I have published 3.1.2 (fdef5d3) to PyPI.

.circleci/python-only.dockerfile
```diff
@@ -11,6 +11,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         gzip \
         language-pack-en \
         pylint3 \
+        python3-mock \
```
wmanley: IIRC `unittest.mock` is in the standard library in Python 3.

drothlis: Fixed in 4442d3e. Thanks!

drothlis:
I was going to use [pypi-alias] to create a dummy `stbt` distribution, but there is already one by that name on PyPI: https://pypi.org/project/stbt/ -- "simple trading backtester".

[pypi-alias]: https://pypi.org/project/pypi-alias/

PR #641, opened 2019-11-29
setup.py: Upgrade Jinja2 dependency

drothlis:
Fixes GitHub warning:

> CVE-2019-10906
> high severity
> Vulnerable versions: < 2.10.1
> Patched version: 2.10.1
>
> In Pallets Jinja before 2.10.1, str.format_map allows a sandbox escape.

Not that it's particularly relevant as any user-provided code that could
affect the Jinja output (used to generate the detailed stbt-debug html)
comes from the user's test scripts which are run outside Jinja's sandbox
already.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ This package doesn't bundle the Tesseract OCR engine, so `stbt.ocr()` and
 
 setuptools.setup(
     name="stb-tester",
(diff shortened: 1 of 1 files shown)


drothlis:
Published 31.1.3 to PyPI.

PR #642, opened 2019-11-29
Add .appveyor.yml to run CI on Windows using free AppVeyor service

drothlis:
TODO:

- [ ] Make the job fail if pytest fails.


wmanley:
Closing.  We have no Windows support planned.

PR #644, opened 2019-12-01
setup.py: Specify pylint & astroid dependencies with pinned versions

drothlis:
So that `pylint --load-plugins=stbt.pylint_plugin` will work without
further installation steps.

I have pinned the versions to match Ubuntu 18.04. The latest versions
(pylint-2.4.4 & astroid-2.3.3) fail with:

>   File "venv/lib/python3.6/site-packages/stbt/pylint_plugin.py", line 23, in <module>
>     from astroid import MANAGER, YES
> ImportError: cannot import name 'YES'

We should probably fix that at some point, but pinning these versions
will avoid problems outside our control whenever there's a new astroid
release. We've had trouble with astroid's backward compatibility before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 .stbt-cflags
 .stbt-ldflags
 .stbt-prefix
+.venv
 __pycache__
 _stbt/libxxhash.so
 _stbt/libstbt.so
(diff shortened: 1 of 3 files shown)


drothlis:
Published 31.1.4 from f622502.

PR #647, opened 2019-12-19
stbt.wait_until: Allow the callable to take an optional "frame" parameter

drothlis:
Previously, you might write code like this:

    assert wait_until(lambda: stbt.match("image1.png") or
                              stbt.match("image2.png"))

Since we're not specifying `frame`, `stbt.match` will call
`stbt.get_frame`. Two different undesirable things can happen:

1. `stbt.match("image1.png")` and `stbt.match("image2.png")` might end
   up operating on different frames.

2. Since `stbt.get_frame` can return the same frame multiple times,
   several iterations of the `wait_until` loop might end up operating on
   the same frame.

By iterating over `stbt.frames()` instead, we guarantee that in a single
iteration of `wait_until`'s loop we see the same frame[1]; and we
guarantee that the next iteration of `wait_until`'s loop will block (if
necessary) until the next frame is available.

I've been meaning to implement this for a while, though I do wonder if
the "undesirable thing # 1" above would really happen in practice,
considering that you really should be using a PageObject property for
complicated matching like this. Still, # 2 above might be reason enough
to implement this.

[1] To really guarantee this you have to write your lambda function to
    pass `frame` through correctly.

### DIFF
--- a/_stbt/wait.py
+++ b/_stbt/wait.py
@@ -10,6 +10,11 @@ import inspect
 
 from .logging import debug
 
+try:
+    from inspect import getfullargspec as getargspec
+except ImportError:  # Python 2
+    from inspect import getargspec
+
 
 def wait_until(callable_, timeout_secs=10, interval_secs=0, predicate=None,
                stable_secs=0):
@@ -22,7 +27,8 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, predicate=None,
     .. _truthy: https://docs.python.org/3.6/library/stdtypes.html#truth-value-testing
 
     :param callable_: any Python callable (such as a function or a lambda
-        expression) with no arguments.
+        expression) with no arguments, or with a single argument called
+        ``frame``.
 
     :type timeout_secs: int or float, in seconds
     :param timeout_secs: After this timeout elapses, ``wait_until`` will return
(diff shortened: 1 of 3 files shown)


drothlis:
Note: This is implemented for functions that take a `frame`, but not for classes whose constructor takes a `frame`. On Python 3 you can do `inspect.getfullargspec(MyClass)` but on Python 2 that raises "TypeError: <class 'MyClass'> is not a Python function". On both Python 2 & 3 you can do `getargspec(MyClass.__init__)`.

drothlis:
> On both Python 2 & 3 you can do `getargspec(MyClass.__init__)`.

But it doesn't work with all classes on Python 2:

```
In [10]: class Dave(object):
    ...:     pass
    ...: 
In [15]: inspect.getargspec(Dave.__init__)
TypeError                                 Traceback (most recent call last)
<ipython-input-15-9ebe67f66381> in <module>()
----> 1 inspect.getargspec(Dave.__init__)

/usr/lib/python2.7/inspect.pyc in getargspec(func)
    823         func = func.im_func
    824     if not isfunction(func):
--> 825         raise TypeError('{!r} is not a Python function'.format(func))
    826     args, varargs, varkw = getargs(func.func_code)
    827     return ArgSpec(args, varargs, varkw, func.func_defaults)

TypeError: <slot wrapper '__init__' of 'object' objects> is not a Python function
```

drothlis:
Closing. Not worth the complexity / effort.

PR #649, opened 2020-01-22
setup.py: dependency stbt-extra-stubs has been renamed to stbt-premium-stubs

drothlis:


drothlis:
Published: https://pypi.org/project/stb-tester/31.1.5/

Issue #648, opened 2019-12-22
OpenCV 4 Support


eugenetriguba:
I've been using the library with OpenCV 4, but it just required some small changes to ``cv2_compat.py`` to get things working. I could submit a pull request for it, just wanted to gauge if this is something that would be desired.

https://stackoverflow.com/questions/54734538/opencv-assertion-failed-215assertion-failed-npoints-0-depth-cv-32

The stack overflow post explains the issue, related to changes to what ``findContours`` returns.

drothlis:
> I could submit a pull request for it

Yes please.

demorph:
cv2_compat.py : line 18
contours = cv2.findContours(image=image, mode=mode, method=method)[0]
fixed it  =)

drothlis:
Fixed in #650.

PR #650, opened 2020-01-30
Add support for OpenCV4

eugenetriguba:
This is in regards to #648. Sorry for the delay.

OpenCV4 changed the behavior of `findContours` back to returning
`(contours, hierachy)` instead of `(image, contours, hierachy)`.
However, the `cv2_compat` file is trying to find contours as the second
return value for any version of cv2 >= 3.2.0.

This is a small fix that ensures contours is still selected
on version of openCV >= 4.0.0.


drothlis:
Merged. Thank you very much!

We'll need to add OpenCV4 to our CI (maybe once Ubuntu 20.04 comes out).

PR #651, opened 2020-02-11
stbt.Region.contains: Accept a `Position` as the argument

drothlis:
Sometimes when you've identified a selection (with stbt.match or some
other mechanism) you want to tell which row/cell it corresponds to (in a
table/grid) and it doesn't matter if the selection is slightly larger
than the cell, so you use `cell.contains(selection.region.center)`.

### DIFF
--- a/_stbt/types.py
+++ b/_stbt/types.py
@@ -86,8 +86,10 @@ class Region(with_metaclass(_RegionClsMethods,
     10
     >>> b.center
     Position(x=8, y=7)
-    >>> b.contains(c), a.contains(b), c.contains(b)
-    (True, False, False)
+    >>> b.contains(c), a.contains(b), c.contains(b), c.contains(None)
+    (True, False, False, False)
+    >>> b.contains(c.center), a.contains(b.center)
+    (True, False)
     >>> b.extend(x=6, bottom=-4) == c
     True
     >>> a.extend(right=5).contains(c)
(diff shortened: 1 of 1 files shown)


drothlis:
See also #638.

drothlis:
@wmanley what do you think? I don't know if the half-open interval problem I had with #638 is fundamentally difficult to deal with (in which case it'd be worth merging this now as a small improvement) or if we should get #638 working instead.

wmanley:
Hmm, what is the difference between a `Position` and a `Region` with `width=1` and `height=1`?  Does a position refer to a pixel?  Is it just to make destructuring easier?

drothlis:
Another alternative proposed by @wmanley: Change `region.center` to return a `Region` with width & height of 1px.

This would break destructuring (`x, y = r.center`) but would fix the immediate use case of `r1.contains(r2.center)` without having to come up with some grand model of how Regions & Positions should interact generally.

_stbt/types.py
```diff
@@ -272,9 +272,18 @@ def to_slice(self):
...
+        """
+        if all(hasattr(other, a) for a in ("x", "y", "right", "bottom")):
+            # a Region
+            return (self.x <= other.x and self.y <= other.y and
+                    self.right >= other.right and self.bottom >= other.bottom)
+        elif hasattr(other, "x") and hasattr(other, "y"):  # a Position
+            return (self.x <= other.x <= self.right and
+                    self.y <= other.y <= self.bottom)
+        else:
+            return False
```
drothlis: This should probably raise instead of returning False.

drothlis:
I still prefer the change as implemented in this pull request -- it has the least side effects.

_stbt/types.py
drothlis: Done. It still returns False if other=None. Unlike the previous implementation it checks explicitly for `None` instead of `not other`, because numpy arrays would raise a confusing error message.

drothlis:
Turns out my implementation had the same half-open behaviour that I complained about in my review of #638. Fixed and added a test.

PR #653, opened 2020-02-14
wait_until: Don't run `predicate` if `value` is falsey

drothlis:
This avoids having to write awkward code like:

    wait_until(MyPageObject, predicate p: p and p.some_property.something)

...where we need the `p and` to guard against falsey FrameObjects where
`p.some_property` will be `None`.

TODO:

- [ ] Decide if this is a good idea.
- [ ] Tests.
- [ ] Documentation.

### DIFF
--- a/_stbt/wait.py
+++ b/_stbt/wait.py
@@ -109,7 +109,10 @@ def wait_until(callable_, timeout_secs=10, interval_secs=0, predicate=None,
     while True:
         t = time.time()
         value = callable_()
-        predicate_value = predicate(value)
+        if value:
+            predicate_value = predicate(value)
+        else:
+            predicate_value = None
 
         if stable_secs:
             if predicate_value != stable_predicate_value:


drothlis:
Hmmm but this will break use cases where you're using `predicate` to wait for, and return, some specific kind of falsey value. I wonder if anyone's actually using it for that.

drothlis:
An alternative solution is #441 where we could ignore exceptions like `AttributeError`.

drothlis:
Counterexample: Yesterday I wrote some code like this:

```
# page : Menu
previous_selection = page.selection.text
stbt.press("KEY_OK")
page = stbt.wait_until(
    Menu,
    predicate=lambda p: (
        not p or (p.selection.text != previous_selection and
                  p.selection.text)),
    stable_secs=1)
if page:
    ...  # on sub-Menu
else:
    ...  # Opened the menu and landed on a different page
```

I was on a Menu and pressing OK will open a submenu (which looks identical, thus it's also modelled by the same Menu page object) or else open the selected thing and land on an entirely different page. There's no way of telling (just from the screen) if the current item will open a submenu or not. So I used the above `wait_until` to wait for the menu selection to change, or for there to be something that isn't a menu. I couldn't use `press_and_wait` for various reasons.

In the end I didn't use the above code, but it shows we shouldn't change wait_until's behaviour.

PR #654, opened 2020-02-19
stbt.draw_text: Also write text to stderr

drothlis:
You'll want to see this message in the logs too, and you don't want to have to write a `stbt.debug(...)` for every `stbt.draw_text(...)` in your test.


drothlis:
Note: There are no internal uses of `stbt.draw_text` (I checked in case of duplicated debug). Internally we use `_stbt.core.DeviceUnderTest.draw_text` directly.

wmanley:
Looks useful. I want to add a `time` parameter to `draw_text` too, so logging related to a particular frame is drawn at the right time.

PR #659, opened 2020-03-06
stbt.load_image: Allow passing an image

drothlis:
This simplifies writing functions like this:

    def my_helper(mask=None):
        if mask:
            mask = stbt.load_image(mask)

...such that `mask` can be a filename, or an image constructed in code
using `numpy.zeros` etc.


wmanley:
One possible issue: we use `flags` to specify the format of the expected image.  e.g. alpha-channel or not, etc.  `load_image` won't do this conversion.  Not a big issue, we can fix this in the future if needed.

PR #661, opened 2020-03-19
docs: Use time from `stbt.press`

wmanley:
This is best practice because it may be more accurate.

### DIFF
--- a/_stbt/transition.py
+++ b/_stbt/transition.py
@@ -107,12 +107,11 @@ def wait_for_transition_to_end(
     transition you can use `wait_for_transition_to_end` as the last
     measurement. For example::
 
-        stbt.press("KEY_OK")  # Launch my app
-        press_time = time.time()
+        keypress = stbt.press("KEY_OK")  # Launch my app
         m = stbt.wait_for_match("my-app-home-screen.png")
-        time_to_first_frame = m.time - press_time
+        time_to_first_frame = m.time - keypress.start_time
         end = wait_for_transition_to_end(m.frame)
-        time_to_fully_populated = end.end_time - press_time
+        time_to_fully_populated = end.end_time - keypress.start_time
 
     :param stbt.Frame initial_frame: The frame of video when the transition
         started. If `None`, we'll pull a new frame from the device under test.


_stbt/transition.py
```diff
@@ -107,8 +107,7 @@ def wait_for_transition_to_end(
     transition you can use `wait_for_transition_to_end` as the last
     measurement. For example::
 
-        stbt.press("KEY_OK")  # Launch my app
-        press_time = time.time()
+        press_time = stbt.press("KEY_OK").start_time  # Launch my app
```
drothlis: It looks a bit clumsy and it obscures the action (the keypress) among a lot of code. This is *slightly* better:

    keypress = stbt.press("KEY_OK")
    ...
    time_to_first_frame = m.time - keypress.start_time

Even better would be to have `stbt.last_keypress` which gets set automatically. We already set `_stbt.core.DeviceUnderTest._last_keypress`, but we don't expose it in a public API. I'm having a go at this in #662. In the meantime we can merge this.

I had a go at making a descriptor (property) on the stbt module itself, so that we could have `stbt.last_keypress` return the latest value of `_stbt.core.DeviceUnderTest._last_keypress`, but you can't have properties on modules. See [PEP 549]:

> Python's descriptor protocol requires that descriptors be members of the *type* of an object. This PEP proposes an extension to the descriptor protocol allowing use of the descriptor protocol for members of *instances*. This would permit using properties in modules.

PEP 549 was rejected though. It does show a workaround, but it's 

[PEP 549]: https://www.python.org/dev/peps/pep-0549/

drothlis: Also: It should probably be `end_time` instead of `start_time`.

wmanley: > Also: It should probably be `end_time` instead of `start_time`.

What's the rationale for using `end_time`?

I put `start_time` because I've got more confidence that `start_time` would be accurate, i.e. actually reflects the time the pulse that was sent started to send.  I'm less confident in what `end_time` actually means.  i.e. when does lirc send the ACK for the press command.  This isn't a good rationale really though because I've not measured either.

We might be able to measure this if we used the GPS PPS signal into one channel of the oscilloscope and had the other on the IR diode.  We would then trigger the oscilloscope on the PPS.  We would write a test script to attempt to send a pulse on the second and compare `start_time` against what we see on the oscilloscope.  This assumes that the node has good NTP sync against the GPS.

drothlis: I was thinking it isn't "fair" to the device-under-test to measure its reaction time until it has seen the entire keypress. But upon further reflection, that's just silly. You're right, the start time will be more precise.

wmanley: Fixed.  Merging.

PR #662, opened 2020-03-19
Add stbt.last_keypress()

drothlis:
We already set `_stbt.core.DeviceUnderTest._last_keypress`, but we don't expose it in a public API.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -118,6 +118,9 @@ class DeviceUnderTest(object):
             self._mainloop.__exit__(exc_type, exc_value, tb)
         self._control = None
 
+    def last_keypress(self):
+        return self._last_keypress
+
     def press(self, key, interpress_delay_secs=None, hold_secs=None):
         if hold_secs is not None and hold_secs > 60:
             # You must ensure that lircd's --repeat-max is set high enough.
--- a/stbt/__init__.py
+++ b/stbt/__init__.py
@@ -90,6 +90,7 @@ __all__ = [
     "grid_to_navigation_graph",
     "is_screen_black",
     "Keyboard",
+    "last_keypress",
     "load_image",
     "match",
     "match_all",
@@ -123,6 +124,15 @@ __all__ = [
 
 # Functions available to stbt scripts
 # ===========================================================================
+
+
+def last_keypress():
+    """Returns information about the last key-press sent to the device under
+    test.
+
+    See the return type of `stbt.press`.
+    """
+    return _dut.last_keypress()
 
 
 def press(key, interpress_delay_secs=None, hold_secs=None):
@@ -279,6 +289,9 @@ def get_frame():
 
 class UnconfiguredDeviceUnderTest(object):
     # pylint:disable=unused-argument
+    def last_keypress(self):
+        return None
+
     def press(self, *args, **kwargs):
         raise RuntimeError(
             "stbt.press isn't configured to run on your hardware")
--- a/tests/test-stbt-py.sh
+++ b/tests/test-stbt-py.sh
@@ -600,29 +600,29 @@ test_that_press_returns_a_pressresult() {
 	import time
 	
 	before = time.time()
-	result = stbt.press("KEY_MENU")
+	result1 = stbt.press("KEY_MENU")
 	after = time.time()
 
-	assert before < result.start_time < result.end_time < after
-	assert result.key == "KEY_MENU"
-	assert isinstance(result.frame_before, stbt.Frame)
+	assert before < result1.start_time < result1.end_time < after
+	assert result1.key == "KEY_MENU"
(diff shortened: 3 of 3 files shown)


drothlis:
This doesn't work (as you can see in the failing `test_that_press_returns_a_pressresult`).

I thought I'd be clever by making `stbt.last_keypress` be a [descriptor] on the `stbt` module (this is how `@property` works). Unfortunately it doesn't work.

[descriptor]: https://docs.python.org/3.6/howto/descriptor.html

drothlis:
You can't have properties on modules. See [PEP 549]:

> Python's descriptor protocol requires that descriptors be members of the *type* of an object. This PEP proposes an extension to the descriptor protocol allowing use of the descriptor protocol for members of *instances*. This would permit using properties in modules.

PEP 549 was rejected though. It does show a workaround, but it requires Python ≥ 3.5:

> Adding support for module properties in pure Python has recently become *possible*; as of Python 3.5, Python permits assigning to the `__class__` attribute of module objects, specifically for this purpose. Here's an example of using this functionality to add a property to a module:
> 
> ```
> import sys, types
> class _MyModuleType(types.ModuleType):
>     @property
>     def prop(self, instance, owner):
>         ...
>
> sys.modules[__name__].__class__ = _MyModuleType
> ```

[PEP 549]: https://www.python.org/dev/peps/pep-0549/

drothlis:
A simple work around would be to make it `stbt.last_keypress()`.

Or, we could make it a property using the technique in the previous comment, and say that it's only supported on Python 3.

wmanley:
I think a function would be fine.

wmanley:
An alternative would be to assign to `stbt.last_keypress` if the DUT in use is `stbt._dut`.

drothlis:
[PEP 562] adds `__getattr__` to modules, but it's Python ≥ 3.7.

[PEP 562]: https://www.python.org/dev/peps/pep-0562/

drothlis:
For the record, this was my previous attempt to make `stbt.last_keypress` a module property rather than a function. It works, but requires Python ≥ 3.5. The technique was described in [PEP 549] (which was rejected, but in the "Rationale" section it shows how you could already do it since Python 3.5).

[PEP 549]: https://www.python.org/dev/peps/pep-0549/#rationale

```
diff --git a/_stbt/core.py b/_stbt/core.py
index 1b48c5ad..259a8931 100644
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -99,7 +99,7 @@ class DeviceUnderTest(object):
         self._sink_pipeline = sink_pipeline
         self._mainloop = mainloop
         self._time = _time
-        self._last_keypress = None
+        self.last_keypress = None
 
     def __enter__(self):
         if self._display:
@@ -133,7 +133,7 @@ class DeviceUnderTest(object):
                 self._control.press(key)
                 out.end_time = self._time.time()
             self.draw_text(key, duration_secs=3)
-            self._last_keypress = out
+            self.last_keypress = out
             return out
         else:
             with self.pressing(key, interpress_delay_secs) as out:
@@ -147,7 +147,7 @@ class DeviceUnderTest(object):
             try:
                 self._control.keydown(key)
                 self.draw_text("Holding %s" % key, duration_secs=3)
-                self._last_keypress = out
+                self.last_keypress = out
                 yield out
             except:  # pylint:disable=bare-except
                 exc_info = sys.exc_info()
diff --git a/stbt/__init__.py b/stbt/__init__.py
index 612a6811..c2ffac82 100644
--- a/stbt/__init__.py
+++ b/stbt/__init__.py
@@ -15,6 +15,8 @@ from __future__ import print_function
 from __future__ import division
 from builtins import *  # pylint:disable=redefined-builtin,unused-wildcard-import,wildcard-import,wrong-import-order
 
+import sys
+import types
 
 from _stbt.black import (
@@ -90,6 +92,7 @@ __all__ = [
     "grid_to_navigation_graph",
     "is_screen_black",
     "Keyboard",
+    "last_keypress",
     "load_image",
     "match",
     "match_all",
@@ -278,6 +281,9 @@ def get_frame():
 # ===========================================================================
 
 class UnconfiguredDeviceUnderTest(object):
+    def __init__(self):
+        self.last_keypress = None
+
     # pylint:disable=unused-argument
     def press(self, *args, **kwargs):
         raise RuntimeError(
@@ -316,3 +322,12 @@ def _set_dut_singleton(dut):
         yield dut
     finally:
         _dut = old_dut
+
+
+if sys.version_info.major > 2:
+    class _StbtModuleType(types.ModuleType):
+        @property
+        def last_keypress(self):
+            return _dut.last_keypress
+
+    sys.modules[__name__].__class__ = _StbtModuleType
```

drothlis:
I have made it a function.

wmanley:
Yes, this is going to come in handy.  Release notes need updating, but we can do that at release time.  Merging.

PR #663, opened 2020-03-24
ImageLogger: Add support for Jupyter Notebooks

drothlis:
If running inside a Jupyter Notebook, display debug images inline.

I don't know what to do about the HTML output (including regions we draw
over an image using CSS) but at least we'll see some of the debug
information.

### DIFF
--- a/_stbt/logging.py
+++ b/_stbt/logging.py
@@ -99,7 +99,6 @@ class ImageLogger(object):
             return
 
         self.images = OrderedDict()
-        self.pyramid_levels = set()
         self.data = {}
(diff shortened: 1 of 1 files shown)


drothlis:
I have changed the implementation so that it displays the stbt-debug html in an iframe (before I force-pushed this branch, it was only displaying the images inline with `display(Image(...))`.

drothlis:
It looks like this:

jupyter

wmanley:
Yeah, this could be very useful when developing image processing functions.  Merging.

PR #664, opened 2020-04-02
stbt.match: Disable pyramid optimisation if too few non-transparent pixels

drothlis:
This fixes false negatives (first pass match threshold < 0.96) at the
smaller pyramid levels when the reference image is mostly
transparent (for example a thin border of opaque pixels around a large
transparent centre).

I chose 400 because it's 20x20 (20 is the threshold we already use for
the minimum size in each dimension for opaque reference images). We
divide by 3 because the mask has 3 channels by this point (for
compatibility with `cv2.matchTemplate`).

Note: As a tiny performance optimisation we build mask_pyramid first
because it's the most likely to result in less levels than requested.
Then we build template_pyramid, skipping levels that were too small for
the mask (this is the performance optimisation: avoid unnecessary
downsampling operations). This might result in fewer levels than
mask_pyramid, if mask was None. Then we build the image_pyramid,
skipping levels that were too small for the template. Previously there
was a bug where we were always building the full image_pyramid (that is,
downsampling the frame 3 times) when there was no transparency (no mask)
even if the template_pyramid had fewer levels, because in that case
mask_pyramid is `[None] * levels`.


drothlis:
Note: The existing testcase (previously called "test_transparent_reference_image_with_hard_edge", now part of "test_transparent_reference_image_false_negative_caused_by_pyramid") is still triggered, so commit 0a4e6a1dc249 (#591) is still necessary. I checked by running `stbt_match.py` manually with additional debug, before & after this commit.

Patch for the additional debug:

    diff --git a/_stbt/match.py b/_stbt/match.py
    index 68f0eef8..8ac8ea22 100644
    --- a/_stbt/match.py
    +++ b/_stbt/match.py
    @@ -777,7 +777,7 @@ def _build_pyramid(image, levels, is_template=False, is_mask=False):
             return [None] * levels
         pyramid = [image]
         previous = image
    -    for _ in range(levels - 1):
    +    for i in range(levels - 1):
             if any(x < 20 for x in previous.shape[:2]):
                 break
             downsampled = cv2.pyrDown(previous, borderType=cv2.BORDER_REPLICATE)
    @@ -798,7 +798,14 @@ def _build_pyramid(image, levels, is_template=False, is_mask=False):
                 # matched at pixel 0,0 of the frame, so we won't need to adjust
                 # the match position afterwards).
                 downsampled = downsampled[1:, 1:]
    -        if is_mask and numpy.count_nonzero(downsampled) // 3 < 400:
    +        nonzero = numpy.count_nonzero(downsampled) // 3
    +        ddebug("_build_pyramid (%s) level %d: %dx%d (%d nonzero = %d%%)"
    +               % ("mask" if is_mask else "template" if is_template else "frame",
    +                  i + 1,
    +                  downsampled.shape[1], downsampled.shape[0],
    +                  nonzero,
    +                  nonzero * 100 / (downsampled.shape[1] * downsampled.shape[0])))
    +        if is_mask and numpy.count_nonzero(downsampled) == 0:
                 break
             pyramid.append(downsampled)
         return pyramid

The debug output was:

    $ ./stbt_match.py -v tests/images/regression/roku-tile-frame1.png \
                         tests/images/regression/roku-tile-selection.png
    [...]
    stbt_match.py: _build_pyramid (mask) level 1: 105x81 (728 nonzero = 8%)
    stbt_match.py: _build_pyramid (mask) level 2: 52x40 (0 nonzero = 0%)

(Note: The full-sized image is level 0.)

drothlis:
## `roku-tile-selection.png` from #591



* level 0 (original size): 214x166.
* level 1: 105x81 (of which 728 are non-transparent = 8%).
* level 2: 52x40 (0 nonzero = 0%) -- this level is already skipped (since #591) because 100% transparent.

## `xfinity-selection.png` from this pull request



* level 0 (original size): 232x316
* level 1: 114x156 (228 nonzero = 1%) -- below new threshold of 400 non-transparent pixels so this level is skipped. Previously it failed to match with 0.9411.

_stbt/match.py
```diff
@@ -797,7 +798,7 @@ def _build_pyramid(image, levels, is_template=False, is_mask=False):
             # matched at pixel 0,0 of the frame, so we won't need to adjust
             # the match position afterwards).
             downsampled = downsampled[1:, 1:]
-        if is_mask and numpy.count_nonzero(downsampled) == 0:
+        if is_mask and numpy.count_nonzero(downsampled) // 3 < 400:
```
wmanley, line 801: Yes, this seems like a reasonable heuristic.

_stbt/match.py
```diff
@@ -580,9 +580,10 @@ def _find_candidate_matches(image, template, match_parameters, imglog):
     else:
         mask = None
 
-    template_pyramid = _build_pyramid(template, levels, is_template=True)
-    mask_pyramid = _build_pyramid(mask, len(template_pyramid), is_mask=True)
-    image_pyramid = _build_pyramid(image, len(mask_pyramid))
+    mask_pyramid = _build_pyramid(mask, levels, is_mask=True)
```
wmanley, line 583: Right, so now `mask` determines the number of pyramid levels.  Sensible

PR #665, opened 2020-04-02
stbt.pylint_plugin: Support pylint 2 / astroid 2

drothlis:


drothlis:
P.S. This is necessary to support Python 3.7, where pylint 1.8.3 / astroid 1.6.0 fail with `RuntimeError: generator 'infer_name' raised StopIteration`.

stbt/pylint_plugin.py
```diff
@@ -283,10 +283,8 @@ def _in_git_repo():
 def _infer(node):
     try:
         for inferred in node.infer():
-            # Sometimes when `infer()` fails it returns `YES` which returns
-            # True to everything, including `callable()` and
-            # `isinstance(YES, <anything else>)`, so it isn't useful.
-            if inferred == YES:
+            # When `infer()` fails it returns this singleton:
+            if inferred is Uninferable:
```
wmanley, line 287: Does this change work with pylint 1.8.3 too?  I guess that would imply `YES == Uninferable`.  Maybe `Uninferable` is like `ANY` and anything `==` `Uninferable`?

drothlis:
```
$ ipython
In [1]: import astroid

In [2]: astroid.__version__
Out[2]: '1.6.0'

In [3]: astroid.YES
Out[3]: Uninferable
```

wmanley: Hah, I see, even simpler.

drothlis: Yes, it works with pylint 1.8.3, which is still tested in our CI jobs "ubuntu1804-python2" and "ubuntu1804-python3".

PR #666, opened 2020-04-02
FrameObject docs: Link to more modern FrameObject tutorial

wmanley:
in our manual.


wmanley:
Hold off on merging this for sec...

wmanley:
Fixed

drothlis:
I always find it annoying when a list of references links to the same page twice (once at a specific sub-section). I open them all up in tabs and then going through them I'm like... haven't I seen this already? I've pushed a fix and I've also updated the links to docs.python.org to point to a consistent Python version (3.6, what's in Ubuntu 18.04).

PR #668, opened 2020-04-20
sink pipeline: Avoid filling up RAM if something goes wrong

wmanley:
`appsrc` does its own buffering, and won't reject frames even if the sink
pipeline isn't running or is in an error state.  This can lead to memory
exhaustion.  This will at least avoid OOM, even if the sink pipeline will
have to be fixed seperately.

100MB is 1.4s at 720p25.

TODO:

- [x] Test with `identity`

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -350,6 +350,9 @@ class SinkPipeline(object):
         self._time = _time
         self._sample_count = 0
 
+        # Just used for logging:
+        self._appsrc_was_full = False
+
         # The test script can draw on the video, but this happens in a different
         # thread.  We don't know when they're finished drawing so we just give
         # them 0.5s instead.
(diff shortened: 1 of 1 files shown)


wmanley:
I think I can use `identity sleep-time=1000000` in a custom sink pipeline to test this.

wmanley:
Tested with:

    ./stbt_run.py -v \
        --source-pipeline='videotestsrc is-live=true ! video/x-raw,format=BGR,framerate=25/1,width=1280,height=720 ! identity' \
        --sink-pipeline='identity sleep-time=1000000 ! xvimagesink sync=false' \
        sleep.py

Without this patch the RAM usage skyrockets, with this patch it's fairly stable.

drothlis:
Given your last comment I assume the TODO item at the top should be ticked. I'll merge.

PR #671, opened 2020-05-13
Add stbt.MultiPress: Enter text using a numeric keypad

drothlis:
In some apps, the search page allows entering text by pressing the keys
on the remote control's numeric keypad: press the number "2" once for
"A", twice for "B", etc.:

        1.,     ABC2    DEF3
        GHI4    JKL5    MNO6
        PQRS7   TUV8    WXYZ9
              [space]0

This is much easier (and faster) than moving a selection around an
on-screen keyboard.

The keys for A-Z are specified in [ITU E.161]. The standard doesn't
specify where "space" goes, but every implementation I have seen has it
on KEY_0. Punctuation varies across implementations; I have arbitrarily
chosen the behaviour of a Sky HD+ box that I had available for testing,
but you can override this by passing `key_mapping={"KEY_1": "@1.,-_}` to
the `stbt.MultiPress` constructor.

[ITU E.161]: https://www.itu.int/rec/T-REC-E.161

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,7 @@ INSTALL_PYLIB_FILES = \
     _stbt/logging.py \
     _stbt/match.py \
     _stbt/motion.py \
+    _stbt/multipress.py \
     _stbt/ocr.py \
     _stbt/precondition.py \
     _stbt/power.py \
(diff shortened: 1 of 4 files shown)


drothlis:
@wmanley please review. This would've been useful in some test development I did recently.

Disclaimer: I haven't tested it on a real device, as I no longer have access to the device I could've used this on. I have a Roku & an Apple TV but neither of those support multi-press in their Search widgets (in fact they don't have a numeric keypad on their remote control). I'm fairly confident in the implementation given the unit tests, but I don't know if the default `interletter_delay_secs` will be sufficient.

drothlis:
Rebased onto master to fix conflicts.

drothlis:
Merged to avoid future conflicts. We can always change the default `interletter_delay_secs` if it needs to be longer.

PR #672, opened 2020-05-13
stbt.load_image: Return stbt.Image (numpy subclass) with additional "filename" attributes

drothlis:


drothlis:
@wmanley please review. It's just a tidy-up really -- `_load_image` and `_ImageFromUser` have annoyed me for years -- but it could be convenient for reporting image filenames.

I might change `MatchResult.image` to be the actual `Image` instance instead of the filename, but I'd do that in a future pull request in the interest of getting this merged -- I've probably started a half-dozen attempts to do this over the years so I don't want to bank this progress.

drothlis:
One feature we can easily build once we have this, is to add caching to `stbt.load_image` so that code like this only loads the image from disk once:

    stbt.wait_until(lambda: stbt.match("image.png"))

Plus a decorator to add caching to other image-processing functions, for example:

    stbt.wait_until(lambda: stbt.match(
        remove_background("image.png"),
        frame=remove_background(stbt.get_frame())))

In this example, `remove_background` would still need to run against every frame, but it only needs to run once against the reference image "image.png".

PR #677, opened 2020-06-09
Libcec fix

msiden:
Bug fix for allowing cec to function with Python 3 under Ubuntu 20.04 and libcec 4.0.4

Tested and verified in the above environment + with Python 2.7 under Ubuntu 18.04
I have NOT been able to test with Python 2 under Ubuntu 20.04

### DIFF
--- a/_stbt/control_gpl.py
+++ b/_stbt/control_gpl.py
@@ -162,7 +162,7 @@ class HdmiCecControl(object):
             device = self.detect_adapter()
             if device is None:
                 raise HdmiCecError("No adapter found")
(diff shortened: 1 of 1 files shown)


drothlis:
Thanks @msiden!

drothlis:
N.B. The unit tests haven't been fixed for Python 3 (they still use `buffer` and `b"..."`) but they're skipped on our Python3/Ubuntu18.04 CI job, because `import cec` fails on Python 3.

In Ubuntu 20.04 the Python 3 bindings are available in the apt package `python3-cec`.

PR #679, opened 2020-08-10
Rename `stbt` Python module to `stbt_core`

drothlis:


wmanley:
This is straightforward.  Merging.

PR #689, opened 2020-09-09
stbt.FrameObject: Rename `_frame` member to `frame`

drothlis:
Because I'm tired of writing `# pylint:disable=protected-access` all over the place. Surely it's not that uncommon or magical for a base class to provide members of its own.

We still keep `_frame` as an alias for backward compatibility.

### DIFF
--- a/_stbt/frameobject.py
+++ b/_stbt/frameobject.py
@@ -151,7 +151,7 @@ class FrameObject(with_metaclass(_FrameObjectMeta, object)):
       from the frame.
     * Inside each property, when you call an image-processing function (like
       `stbt.match` or `stbt.ocr`) you must specify the parameter
-      ``frame=self._frame``.
+      ``frame=self.frame``.
 
     The following behaviours are provided automatically by the FrameObject
     base class:
@@ -196,6 +196,7 @@ class FrameObject(with_metaclass(_FrameObjectMeta, object)):
             frame = get_frame()
         self.__frame_object_cache = {}
         self.__local = threading.local()
+        self.frame = frame
         self._frame = frame
 
     def __repr__(self):
--- a/_stbt/keyboard.py
+++ b/_stbt/keyboard.py
@@ -213,7 +213,7 @@ class Keyboard(object):
 
                 @property
                 def selection(self):
-                    m = stbt.match("keyboard-selection.png", frame=self._frame)
+                    m = stbt.match("keyboard-selection.png", frame=self.frame)
                     if not m:
                         return stbt.Keyboard.Selection(None, None)
                     try:
--- a/stbt_core/pylint_plugin.py
+++ b/stbt_core/pylint_plugin.py
@@ -48,17 +48,17 @@ class StbtChecker(BaseChecker):
                   '(such as a function or a lambda expression).'),
         'E7004': ('"%s" missing "frame" argument',
                   'stbt-frame-object-missing-frame',
-                  'FrameObject properties must always provide "self._frame" as '
+                  'FrameObject properties must always provide "self.frame" as '
                   'the "frame" parameter to functions such as "stbt.match".'),
         'E7005': ('Image "%s" not committed to git',
                   'stbt-uncommitted-image',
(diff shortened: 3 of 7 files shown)


wmanley:
The reason it's protected is that the behaviour of the methods don't take into account that the actual frame in use is accessible. We pretend that it's not.  This way `__eq__`, `__hash__`, etc. can be defined purely in terms of public methods/properties.

Can you give some examples of where `._frame` is accessed (in private)?  We've talked in the past about various changes we'd make to a `FrameObject` v2.  Maybe this can fit in there.

drothlis:
> The reason it's protected is that the behaviour of the methods don't take into account that the actual frame in use is accessible. We pretend that it's not. This way `__eq__`, `__hash__`, etc. can be defined purely in terms of public methods/properties.

That's the conceptual model, but it isn't true of the current implementation: You can add (non-property) attributes to your subclass and they aren't taken into account for `__eq__`, `__hash__`, etc.

I still think we can expose `frame` publically -- the practical reasons outweight the conceptual model.

> Can you give some examples of where `._frame` is accessed

Here are 4 examples from real-world code:

- `page._frame.time` used for logging / visualisation a la "data.html".

- Methods of a FrameObject that are operating on a FrameObject of the same type, but not `self`. For example `Carousel.iterate_tiles()` that is checking for wrap-around:

  ```
  def iterate_tiles(self):
      page = self
      first_tile = self
      while True:
          yield page
          assert stbt.press_and_wait("KEY_RIGHT")
          page = page.refresh()
          assert page
          if stbt.match(first_tile.get_selected_thumbnail(),
                        frame=page._frame,  # pylint:disable=protected-access
                        region=page.selected_region):
              stbt.draw_text("Reached end of carousel (wrapped around)")
              return
  ```

- Calling `stbt.wait_for_transition_to_end` -- used in a PageObject's `wait_until_visible` factory method:

  ```
  @classmethod
  def wait_until_visible(cls, timeout_secs=10, stable_secs=1):
      page = stbt.wait_until(cls, timeout_secs=timeout_secs)
      if page:
          stbt.wait_for_transition_to_end(
              initial_frame=page._frame,  # pylint:disable=protected-access
              region=getattr(cls, "region", stbt.Region.ALL),
              timeout_secs=timeout_secs,
              stable_secs=stable_secs)
          return page.refresh()
  ```

  In this case it was a workaround for the fact that `press_and_wait` was being fooled into returning too early, after it had detected some movement. There was a brief blip on the selected button when pressing OK, but then no change (for long enough to fool press_and_wait) into returning. Instead, we need to wait for the page to appear (is_visible) and *then* wait for it to populate fully (wait_for_transition_to_end) as it had various bits & pieces that would pop in as the data loaded.

- Some navigation helper functions that are designed to work hand-in-hand with Page Objects:

  ```
  class StreamingMenu(stbt.FrameObject, Movable):
      ...
  
      def move_one(self, key):
          page = self.home_menu.move_one(key)
          specific = StreamingMenu(frame=page._frame)  # pylint:disable=protected-access
          if specific:
              return specific
          else:
              return page
  ```

  I suppose in this case the reason we're accessing a protected member of a different class is that we've modelled some kind of sub-type relationship between HomeMenu and StreamingMenu.

drothlis:
I have thought of an alternate solution: Disable pylint's "protected-access" checker in `.pylintrc`.

One could argue that the leading underscore should be enough of an indicator that you should be careful about what you're doing.

drothlis:
> In this case it was a workaround for the fact that press_and_wait was being fooled into returning too early, after it had detected some movement. There was a brief blip on the selected button when pressing OK, but then no change (for long enough to fool press_and_wait) into returning. Instead, we need to wait for the page to appear (is_visible) and then wait for it to populate fully (wait_for_transition_to_end) as it had various bits & pieces that would pop in as the data loaded.

This is the wrong ticket to talk about this, but I just had a thought about handling this situation: press_and_wait could take a `min_size` argument -- useful when you're waiting for the whole screen to change, e.g. when navigating to a different type of page.

PR #690, opened 2020-09-17
stbt.Frame, stbt.Image: Allow indexing by a stbt.Region 

drothlis:
> stbt.Frame, stbt.Image: Allow indexing by a stbt.Region 

This sounds convenient but it might not be worth the confusion because you can't do it with a native numpy array (for example after calling `cv2.cvtColor` or any other image processing). On the other hand, do we really expect user code to be using cv2 / numpy directly?

> stbt.crop: Implicitly crop at edge of frame, instead of raising

I think this will be worth merging, at least.

### DIFF
--- a/_stbt/imgutils.py
+++ b/_stbt/imgutils.py
@@ -145,10 +145,11 @@ def crop(frame, region):
       of the source frame. This is a view onto the original data, so if you
       want to modify the cropped image call its ``copy()`` method first.
     """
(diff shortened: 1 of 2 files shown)


drothlis:
I have raised #694 with the first commit (`crop` clamps to the edge of the frame).

I'll close this one due to the drawbacks mentioned in the previous comment. It isn't worth the extra code just to show off that I've learned about `__getitem__`. :-)

PR #691, opened 2020-09-17
pylint plugin: Verify that FrameObject refresh() return value is used

drothlis:
It's a mistake to think that `refresh()` modifies the instance in place.
FrameObjects are immutable.


drothlis:
We may also need to revisit this matcher if we ever add the global `stbt.current_page` that we've talked about. But in the meantime it'll catch bugs so I think we should merge it. @wmanley please review.

wmanley:
Really helpful.

PR #692, opened 2020-09-18
WIP stbt.Keyboard and stbt.press_and_wait: `mask` accepts a Region

drothlis:
TODO:

- Implement it.
- Other functions probably need this too.
- Sometimes we want a mask with 1 channel, sometimes with 3? Depends on
  the function, e.g. is it operating on greyscale or not.


drothlis:
Here's how `wait_for_motion` and `is_screen_black` behave since v28:         If you specify both `region` and `mask`, the mask must be the same size as the region.

wmanley:
Closing.  Obsoleted by #712.

PR #693, opened 2020-09-20
stbt.press_and_wait: Add min_size parameter

drothlis:
When testing an Xfinity set-top-box, I think it was, pressing OK on a button caused that button to "blip" briefly for a single frame or 2, then followed by no motion, before actually starting the transition to the new page. Sometimes that period of no motion was longer than 1 second, so it would fool `press_and_wait` into thinking the transition had completed before it had even started.

Setting `stable_secs` to a very large number isn't ideal, as it would slow down all test-runs for the sake of handling this intermittent behaviour.

TODO:

- [x] Implement it.
- [x] Unit tests.

### DIFF
--- a/_stbt/diff.py
+++ b/_stbt/diff.py
@@ -9,7 +9,7 @@ import cv2
 
 from .config import get_config
 from .imgutils import (crop, _frame_repr, pixel_bounding_box, _validate_region)
-from .logging import ImageLogger
+from .logging import ddebug, ImageLogger
 from .types import Region
 
 
@@ -22,10 +22,12 @@ class FrameDiffer(object):
     repeat that work when you need to compare against frame C.
     """
 
-    def __init__(self, initial_frame, region=Region.ALL, mask=None):
+    def __init__(self, initial_frame, region=Region.ALL, mask=None,
+                 min_size=None):
         self.prev_frame = initial_frame
         self.region = _validate_region(self.prev_frame, region)
         self.mask = mask
+        self.min_size = min_size
         if (self.mask is not None and
                 self.mask.shape[:2] != (self.region.height, self.region.width)):
             raise ValueError(
@@ -43,8 +45,8 @@ class MotionDiff(FrameDiffer):
     """The `wait_for_motion` diffing algorithm."""
 
     def __init__(self, initial_frame, region=Region.ALL, mask=None,
-                 noise_threshold=None):
-        super(MotionDiff, self).__init__(initial_frame, region, mask)
+                 min_size=None, noise_threshold=None):
+        super(MotionDiff, self).__init__(initial_frame, region, mask, min_size)
         if noise_threshold is None:
             noise_threshold = get_config(
                 'motion', 'noise_threshold', type_=float)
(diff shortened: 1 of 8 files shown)


drothlis:
Previously to handle this situation I have been writing a pageobject class method like this:

```
@classmethod
def wait_until_visible(cls, timeout_secs=10, stable_secs=1):
    page = stbt.wait_until(cls, timeout_secs=timeout_secs)
    if page:
        stbt.wait_for_transition_to_end(
            initial_frame=page._frame,  # pylint:disable=protected-access
            region=getattr(cls, "region", stbt.Region.ALL),
            timeout_secs=timeout_secs,
            stable_secs=stable_secs)
        return page.refresh()
```

wmanley:
Yeah, this looks like it could be very useful.  The alternative in these cases would be to specify a region in which motion should occur, but it's not always easy to know what region that will be (like with Apple TV carousels).

drothlis:
> It would also be useful for keyboards to ignore the blinking text-position cursor.

Blinking text cursors will already be ignored since #554, thanks to MotionDiff's erode step.

_stbt/diff.py
```diff
@@ -87,7 +90,11 @@ def diff(self, frame):
             # Undo crop:
             out_region = out_region.translate(self.region)
 
-        motion = bool(out_region)
+        motion = bool(out_region and (
+            self.min_size is None or
+            (out_region.width >= self.min_size[0] and
+             out_region.height >= self.min_size[1])))
```
wmanley, line 106: > At first I wasn't sure if `min_size` should affect the decision whether to update `previous_frame`. I decided that it *should* affect the decision -- that is, if we found motion but it's still smaller than `min_size`, we treat it exactly the same as "no motion" and we don't update `previous_frame`. That way, if something is moving along slowly, the differences will eventually exceed `min_size` and count as "motion". This way the algorithm will eventually detect motion regardless of the frame-rate it's running at.

Sounds like the right thing to do.

_stbt/diff.py
```diff
@@ -45,12 +46,22 @@ class MotionDiff(FrameDiffer):
...
                 'motion', 'noise_threshold', type_=float)
         self.noise_threshold = noise_threshold
+
+        if isinstance(erode, numpy.ndarray):  # For power users
+            kernel = erode
+        elif erode:
+            kernel = cv2.getStructuringElement(cv2.MORPH_ELLIPSE, (3, 3))
+        else:
+            kernel = None
+        self.kernel = kernel
```
wmanley, line 63: What's the reasoning behind this. Under what situations do you think it will be useful?

wmanley (approved):

_stbt/diff.py
drothlis: Mainly I was thinking of the fact that the new press_and_wait algorithm will now ignore the blinking text cursor in a search page, due to the erode. That's actually desirable for the purposes of navigating a search keyboard! But it is a difference, and this gives us a way of disabling the erode step without copy-pasting all of MotionDiff.

Also, this is consistent with `find_selection_from_background` (the same parameter name and implementation) and consistent-ish with `match` (which has `erode_steps` to disable the erode in its second pass).

PR #695, opened 2020-09-21
Keyboard: Add Movable interface to be implemented by page objects

wmanley:
This allows page objects to override the mechanism for doing `press`es.


_stbt/keyboard.py
```diff
@@ -604,9 +601,7 @@ def enter_text(self, text, page, verify_every_keypress=False):
         for letter in text:
             page = self.navigate_to({"text": letter},
                                     page, verify_every_keypress)
-            stbt.press_and_wait("KEY_OK", mask=self.mask, stable_secs=0.5,  # pylint:disable=stbt-unused-return-value
-                                timeout_secs=1)
-            page = page.refresh()
+            page = self._move_one(page, "KEY_OK")
```
drothlis, line 604: This is a change in behaviour: We are now asserting the `press_and_wait` result, and we increased `timeout_secs` to 10.

If you have specified a mask that excludes the text box where the typed-text goes —a common thing to exclude, because it usually has a blinking cursor that breaks `press_and_wait` when used for navigation— then this will fail.

It seems that `move_one` isn't the right abstraction for this operation, which isn't about moving (maybe it "moves" —i.e. changes to a different key— if you're in shift mode or if you're pressing OK on a mode-change key, but most of the time it doesn't). Because of the short timeout and lack of `assert`, this use of `press_and_wait` should be OK even for UIs with transparency or some other reason why `press_and_wait` won't work for navigation.

drothlis: Note: This used to be a plain `stbt.press("KEY_OK")` (sometimes with `interpress_delay_secs=1`) before I added support for shift modes (see commit 07555d0). I wonder if it would make sense to check whether the node we're on has an outgoing edge labelled with `key="KEY_OK"` and use that information to choose `press` vs. `press_and_wait`. It would make `enter_text` a tiny bit faster, by not having to call `page.refresh()`.

PR #701, opened 2020-10-26
Logging tweaks

drothlis:


_stbt/imgutils.py
```diff
@@ -119,6 +122,9 @@ def __repr__(self):
         return "<Image(filename=%r, dimensions=%s)>" % (
             self.filename, dimensions)
```
wmanley, line 123: relative_filename would be good here I think.

PR #704, opened 2020-11-06
Image logging: Use relative_filename where appropriate

wmanley:
If an image is in the test pack then we will print the name relative to
the test-pack root.  This is more useful for logging as it's unambiguous
which image is being used.


drothlis:
I hadn't done this initially because I didn't want the value in the debug logging to be different than the real value of the `filename` attribute, and I didn't want to bloat the logging by saying `relative_filename=...` instead of `filename=...`. But now I think it's worth it for the convenience.

PR #706, opened 2020-12-06
stbt.Keyboard: More robust for slow/buggy keyboards

drothlis:
I was testing the Search keyboard on a Comcast Xfinity box. This has the keys in a single line A - Z so there can be a lot of keypresses to get to the next letter. When `verify_every_keypress=False`, the selection can settle on one of the intermediate letters for long enough to trick our `stable_secs=0.5`, causing the test to fail. After the test failed, the selection carried on moving and did reach the expected target.

I tried increasing stable_secs to 1. This made it fail less often, but it still happens sometimes. Setting `verify_every_keypress=True` fixes this problem but makes it intolerably slow. I don't want to increase stable_secs for every letter because this would slow it down unnecessarily in the usual, successful case.

### DIFF
--- a/_stbt/android.py
+++ b/_stbt/android.py
@@ -154,7 +154,7 @@ class AdbDevice(object):
 
         if _config is None:
             import _stbt.config
-            _config = _stbt.config._config_init()  # pylint:disable=protected-access
+            _config = _stbt.config._config_init()
 
         self.adb_server = adb_server or _config.get("android", "adb_server",
                                                     fallback=None)
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -678,7 +678,7 @@ class Display(object):
         frame.flags.writeable = False
 
         # See also: logging.draw_on
-        frame._draw_sink = weakref.ref(self._sink_pipeline)  # pylint: disable=protected-access
+        frame._draw_sink = weakref.ref(self._sink_pipeline)
         self.tell_user_thread(frame)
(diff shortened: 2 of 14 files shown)


drothlis:
I'm having an issue that (I think) is caused by the difference in diff algorithm between press_and_wait (MotionDiff) vs. find_selection_from_background used in the page object's `selection` property.

**stable_frame.png**
stable_frame

**screenshot.png**
screenshot

As you can see, the "selection" bar has moved slightly, and the "?123" text has changed colour, but MotionDiff doesn't consider either of those significant.

The problem is that the `selection` diff finds a region that encompasses both ?123 and A, and fails its max_size requirement.

drothlis:
### MotionDiff

detect_motion: Didn't find motion

ROI Gray:
image

Previous frame ROI Gray:
image

Absolute difference:
image

Threshold (noise_threshold=0.84):
image

Eroded:
image

drothlis:
If I change the MotionDiff algorithm to use colour distance instead of converting to grayscale and taking the absdiff:

Previous frame:
image

Differences:
image

Differences above threshold (25):
image

After opening:
image

drothlis:
#695 would also help here.

drothlis:
Rebased onto master to fix conflicts with #722.

drothlis:
One of our customers has been using this version of stbt.Keyboard for 6 months and it has fixed the issues it was intended to fix, so I'm going to merge it.

PR #707, opened 2020-12-08
MotionDiff: Take color information into account

drothlis:
Previously, MotionDiff converts the 2 frames to grayscale before comparing them. This means that it doesn't catch text in 2 different colors, but with similar overall intensity. The new algorithm does a simple color distance calculation (consistent with `find_selection_from_background` and with OCR's `text_color` parameter) before doing the threshold.

Along with this, the parameter name has changed from `noise_threshold` to `threshold` and its range has changed from [1.0, 0.0] to [0, 255].

I have copied the old implementation as `GrayscaleDiff`. This allows existing users to upgrade to v33 while still keeping the old behaviour, by running the following code early, for example at the top level of `tests/__init__.py` (this is similar to what we did for `press_and_wait` when we changed its algorithm in v32):

    stbt.detect_motion.differ = stbt.GrayscaleDiff

TODO:

- [ ] Tests.
- [ ] Check debug html.
- [ ] Validate threshold with real-life use of `wait_for_motion`.

### DIFF
--- a/_stbt/diff.py
+++ b/_stbt/diff.py
@@ -80,16 +80,147 @@ class FrameDiffer(object):
 
 
 class MotionDiff(FrameDiffer):
-    """The `wait_for_motion` diffing algorithm."""
+    """Compares 2 frames by calculating the color-distance between them.
+
+    The algorithm is a simple color distance calculation. Conceptually, for
+    each pair of corresponding pixels in the 2 frames we perform the following
+    calculations:
+
+    - Calculate the squared difference separately for each color channel.
+    - Sum the 3 results (for the 3 colour channels).
+    - Normalise back to the original range (0-255) by dividing by 3 and taking
+      the square root.
+    - Compare against the specified threshold. Pixel values greater than the
+      threshold are counted as differences (that is, motion).
+
+    Then, an "erode" operation removes differences that are only 1 pixel wide
+    or high. If any differences remain, the 2 frames are considered different.
+
+    This is the default diffing algorithm for `wait_for_motion`,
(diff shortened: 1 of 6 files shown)


drothlis:
The new algorithm is more strict (it will detect motion that before would have been ignored).

Also the threshold of 25 is more strict: (1 - the previous noise_threshold of 0.84) * 255 = 40.8.
I chose 25 for consistency with `find_selection_from_background` and OCR's `text_color_threshold`.

drothlis:
I was having an issue that (I think) is caused by the difference in diff algorithm between press_and_wait (MotionDiff) vs. find_selection_from_background used in the page object's `selection` property.

**stable_frame.png**
stable_frame

**screenshot.png**
screenshot

As you can see, the "selection" bar has moved slightly, and the "?123" text has changed colour, but MotionDiff doesn't consider either of those significant.

The problem is that the `selection` diff finds a region that encompasses both ?123 and A, and fails its max_size requirement.

---

### MotionDiff

detect_motion: Didn't find motion

ROI Gray:
image

Previous frame ROI Gray:
image

Absolute difference:
image

Threshold (noise_threshold=0.84):
image

Eroded:
image

---

If I change the MotionDiff algorithm to use colour distance instead of converting to grayscale and taking the absdiff:

Previous frame:
image

Differences:
image

Differences above threshold (25):
image

After opening:
image

drothlis:
I extracted the GrayscaleDiff refactorings & fixes to #775 (merged); rebased this onto "main" and reduced the scope so that we're only adding the new differ, but not making it the default.

PR #708, opened 2020-12-08
ocr: text_color_threshold=None lets tesseract do the threshold

drothlis:
...after we've done our color distance.

TODO:

- [x] See if doing text_color difference but letting Tesseract do the threshold gives OK result on `tests/ocr/Buy 14.99 HD.png`. **Yes it does**
- [x] Consider changing stbt.ocr's behaviour when `text_color_threshold=None`.
- [x] Do read default value from config, to provide a migration path / backward compatibility when upgrading to v33.
- [x] Find more examples / add unit tests.

Unfortunately letting Tesseract do its own binarisation after applying our text_color color distance, usually has the same problem that we had when we don't do the color distance at all. I have addded tests in the latest commit to demonstrate this.


drothlis:
In `tests/ocr/Buy 14.99 HD.png` there are some very faint pixels around the letters, and Tesseract's binarisation algorithm thinks they are foreground pixels:

image

drothlis:
Letting Tesseract do its own binarisation after applying our `text_color` color distance, usually has the same problem that we had when we don't do the color distance at all. I have addded tests in the latest commit to demonstrate this.

drothlis:
See also #537 where I did exactly the same experiment and reached the same conclusion, 2 years earlier.

PR #709, opened 2020-12-14
pip package: Make lxml and Jinja optional dependencies

wmanley:
Saves having to download and install `libxml2` with pip.  We usually
can't use ocr with pip installed package anyway as we can't express
a dependency on tesseract.


drothlis (approved):
All the existing imports of jinja2 are already inside the logging functions. :+1:

PR #710, opened 2020-12-14
Fix importing stbt_core when libstbt.so isn't built/installed

drothlis:
In 0eb1f63 ("stbt.match: Make the sqdiff C optimisation optional") we catch ImportError when importing `_stbt.sqdiff`, so that we could publish a pure-python package to PyPI without having to worry about C compilation or binary wheels. However it seems that it never worked:

```
tests/test_match.py:17: in <module>
    import stbt_core as stbt
stbt_core/__init__.py:44: in <module>
    from _stbt.match import (
_stbt/match.py:34: in <module>
    from .sqdiff import sqdiff
_stbt/sqdiff.py:18: in <module>
    _libstbt = ctypes.CDLL(_find_file("libstbt.so"))
/usr/lib/python3.6/ctypes/__init__.py:348: in __init__
    self._handle = _dlopen(self._name, mode)
E   OSError: /home/drothlis/work/stb-tester.com/stb-tester/_stbt/libstbt.so: cannot open shared object file: No such file or directory
```

...which makes me wonder what exactly we are testing in the "python-package" CI job‽

- [x] Fix CI so that it fails.
- [ ] Catch OSError when trying to import `_stbt.sqdiff`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,8 +67,8 @@ commands:
             rm -rf stbt_core _stbt &&
             python3 -m venv build_env &&
             source build_env/bin/activate &&
-            pip3 install pytest==4.6.9 pylint==2.4.4 astroid==2.3.3 &&
-            pip3 install dist/*.tar.gz &&
+            pip install pytest==4.6.9 pylint==2.4.4 astroid==2.3.3 &&
+            pip install dist/*.tar.gz &&
             make check-pythonpackage PYTEST=pytest
 
 workflows:
--- a/_stbt/match.py
+++ b/_stbt/match.py
@@ -27,13 +27,9 @@ from .imgutils import (crop, _frame_repr, _image_region, limit_time, load_image,
                        _validate_region)
 from .logging import (_Annotation, ddebug, debug, draw_on, get_debug_level,
                       ImageLogger)
+from .sqdiff import sqdiff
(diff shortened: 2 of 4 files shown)


drothlis:
This is odd... in the CI output we see this pytest command (from `make check-pythonpackage`):

```
export STBT_CONFIG_FILE=$PWD/tests/stbt.conf && \
pytest -vv -rs  \
    tests/subdirectory/test_load_image_from_subdirectory.py \
    tests/test_android.py \
    tests/test_config.py \
    tests/test_core.py \
    tests/test_grid.py \
    tests/test_keyboard.py \
    tests/test_match.py \
    tests/test_motion.py \
    tests/test_transition.py && ...
```

but in the pytest output there's no mention of `test_match.py` (all the other ones are mentioned).

drothlis:
Nevermind, my understanding of the problem was wrong.

In the pip package we don't install `_stbt/sqdiff.py` so we do actually get an ImportError. The OSError is only when it finds `_stbt/sqdiff.py` but doesn't find `_stbt/libstbt.so`. We want to keep this behaviour (raise OSError) so that we notice if we fail to build/install libstbt.so.

drothlis:
> Fix check-pythonpackage to use venv's pip

This was invalid too -- `pip3` is from inside the venv too.

PR #711, opened 2020-12-14
load_image: Replace `flags` parameter with `color_channels`

wmanley:
This is a bit more explicit about exactly what shape the output image will
be.  We now have the simple post-condition:

    out.shape[2] in color_channels

whereas previously if we passed an existing image as `filename` it would
be returned unmodified.  I've hit issues where I try to pre-load a mask image, but then it ends up with the wrong dimensions.

We also remove (deprecate) the dependency on `cv2` from our interface which
is poorly documented in the opencv documentation.  This means that
customers who use `load_image` no longer need to `import cv2`.


drothlis (approved):

tests/test_core.py
```diff
@@ -78,21 +78,55 @@ def test_load_image_with_unicode_filename():
...
+    b = numpy.zeros((720, 1280, 1), dtype=numpy.uint8)
+    c = numpy.zeros((720, 1280, 3), dtype=numpy.uint8)
+    d = numpy.zeros((720, 1280, 4), dtype=numpy.uint8)
+
+    channels = [(1,), (3,), (4,), (1, 3), (1, 4), (3, 4), (1, 3, 4)]
+    for x in [a, b, c, d]:
+        for y in channels:
+            img = stbt.load_image(x, color_channels=y)
+            assert img.shape[2] in y
+            assert isinstance(img, stbt.Image)
```
drothlis, line 99: Nice test. I have pushed a similar test but passing a filename instead of a numpy array.

PR #712, opened 2020-12-15
Pass regions and the inverse as mask= parameters

wmanley:
This enables:

```python
    SPINNER = stbt.Region(34, 433, right=44, bottom=444)
    stbt.wait_for_motion(mask=mask_out(SPINNER))
```

There is an interesting overlap here with #624.  Conceptually `mask_out(Region.ALL)` is `Region.NONE`.

**TODO:**

- [ ] Tests for `load_mask`
- [ ] Tests for `preload_mask`
- Tests for passing `region and `mask_out(Region)` to
    - [ ] `is_screen_black`
    - [ ] `wait_for_motion`
    - [ ] `press_and_wait`
- [ ] Update documentation to say that we can pass a region
- [ ] Document `mask_out`
- [ ] Change `repr(NotRegion(r))` to read `mask_out(r)`?

### DIFF
--- a/_stbt/black.py
+++ b/_stbt/black.py
@@ -11,7 +11,7 @@ https://github.com/stb-tester/stb-tester/blob/master/LICENSE for details).
 import cv2
 
 from .config import get_config
-from .imgutils import (crop, _frame_repr, load_image, pixel_bounding_box,
+from .imgutils import (crop, _frame_repr, load_mask, pixel_bounding_box,
                        _validate_region)
 from .logging import debug, ImageLogger
 from .types import Region
@@ -64,10 +64,9 @@ def is_screen_black(frame=None, mask=None, threshold=None, region=Region.ALL):
         from stbt_core import get_frame
         frame = get_frame()
 
-    if mask is not None:
-        mask = load_image(mask, color_channels=1)
-
     region = _validate_region(frame, region)
+    if mask is not None:
+        mask = load_mask(mask, shape=(region.height, region.width, 1))
 
     imglog = ImageLogger("is_screen_black", region=region, threshold=threshold)
     imglog.imwrite("source", frame)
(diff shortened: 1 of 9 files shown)


drothlis:
Can you compose these? I've long wished for a `Mask` class, something like this:

    mask = mask_in(Region.ALL).mask_out(clock_region).mask_out(some_other_region)

Under the covers I suppose it would keep a list of Regions/NotRegions, that it can render to a numpy array. Possibly with enough smarts to calculate a bounding box that the algorithms (e.g. stbt.match) can use to limit processing to a smaller region.

wmanley:
> Can you compose these

That would be the next step.  Once all the image processing functions convert their `mask` inputs into `ndarray`s using `load_image` we can then enhance `load_image` alone and the advantages will be felt by all the callers.

drothlis:
Great. So I gather that `NotRegion` isn't a public API, only `mask_out` is. Maybe `load_mask` should be public API, similar to `load_image`?

wmanley:
> Great. So I gather that `NotRegion` isn't a public API, only `mask_out` is.

Exactly

> Maybe `load_mask` should be public API, similar to `load_image`?

Maybe later. I'm anticipating a little bit of churn here.  For example we might want to return a region, ndarray tuple, or a list of regions to include for efficiency.

drothlis:
Rebased onto main.

drothlis:
Here's how `wait_for_motion` and `is_screen_black` behave since v28:         If you specify both `region` and `mask`, the mask must be the same size as the region (and the mask is relative to the top left corner of the region).

drothlis:
I want to play around with the API so I don't want to go overboard on the tests just yet. I'll merge this as the existing usage is still working (at least the tests are passing) and we can add more tests and document this officially in a future pull request (which I'm working on).

PR #715, opened 2021-01-11
core: Don't restart source pipeline on EOS

drothlis:
This behaviour was introduced to handle the VidiU's H.264 encoder restart (see commit 6b7ba0a "stbt.py: Handle VidiU H.264 encoder restart"). As far as we know, the only VidiU user (YouView) had some kind of repeater / HDMI doctor in between the STB and the VidiU, which provided a continuous signal even when the STB rebooted or changed mode, thus avoiding this behaviour from the VidiU.

I'm removing it because:

- It isn't used (as per the previous paragraph).
- It complicates the implementation.
- It obscures real problems with the source pipeline.
- It interacts poorly with the sink pipeline: the recorded video ends up corrupted, with wrong timestamps, or huge gaps, with lots of errors in the logs.

Now we ignore the EOS; `get_frame` will soon time out and raise `NoVideo`.

### DIFF
--- a/_stbt/core.py
+++ b/_stbt/core.py
@@ -38,7 +38,7 @@ from _stbt.types import NoVideo, Region
 from _stbt.utils import text_type, to_native_str, to_unicode
 
 gi.require_version("Gst", "1.0")
-from gi.repository import GLib, GObject, Gst  # pylint:disable=wrong-import-order
+from gi.repository import GLib, Gst  # pylint:disable=wrong-import-order
 
 Gst.init(None)
 
@@ -411,8 +411,8 @@ class SinkPipeline(object):
             Gst.debug_bin_to_dot_file_with_ts(
                 self.sink_pipeline, Gst.DebugGraphDetails.ALL, "ERROR")
         err, dbg = message.parse_error()
-        self._raise_in_user_thread(
-            RuntimeError("%s: %s\n%s\n" % (err, err.message, dbg)))
+        self._raise_in_user_thread(RuntimeError(
+            "Error from sink pipeline: %s: %s\n%s" % (err, err.message, dbg)))
 
     def __enter__(self):
         self.received_eos.clear()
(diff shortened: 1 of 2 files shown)


drothlis:
Note: This is different than "restart the source pipeline when we don't receive any frames *nor* EOS". We removed *that* optional behaviour 2 years ago in #596 (v31).

_stbt/core.py
```diff
@@ -719,24 +719,6 @@ def on_warning(self, _bus, message):
     def on_eos_from_source_pipeline(self, _bus, _message):
         if not self.tearing_down:
             warn("Got EOS from source pipeline")
```
wmanley, line 721: What do you think about adding:

```python
        self.tell_user_thread(RuntimeError("Received EOS from source pipeline"))
```

Here?

drothlis: Sounds like a good idea. It'd end up raising RuntimeError instead of NoVideo (which inherits from Exception) but I think that's fine. We could change the other NoVideo raised from `get_frame` (on underrun, i.e. no frames received) to also be a RuntimeError.

drothlis: Or we could change them both to NoVideo (i.e. make `get_frame` raise any exceptions from `tell_user_thread` as NoVideo). It turns out that `stbt_core.NoVideo` is part of our `stbt_core` public API.

wmanley: Yeah, that'll make it easier to track.

drothlis: Done:

- EOS will raise `NoVideo: EOS from source pipeline`
- Errors will raise `NoVideo: Error from source pipeline: ...`
- Underrun will raise `NoVideo: No frames received in 10s`
- Errors from the sink pipeline will raise `RuntimeError: Error from sink pipeline: ...`

_stbt/core.py
```diff
@@ -647,6 +647,8 @@ def get_frame(self, timeout_secs=10, since=None):
                         self.last_frame.time > since):
                     self.last_used_frame = self.last_frame
                     return self.last_frame
+                elif isinstance(self.last_frame, NoVideo):
+                    raise NoVideo(text_type(self.last_frame))
                 elif isinstance(self.last_frame, Exception):
                     raise RuntimeError(text_type(self.last_frame))
```
wmanley, line 653: I wonder why we don't `raise self.last_frame` here?

wmanley (approved):
> * Errors from the sink pipeline will raise RuntimeError: Error from sink pipeline: ...

I wonder if we don't want to error if there's something wrong with the sink pipeline, but let's not worry about that now, hopefully we'll ditch the sink pipeline before long.

PR #718, opened 2021-03-31
pip package: Make networkx an optional dependency

wmanley:
I've had a little difficulty with it on Windows.  It's not necessary
for pylint to work as it's no-longer exposed in our APIs.


setup.py
```diff
@@ -59,12 +59,12 @@
     extras_require={
         "ocr": ["lxml==4.2"],
         "debug": ["Jinja2==2.10.1"],
+        "keyboard": ["networkx==1.11"],
```
drothlis, line 62: How do you tell pip to install these extras?

PR #720, opened 2021-04-13
stbt.frames, get_frame docstrings: Add more information & examples

drothlis:
* Saving a frame to disk with `cv2.imwrite` was documented in our manual but repeating it here will make the information easier to find.
* The behaviour of `stbt.get_frame` re. blocking (or the lack thereof) wasn't documented anywhere.


drothlis:
Oof... more than 20 minutes for CircleCI to install the `python-package` dependencies.

PR #724, opened 2021-06-23
stbt.Frame: Add `width` and `height` properties

drothlis:
Similar to the recent commit 656a59f2d3a4 "stbt.Image: Add `width` and
`height` properties", but for `stbt.Frame`.

Recently I was checking for the width of a region obtained with
`right_of`. But this region has infinite width, so I needed to intersect
it with the size of the frame, but this isn't currently possible without
reaching into `shape` which is numpy-specific and has unintuitive
coordinate order. Now I can do something like
`r.right_of().replace(right=self._frame.right)`.

Note that the current documentation of `right_of` is incorrect, or at
least misleading: "Returns: A new region to the right of the current
region, extending **to the right edge of the frame** (or to the
specified width)" -- emphasis mine.

### DIFF
--- a/_stbt/imgutils.py
+++ b/_stbt/imgutils.py
@@ -62,6 +62,14 @@ class Frame(numpy.ndarray):
     def __str__(self):
         return repr(self)
 
+    @property
+    def width(self):
+        return self.shape[1]
+
+    @property
+    def height(self):
+        return self.shape[0]
+
 
 class Image(numpy.ndarray):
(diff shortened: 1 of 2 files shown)


wmanley:
I do:

    Region.intersect(r.right_of(), _image_region(self._frame))

drothlis:
`_image_region` isn't a public API though. This was user code in a test-pack.

PR #725, opened 2021-07-19
Move stbt_core.pylint_plugin to _stbt

wmanley:
So pylint can import it without having to import the whole of `stbt_core` (including cv2, etc.).


wmanley:
I also checked manually that `pylint --load-plugins=stbt_core.pylint_plugin` still works.

PR #733, opened 2021-12-02
_stbt.control: Make HTTP timeouts configurable

drothlis:
...for RokuHttpControl and RedRatHttpControl.

The motivation is that we occasionally see the RedRat Server take 4-5
seconds to respond, when there are multiple clients sending requests
to the same RedRat Server / the same RedRat-X.

### DIFF
--- a/_stbt/control.py
+++ b/_stbt/control.py
@@ -11,7 +11,7 @@ from distutils.spawn import find_executable
 import requests
 
 from . import irnetbox
-from .config import ConfigurationError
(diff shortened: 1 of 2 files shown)


drothlis:
As @wmanley pointed out this doesn't play nicely with stbt-control-relay which doesn't have access to the test-pack's stbt.conf. So there was a reason after all why `_stbt/control.py` never uses `get_config`. We'll need to supply this from the caller instead. Closing.

PR #736, opened 2021-12-12
Rework _stbt.android to focus on Android TV set-top-boxes

drothlis:
This is a breaking change but it won't affect any of our customers as `stbt.android` isn't a public API yet. Aiming for v33.

I originally wrote this module years ago for a proof-of-concept we did for mobile phones & tablets. No-one is actually using it for that as far as I know. The main changes are:

- Changed the order, names, and defaults of `AdbDevice` parameters.
- Made `AdbDevice.adb` more like [subprocess.run].
- Added `adb` free function (equivalent to creating an `AdbDevice` with default parameters from config, and calling its `adb` method).
- Added `logcat` decorator that saves logcat logs to "logcat.log" while the decorated function is running.

Compared to running `subprocess.run(["adb", ...])` directly, this module:

- Uses the configured adb_binary, adb_server, and device address automatically.
- Does `adb connect` to establish a Network ADB connection, if the device isn't already connected.
- Gives us an obvious place to set the address of the adb server on the Stb-tester HDMI Node.
- We can add other helpers such as the `logcat` decorator I added.

[subprocess.run]: https://docs.python.org/3.8/library/subprocess.html#subprocess.run


drothlis:
Rebased onto master now that #735 is merged.

drothlis:
Rebased onto main.

PR #737, opened 2021-12-18
stbt.Keyboard: Better error message when page.selection is None

drothlis:
If your Page Object is buggy, and it returns `is_visible=True` but `selection=None` we'd get an inscrutable TypeError from the depths of our implementation:

      File "_stbt/keyboard.py", line 707, in _keys_to_press
        for s, t in zip(path[:-1], path[1:]):
    TypeError: unhashable type: 'slice'

This is because we call `networkx.shortest_path(G, source, target)`. Normally it returns a list, but when `source` is None it returns a dict (with the shortest path for each node in the graph to `target`, keyed by the starting node).

### DIFF
--- a/_stbt/keyboard.py
+++ b/_stbt/keyboard.py
@@ -663,6 +663,8 @@ class Keyboard():
             self.G_ = _strip_shift_transitions(self.G)
 
         assert page, "%s page isn't visible" % type(page).__name__
+        assert page.selection in self.G_, \
+            "page.selection (%r) isn't in the keyboard" % (page.selection,)
         deadline = time.time() + self.navigate_timeout
         current = page.selection
         while current not in targets:
--- a/tests/test_keyboard.py
+++ b/tests/test_keyboard.py
@@ -234,18 +234,24 @@ def slow_dut():
         yield dut
 
 
+class _NotSpecified():  # sentinel value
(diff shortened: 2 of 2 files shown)


_stbt/keyboard.py
```diff
@@ -663,6 +663,8 @@ def navigate_to(self, target, page, verify_every_keypress=False, retries=2):
             self.G_ = _strip_shift_transitions(self.G)
 
         assert page, "%s page isn't visible" % type(page).__name__
+        assert page.selection in self.G_, \
+            "page.selection (%r) isn't in the keyboard" % (page.selection,)
```
drothlis, line 667: I *think* this is the right exception. This will be a test **failure** (similar to a non-visible `page` in the previous line), vs. the ValueError (a test **error**) if `target` isn't in the keyboard:

https://github.com/stb-tester/stb-tester/blob/b1b9d4b8c991c24cd461561e96fa0354d86cc432/_stbt/keyboard.py#L657-L659

My thinking is that `target` (or `text`) is usually specified explicitly as a literal string in the test script, whereas a bad `page.selection` could be a bug in the device-under-test.

drothlis:
Force-pushed to remove some debug logging I had committed by accident.

PR #738, opened 2022-01-13
Don't build our own libxxhash

drothlis:
It's available in Ubuntu 18.04 onward (package libxxhash0).


wmanley:
Great. Ditching the submodule makes this easier to build and integrate.

PR #739, opened 2022-02-01
Fix stbt-debug rendering in Jupyter Notebooks

drothlis:
With Jupyter Notebook 6.4.6 running on Ubuntu 20.04, the IFrame with the
stbt-debug index.html doesn't show any of the images. Looking in the
network tab of my browser's developer tools, I can see that the requests
for the images are being redirected to
"login?next=<path to the actual image.png>".

If I right-click the broken image and open it in a new tab, it works;
but it still doesn't display the image in the HTML.

This commit renders correctly in Jupyter but obviously it's a hack
because it breaks the normal index.html when viewed directly.

TODO: Fix it properly.

### DIFF
--- a/_stbt/logging.py
+++ b/_stbt/logging.py
@@ -190,20 +190,19 @@ class ImageLogger():
 
         index_html = os.path.join(self.outdir, "index.html")
         with open(index_html, "w") as f:
-            f.write(jinja2.Template(_INDEX_HTML_HEADER)
-                    .render(frame_number=self.frame_number,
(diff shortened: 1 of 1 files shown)


drothlis:
> With Jupyter Notebook 6.4.6 running on Ubuntu 20.04

I don't know what I was testing because today, 4 months later, the version of jupyter-notebook in Ubuntu 20.04's apt repository is 6.0.3-2.

Must've been some newer version in a virtualenv.

PR #745, opened 2022-05-27
Add Color type

drothlis:
So far we have been specifying colors as (blue, green, red) tuples. This supports that same format, but also supports the more user-friendly and familiar "#rrggbb" format in a string.

TODO:

- [x] Decide what to do about the type annotations: ~Remove them or~ accept keyword args with those names & types in the actual implementation.

### DIFF
--- a/_stbt/imgutils.py
+++ b/_stbt/imgutils.py
@@ -186,21 +186,36 @@ class Color:
     @overload
     def __init__(self, bgr: Tuple[int, int, int]) -> None:
         ...
-    def __init__(self, *args):
-        self.array: numpy.ndarray = None  # BGR with shape (1, 1, 3) or BGRA
-        if len(args) == 1:
-            if isinstance(args[0], Color):
-                self.array = args[0].array
-            elif isinstance(args[0], str):
-                self.array = Color._from_string(args[0])
-            elif (isinstance(args[0], (list, tuple, numpy.ndarray)) and
-                    len(args[0]) in (3, 4)):
-                self.array = Color._from_sequence(*args[0])
+    def __init__(self, *args,
+                 hexstring: str = None,
+                 blue: int = None, green: int = None, red: int = None,
+                 bgr: Tuple[int, int, int] = None):
+
+        self.array: numpy.ndarray  # BGR with shape (1, 1, 3) or BGRA
+
+        if len(args) == 1 and isinstance(args[0], Color):
+            self.array = args[0].array
+
+        elif len(args) == 1 and isinstance(args[0], str):
+            self.array = Color._from_string(args[0])
+        elif not args and hexstring is not None:
+            self.array = Color._from_string(hexstring)
+
+        elif (len(args) == 1 and
+                isinstance(args[0], (list, tuple, numpy.ndarray)) and
+                len(args[0]) in (3, 4)):
+            self.array = Color._from_sequence(*args[0])
         elif len(args) in (3, 4):
             self.array = Color._from_sequence(*args)  # pylint:disable=no-value-for-parameter
-        if self.array is None:
+        elif not args and bgr is not None:
+            self.array = Color._from_sequence(*bgr)
+        elif not args and all(x is not None for x in (blue, green, red)):
+            self.array = Color._from_sequence(blue, green, red)
+
+        else:
             raise TypeError("Color: __init__() expected a Color, '#rrggbb' "
                             "string, or 3 integers in Blue-Green-Red order. ")
+
         self.hexstring = (
             "#{0:02x}{1:02x}{2:02x}{3}".format(
                 self.array[0][0][2], self.array[0][0][1], self.array[0][0][0],
--- a/_stbt/imgutils.py
+++ b/_stbt/imgutils.py
@@ -2,6 +2,7 @@
 
 import inspect
 import os
+import re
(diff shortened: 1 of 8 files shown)


_stbt/imgutils.py
```diff
@@ -157,6 +158,97 @@ def _frame_repr(frame):
...
+                self.array = Color._from_string(args[0])
+            elif isinstance(args[0], (list, tuple)) and len(args[0]) in (3, 4):
+                self.array = Color._from_sequence(*args[0])
+        elif len(args) in (3, 4):
+            self.array = Color._from_sequence(*args)  # pylint:disable=no-value-for-parameter
+        if self.array is None:
+            raise TypeError("Color: __init__() expected a Color, '#rrggbb' "
+                            "string, or 3 integers in Blue-Green-Red order. ")
+        self.hexstring = (
+            "#{0:02x}{1:02x}{2:02x}{3}".format(
```
wmanley, line 206: It might be better to make this a method - someone could update `self.array`, but this would still show the old value.

```suggestion

        def hexstring(self):
            return "#{0:02x}{1:02x}{2:02x}{3}".format(
```

_stbt/imgutils.py
```diff
@@ -157,6 +158,97 @@ def _frame_repr(frame):
...
+                raise ValueError(f"Color: __init__ expected a value between 0 "
+                                 f"and 255: Got {x}")
+            out.append(x)
+        return out
+
+    def __repr__(self):
+        return f"Color('{self.hexstring}')"
+
+    def __eq__(self, other):
+        return isinstance(other, Color) and self.hexstring == other.hexstring
```
wmanley, line 261: I wonder if `#012345ff` should == `#012345`.  Not important for now.

_stbt/imgutils.py
```diff
@@ -157,6 +158,97 @@ def _frame_repr(frame):
...
+    >>> Color('#f77f00')
+    Color('#f77f00')
+
+    Or from Blue, Green, Red values in the range 0-255:
+
+    >>> Color(0, 127, 247)
+    Color('#f77f00')
+
+    Note: When you specify the colors in this way, the BGR order is the
+    opposite of the HTML-style RGB order. This is for backwards compatibility
```
wmanley, line 176:
```suggestion
    opposite of the HTML-style RGB order. This is for compatibility
```

_stbt/imgutils.py
```diff
@@ -157,6 +158,97 @@ def _frame_repr(frame):
...
+    with the way OpenCV stores colors.
```
wmanley, line 177: Need to document the array member? Or make it private.

_stbt/ocr.py
```diff
@@ -521,25 +516,32 @@ def _tesseract(frame, region, mode, lang, _config, user_patterns, user_words,
...
                                  user_patterns, user_words, upsample,
                                  engine, char_whitelist, imglog,
                                  tesseract_version, use_cache=True)
 
 
+def bgr_diff(frame, color, threshold, imglog):
+    # Calculate distance of each pixel from `text_color`, then discard
+    # everything further than `text_color_threshold` distance away.
+    diff = numpy.subtract(frame, Color(color).array, dtype=numpy.int32)
+    frame = numpy.sqrt((diff[:, :, 0] ** 2 +
```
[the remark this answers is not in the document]
drothlis: Agreed, but this code isn't added by this PR, just moved around.

_stbt/imgutils.py
```diff
@@ -157,6 +159,111 @@ def _frame_repr(frame):
...
+    @overload
+    def __init__(self, blue: int, green: int, red: int) -> None:
+        ...
+    @overload
+    def __init__(self, bgr_tuple: Tuple[int, int, int]) -> None:
+        ...
+    @overload
+    def __init__(self, bgra_tuple: Tuple[int, int, int, int]) -> None:
+        ...
+    def __init__(self, *args):
```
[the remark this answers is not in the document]
drothlis: > Does this complete nicely in VSCode?

See comment below: https://github.com/stb-tester/stb-tester/pull/745#issuecomment-1141806927

> We'd need to keep the conversion from `'#222222'` and `(b, g, r)` because we'll be applying these automatically. Probably not worth changing.

Exactly. The whole point is to have a single place to convert from any accepted input to our internal representation.

_stbt/imgutils.py
```diff
@@ -157,6 +159,111 @@ def _frame_repr(frame):
...
+        self.array: numpy.ndarray = None  # BGR with shape (1, 1, 3) or BGRA
+        if len(args) == 1:
+            if isinstance(args[0], Color):
```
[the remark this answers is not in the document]
drothlis: Good idea. Done in 93afec79.

_stbt/imgutils.py
drothlis: Done.

_stbt/imgutils.py
drothlis: I think of these as immutable, but I don't know how to enforce that. Really all the members are private but I didn't want to go adding underscores everywhere. I suppose I'll just do it. I think of this as a private class that we need to expose for the sake of documentation / Sphinx — in practice users don't ever need to construct a `Color` explicitly.

PR #753, opened 2022-06-14
match: Show full sized frame in html debug

### LINKED ISSUE #705: stbt.match html debug: MatchResult.region is drawn in wrong place if region given to match
It shows this:

image

whereas `level0-source_with_match.png` correctly shows this:

image

drothlis:
...instead of showing only the part that's cropped to `region`.

This fixes the position of the match region drawn onto the frame, if you specified `region`.

Also, use our existing `ImageLogger._draw_annotated_image` infrastructure for drawing these regions. This will also add a purple rectangle showing the input `region`, and it uses CSS to draw the rectangles instead of creating another PNG.

Fixes #705.

Manual testing:

- [x] stbt.match with & without a region.
- [x] stbt.match_all with & without a region.

### DIFF
--- a/_stbt/match.py
+++ b/_stbt/match.py
@@ -376,7 +376,8 @@ def _match_all(image, frame, match_parameters, region):
     imglog = ImageLogger(
         "match", match_parameters=match_parameters,
         template_name=t.filename or "<Image>",
-        input_region=input_region)
+        region=input_region)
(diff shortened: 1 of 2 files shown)


drothlis:
ImageLogger needs a total overhaul but this fixes the immediate bug (#705).

drothlis:
image

PR #759, opened 2022-06-23
Add type stbt.Direction

drothlis:
For APIs such as segmentation that process the image in a particular direction (horizontal or vertical).

### DIFF
--- a/_stbt/types.py
+++ b/_stbt/types.py
@@ -2,6 +2,7 @@
 # Don't import anything not in the Python standard library from this file
 
 from collections import namedtuple
+from enum import Enum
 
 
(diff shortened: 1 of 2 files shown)


_stbt/types.py
```diff
@@ -13,6 +14,19 @@ class Size(namedtuple('Size', 'width height')):
     """Size of a rectangle with ``width`` and ``height``."""
 
 
+class Direction(Enum):
+
+    #: Process the image from left to right
+    HORIZONTAL = "horizontal"
```
wmanley, line 20: Do these need values?

drothlis: How would you define an enum without them?

PR #760, opened 2022-06-23
wait_until: Add type annotations

drothlis:
TODO:

- [x] Test autocompletion in VS Code.

### DIFF
--- a/_stbt/wait.py
+++ b/_stbt/wait.py
@@ -1,13 +1,20 @@
-"""Copyright 2015-2019 Stb-tester.com Ltd."""
+"""Copyright © 2015-2022 Stb-tester.com Ltd."""
 
 import functools
 import inspect
+from typing import Callable, TypeVar
 
 from .logging import debug
 
 
-def wait_until(callable_, timeout_secs=10, interval_secs=0, predicate=None,
-               stable_secs=0):
+T = TypeVar("T")
+
+
+def wait_until(callable_: Callable[[], T],
+               timeout_secs: float = 10,
+               interval_secs: float = 0,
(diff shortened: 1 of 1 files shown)


drothlis:
It works! Before:

vscode before

After:

vscode after

_stbt/wait.py
```diff
@@ -1,13 +1,20 @@
...
 
 
-def wait_until(callable_, timeout_secs=10, interval_secs=0, predicate=None,
-               stable_secs=0):
+T = TypeVar("T")
+
+
+def wait_until(callable_: Callable[[], T],
+               timeout_secs: int = 10,
+               interval_secs: int = 0,
```
wmanley, line 15: These should be `float`

```suggestion
               timeout_secs: float = 10,
               interval_secs: float = 0,
```

_stbt/wait.py
```diff
@@ -1,13 +1,20 @@
...
+               predicate: Callable[[T], bool] = None,
+               stable_secs: int = 0) -> T | None:
```
wmanley, line 17:
```suggestion
               stable_secs: float = 0) -> T | None:
```

wmanley (changes requested):

wmanley:
Very helpful.  I had a branch where I'd annotated all of our types at some point, I'll have to find it.

_stbt/wait.py
drothlis: Done.

_stbt/wait.py
drothlis: Done.

wmanley (approved):

PR #761, opened 2022-06-23
FrameObject.refresh: Use self.__class__ instead of type(self)

drothlis:
This helps pylint to infer the correct type.


drothlis:
This fixes several new false positives I've seen with Ubuntu 22.04's version of pylint.

PR #765, opened 2022-07-03
FrameObject repr: Only print values of properties already calculated

drothlis:
To avoid triggering expensive calculations.

### DIFF
--- a/_stbt/frameobject.py
+++ b/_stbt/frameobject.py
@@ -55,6 +55,13 @@ def _memoize_property_fn(fn):
     return inner
 
 
+def _convert_is_visible_to_bool(fn):
+    @functools.wraps(fn)
+    def inner(self):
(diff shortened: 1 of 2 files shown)


drothlis:
@wmanley do you remember what was the use-case we had that required this?

drothlis:
> @wmanley do you remember what was the use-case we had that required this?

I remember now: It's so that we could add debug that prints the return value from `wait_until`.

wmanley:
Good stuff.  It would probably be worth revisiting some more of these original `FrameObject` decisions.  The repr used to be very important as we were automatically generating doctests.  Since the Object Repo has taken over that role there are probably some other things we could revisit - such as including the `frame` in the repr output too.  This would be nice in the log viewer as users could then click on the timestamp to see the video.

PR #771, opened 2022-07-09
Print deprecation warning for deprecated "region" parameter

drothlis:
Note: stacklevel=2 shows the line of code that called is_screen_black (etc) instead of the line of code in is_screen_black itself that is calling `warnings.warn`.

stacklevel=1:

    _stbt/black.py:72: DeprecationWarning: stbt.is_screen_black: The 'region' parameter is deprecated; pass your Region to 'mask' instead
      warnings.warn(

stacklevel=2:

    test.py:4: DeprecationWarning: stbt.is_screen_black: The 'region' parameter is deprecated; pass your Region to 'mask' instead
      print(stbt.is_screen_black(region=stbt.Region(0,0,100,100)))

I also changed the warnings filter so that these messages actually get shown. The "default" action prints the warning once per (file, line number) — i.e. once per call-site.

### DIFF
--- a/_stbt/black.py
+++ b/_stbt/black.py
@@ -8,6 +8,7 @@ License: LGPL v2.1 or (at your option) any later version (see
 https://github.com/stb-tester/stb-tester/blob/master/LICENSE for details).
 """
 
+import warnings
 from typing import Optional
 
 import cv2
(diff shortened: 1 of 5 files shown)


drothlis:
I tested that with `stbt.wait_until(lambda: stbt.is_screen_black(region=r), timeout_secs=0.1)` the deprecation warning is shown only once for that line, not once per frame.

PR #778, opened 2022-09-28
LIRC control: Increase socket timeout

drothlis:
The first keypress can take a while if the lircd (or stbt-control-relay) process is socket-activated and hasn't started yet.


drothlis:
In particular stbt-control-relay with HdmiCecControl can take several seconds to start up. With libcec3 it was only taking a second or two, but with libcec4 the call to `RescanActiveDevices` is taking up to 6 seconds.

PR #781, opened 2022-11-01
Convert stbt_run.py to be a pytest plugin 

drothlis:
This allows running your test-scripts with pytest, to get pytest's features (in particular [assertion introspection] and [fixtures]).

[assertion introspection]: https://docs.pytest.org/en/6.2.x/assert.html
[fixtures]: https://docs.pytest.org/en/7.1.x/explanation/fixtures.html#about-fixtures

### DIFF
--- a/_stbt/wait.py
+++ b/_stbt/wait.py
@@ -1,6 +1,5 @@
 """Copyright © 2015-2022 Stb-tester.com Ltd."""
 
-import functools
 import inspect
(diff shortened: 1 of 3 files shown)


drothlis:
> I guess pytest uses `functools.partial` in the source-code rewriting it does to support assertion introspection.

Or it manipulates the line numbers in the new AST it writes.

PR #783, opened 2022-11-10
load_image: Normalize alpha channel when loading numpy arrays

drothlis:
Passing a numpy array now behaves the same as passing a PNG.


drothlis:
@wmanley please review. This affected a customer who was downloading images from a backend server and passing them directly to `stbt.match` without writing them to disk first.

wmanley (approved):

PR #785, opened 2022-11-22
Make BGRDiff the default for press_and_wait / detect_motion / wait_for_motion

drothlis:
The previous default (GrayscaleDiff) doesn't detect changes that are different in colour but of a similar overall intensity, such as the XFinity keyboard shown in `tests/test_diff.py::test_bgrdiff`.

### DIFF
--- a/_stbt/motion.py
+++ b/_stbt/motion.py
@@ -5,7 +5,7 @@ from collections import deque
 from typing import Iterator, Optional
 
 from .config import ConfigurationError, get_config
-from .diff import FrameDiffer, GrayscaleDiff, MotionResult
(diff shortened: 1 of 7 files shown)


wmanley (approved):

wmanley:
Note: the docs for `press_and_wait`, `detect_motion` and `wait_for_motion` need to be updated with the new meaning for `threshold`, and probably a "new in v34" note about it.

drothlis:
> Note: the docs for `press_and_wait`, `detect_motion` and `wait_for_motion` need to be updated with the new meaning for `threshold`, and probably a "new in v34" note about it.

#800

PR #790, opened 2023-04-24
Mask.to_array: Log a warning if binary op results in an empty region

drothlis:
I found it useful for debugging issues with my coordinates when porting a test-script from 720p reference images to 1080p.


_stbt/mask.py
```diff
@@ -245,6 +246,8 @@ def _to_array_and_bounding_box_cached(
         bounding_box = Region.intersect(region, mask._region)
     else:
         bounding_box = pixel_bounding_box(array)
+        if bounding_box is None:
+            logger.warning("%r is an empty Region", mask)
```
wmanley, line 250: This is a side-effect inside a `lru_cache`ed function, but that's probably a good thing - this message won't be spammed too much.  On the downside the warning may be far in the logs from the error the user is debugging.  On balance I think this will be a good thing, and thanks to the caching we could possibly even include a backtrace here.

wmanley (approved):

PR #791, opened 2023-04-24
Mask: Add `from_alpha_channel` constructor

drothlis:
Useful when you have a reference image with some transparent parts, but you want to use that same mask for an operation other than `match`.

### DIFF
--- a/_stbt/mask.py
+++ b/_stbt/mask.py
@@ -8,7 +8,7 @@ import cv2
 import numpy
 
 from .imgutils import (
-    _convert_color, crop, find_file, Image, ImageT, load_image,
+    _convert_color, crop, find_file, Image, ImageT, _image_region, load_image,
     pixel_bounding_box, _relative_filename)
 from .logging import logger
 from .types import Region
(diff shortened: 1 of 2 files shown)


_stbt/mask.py
```diff
@@ -102,6 +102,24 @@ def __init__(self, m: MaskTypes, *, invert: bool = False) -> None:
...
+          An image with an alpha (transparency) channel. This can be the
+          filename of a png file on disk, or an image previously loaded with
+          `stbt.load_image`.
+
+          Filenames should be relative paths. See `stbt.load_image` for the
+          path lookup algorithm.
+        """
+        image = load_image(image)
+        if image.shape[2] != 4:
+            raise ValueError(f"{image} doesn't have an alpha channel")
```
wmanley, line 120: Conceptually speaking we treat an image with `shape[2] < 4` as having an alpha channel of all `255`s.  So I think this should be:

```suggestion
            return Mask(_image_region(image))
```

That way calling code may not need to check `shape[2]` itself at all.  Interestingly then `Mask.from_alpha_channel` becomes a generalisation of `_image_region`.

wmanley (changes requested):
This will be handy :+1:

_stbt/mask.py
drothlis: Done.

wmanley (approved):
>This is the size of the image instead of `Region.ALL` because we don't know how it'll be used, so we don't want to lose that information (potentially the caller could translate it within a larger image?).

I see it as for consistency: preserving the invariant that an image with an alpha channel of all `0xFF` behaves exactly the same as an image without an alpha channel.

PR #792, opened 2023-04-25
match: Don't ignore reference image's alpha channel if frame is single-channel

### LINKED ISSUE #789: match on single channel images ignores template alpha
In `match` if frame is a single channel array (like from `color_diff`) then alpha in the template is ignored:

https://github.com/stb-tester/stb-tester/blob/71b907aa4a3e5208843cc63dc137d818818ce83c/_stbt/match.py#L354

So this behaves incorrectly:

```py
diff = color_diff(self._frame, "#ffffff")
stbt.match("template-with-alpha.png", frame=diff)
```

It should behave as if `diff` had three channels.

Workaround:

```py
diff = color_diff(self._frame, "#ffffff")
diff = load_image(diff, color_channels=3)
stbt.match("template-with-alpha.png", frame=diff)
```

drothlis:
Fixes #789: If frame is a single channel array (like from `color_diff`) then alpha in the template is ignored.

TODO:

- [x] Testcase.
- [ ] Fix it.


wmanley (approved):
>This is the size of the image instead of `Region.ALL` because we don't know how it'll be used, so we don't want to lose that information (potentially the caller could translate it within a larger image?).

I see it as for consistency: preserving the invariant that an image with an alpha channel of all `0xFF` behaves exactly the same as an image without an alpha channel.

wmanley (commented):
Ignore this review, I'm trying to undo the previous review.

wmanley:
I'd be tempted to change the behaviour of `load_image` - so that it raises rather than throwing away information - if an image contains colour or an alpha channel then it's probably a bug to throw that information away.  This would have helped diagnose #789.

wmanley:
@drothlis: Please review.

PR #797, opened 2023-05-05
API: press_and_wait/wait_for_transition_to_end: Add public frames parameter

wmanley:
Consistent with `wait_for_motion`.  This will allow performing some transformation on the video before it's processed by `wait_for_motion`.

### DIFF
--- a/_stbt/transition.py
+++ b/_stbt/transition.py
@@ -130,7 +130,7 @@ def press_and_wait(
             DeprecationWarning, stacklevel=2)
         mask = region
 
-    t = _Transition(mask, timeout_secs, stable_secs, min_size, _dut)
+    t = _Transition(mask, timeout_secs, stable_secs, min_size, _dut.frames())
     press_result = _dut.press(key)
     debug("transition: %.3f: Pressed %s" % (press_result.end_time, key))
     result = t.wait(press_result)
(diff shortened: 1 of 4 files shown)


_stbt/transition.py
```diff
@@ -30,7 +37,8 @@ def press_and_wait(
     region: Region = Region.ALL,
     timeout_secs: float = 10,
     stable_secs: float = 1,
-    min_size: SizeT = None,
+    min_size: "SizeT|None" = None,
+    frames: "typing.Iterator[core.Frame]|None" = None,
```
drothlis: I suppose you had to add `|None` to satisfy vscode / pylance?

Interesting that Sphinx automatically works out that it's Optional, even if you don't specify it. See https://stb-tester.com/manual/python-api#stbt.detect_motion vs. https://github.com/stb-tester/stb-tester/blob/1894d831b93ab28a77f527947d2b041d745e6fbd/_stbt/motion.py#L15-L21

_stbt/transition.py
```diff
@@ -30,7 +37,8 @@ def press_and_wait(
...
     _dut=None,
```
drothlis, line 35: I see, `press_and_wait` still needs `_dut` for `_dut.press()` (for the unit tests). At least we removed `_dut` from `wait_for_transition_to_end`. :+1:

_stbt/transition.py
```diff
@@ -22,11 +22,12 @@
 from .imgutils import FrameT
 from .logging import ddebug, debug, draw_on
 from .mask import MaskTypes
-from .types import KeyT, Region, SizeT
+from .types import KeyT, Region
 
 if typing.TYPE_CHECKING:
     # pylint:disable=unused-import
     from . import core
+    from .types import SizeT
```
drothlis: We already use `from __future__ import annotations` (see 905e2a98) and we already import `.types` so we don't need all this ceremony. I have pushed a commit.

drothlis (approved):

PR #799, opened 2023-05-17
Refactor: Replace subclassing FrameDiffer with subclassing Differ

wmanley:
Instead of subclassing `FrameDiffer` we now subclass `Differ` and pass an
instance of that subclass to `FrameDiffer`.  So before you'd write:

    press_and_wait.differ = BGRDiff

and now it's

    press_and_wait.differ = BGRDiff()

This means that we can customise differ parameters by passing arguments to
the constructor like:

    press_and_wait.differ = BGRDiff(threshold=60)

To support this differs are now immutable.  The `previous_frame` state that
used to live in `BGRDiff` and `GreyscaleDiff` is now implemented in
`FrameDiffer` which is now a concrete class.

We still need to support passing differ parameters directly to
`press_and_wait`, etc. like `press_and_wait(min_size=...)`, so there's a
`.replace` argument which returns a new differ, but with those parameters
set.

`Differ` is designed so we can efficiently implement `GrayscaleDiff`
without doing the colour -> grayscale conversion on the same image twice -
just as we could before.  To avoid internal state the interface makes
preprocessing explicit.  This makes the classes harder to use directly, but
that's ok because users won't be using them directly.

Currently we don't override `preprocess_mask` anywhere, but it seems to
fall out naturally from the design.

**TODO**

* [x] Rebase now that #785 is merged
* [x] Make tests pass
* ~~Pass ImageLogger into methods?~~

**Future work**

* Allow passing a `Differ` as `second_pass_method` to match.
* Global stbt.default_differ?
* Pass `differ` argument into `press_and_wait`, etc.

### DIFF
--- a/_stbt/diff.py
+++ b/_stbt/diff.py
@@ -5,8 +5,7 @@ import typing
 import cv2
 import numpy
 
-from .imgutils import (
-    Frame, FrameT, crop, _frame_repr, _image_region, pixel_bounding_box)
+from .imgutils import Frame, FrameT, crop, _frame_repr, pixel_bounding_box
 from .logging import ddebug, ImageLogger
 from .mask import load_mask, MaskTypes
 from .types import Region, SizeT
@@ -211,36 +210,6 @@ class BGRDiff(Differ):
         ddebug(str(result))
         imglog.html(BGRDIFF_HTML, result=result)
         return result
-
-
-class FrameDiffer():
-    """Interface for different algorithms for diffing frames in a sequence.
-
-    Say you have a sequence of frames A, B, C. Typically you will compare frame
-    A against B, and then frame B against C. This is a class (not a function)
-    so that you can remember work you've done on frame B, so that you don't
-    repeat that work when you need to compare against frame C.
-    """
-    def __init__(self, differ: Differ, initial_frame: FrameT,
-                 mask: MaskTypes = Region.ALL):
-        self.differ: Differ = differ
-        self.mask_tuple = differ.preprocess_mask(
-            mask, _image_region(initial_frame))
-        self.prev_frame = differ.preprocess(initial_frame, self.mask_tuple)
-
-    def diff(self, frame: FrameT) -> MotionResult:
-        new_frame = self.differ.preprocess(frame, self.mask_tuple)
-        motion = self.differ.diff(self.prev_frame, new_frame, self.mask_tuple)
-
-        if motion:
-            # Only update the comparison frame if it's different to the previous
-            # one.  This makes `detect_motion` more sensitive to slow motion
-            # because the differences between frames 1 and 2 might be small and
-            # the differences between frames 2 and 3 might be small but we'd see
-            # the difference by looking between 1 and 3.
-            self.prev_frame = new_frame
-
-        return motion
 
 
 BGRDIFF_HTML = """\
--- a/_stbt/diff.py
+++ b/_stbt/diff.py
@@ -53,19 +53,27 @@ class UNSET:
 
 
 class Differ:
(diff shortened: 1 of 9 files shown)


drothlis (commented):
Haven't finished the review, just jotting down some thoughts.

So if you want to implement a new diffing algorithm you subclass Differ and it needs to accept `min_size`, `threshold`, and `erode`. It can have new parameters too, but it must accept those.

It feels a bit weird to have `min_size` be part of each diff algorithm, but I suppose it's better to repeat that logic in each differ, than to repeat it in each caller (press_and_wait, etc). (Note: This isn't new, just an observation about the existing code.)

FrameDiffer should be removed from the public API, and Differ added to the public API (so that you can create your own subclasses — that's the only reason FrameDiffer was there).

There might be a better name for FrameDiffer. Its responsibilities:

- Remember the work done on seen frames.
- The logic for when we update the "reference" frame, which we added to
  help it detect slow gradual motion.

It's really the core of the detect_motion algorithm, but in an API that's more convenient to use from press_and_wait.

I'm not a fan of `preprocess_mask` — why do we need a wrapper around `load_mask` / `Mask.to_array`, do you have a concrete plan / use case for it?

Is it that some future differ might want a 3-channel mask? Both of the current differs want a single-channel mask, which is what the default implementation of `Differ.preprocess_mask` gives.

I don't like passing `mask_tuple` —I don't like passing ad-hoc tuples around, really— when we could pass it as `mask_pixels` and `region`.

Note that `Mask.to_array` is memoized, so we could call it from `BGRDiffer.{preprocess,diff}` directly? No, that feels wrong.

_stbt/diff.py
```diff
@@ -49,20 +49,77 @@ def __repr__(self) -> str:
...
+    compared.  It's used by `wait_for_motion` and `press_and_wait`.
+
+    `Differ` is designed so we can efficiently implement `GrayscaleDiff`
+    without doing the colour -> grayscale conversion on the same image twice -
+    just as we could before.  To avoid internal state the interface makes
+    preprocessing explicit.  This makes the classes hard to use, but that's ok -
+    it's a low-level interface intended mostly for internal use.
+
+    The only public interface is the subclass constructors.  The methods are not
+    intended for use by external code.
```
drothlis, line 68: Are we going to support users implementing their own Differ subclasses? If so, they'll have to implement `preprocess` and `diff`, so those are technically public interfaces too.

drothlis: Probably not — the interface is too complex. The real reason for making the base class public is for type annotations.

drothlis:
@wmanley I pushed a few commits implementing my earlier comments & tidying up the docstrings for public APIs. Let me know what you think.

wmanley:
I dropped your commit where `PreProcessedMask` was removed and moved the implementations of `preprocess_mask` into the concrete classes, hopefully to make it clearer.  I've also made the types "protected" with a leading underscore, hopefully to make the design clearer.  What do you think now?

drothlis (approved):
I'm a bit concerned about the complexity here: We have the `preprocess_mask` for the sake of mirroring `preprocess_frame`, but no specific use-case for it; a hypothetical way of typing them that we don't actually use (we aren't specifying the return types in the subclasses) and that doesn't type-check (and maybe isn't possible to actually express with Python's typing?); and a parameter called `mask` (in the subclasses) that isn't a `Mask`. I suppose it's not so bad, hopefully I'll still understand it next time I come to work in this area. :-)

drothlis:
On a more positive note, I'm excited about the possibility of passing custom differs into various APIs, including the `match` second pass!

PR #801, opened 2023-06-12
BGRDiff: Implement inner loop in C

wmanley:
This removes a bunch of large temporary arrays from the BGRDiff hot path. From looking at godbolt it looks like it vectorises well too.

This code is basically untested.  More testing required.  Raising now for profiling.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,7 +6,6 @@ exclude _stbt/gst_hacks.py
 exclude _stbt/gst_utils.py
 exclude _stbt/irnetbox.py
 exclude _stbt/power.py
-exclude _stbt/sqdiff.py
 exclude _stbt/stbt_run.py
 exclude _stbt/x11.py
 exclude _stbt/xxhash.py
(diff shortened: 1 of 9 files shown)


drothlis:
On the Stb-tester Node 1st gen hardware:

image

In reality it'll be faster than 40ms, that's just the video-capture framerate of my test (i.e. the rate that `stbt.frames()` is feeding frames to `press_and_wait`).

<details><summary>Methodology:</summary>
 
I ran a test that just does `stbt.press_and_wait("KEY_LEFT")` and then I took the timestamps from the debug output with the following script, and pasted the output into a spreadsheet:

```
❯ cat tools/compare_press_and_wait_performance.py 
#!/usr/bin/python3

import re
import sys


def main():
    prev = None
    for line in open(sys.argv[1]):
        if m := re.search(r"DEBUG:stbt:transition: ([0-9.]+)", line):
            ts = float(m.group(1))
            if prev is not None:
                print("%.3f" % (ts - prev))
            prev = ts


if __name__ == "__main__":
    main()
```

</details>

PR #804, opened 2023-07-25
Type fixes

drothlis:


stbt_core/__init__.py
```diff
@@ -159,12 +159,14 @@
 from _stbt.imgutils import ImageT
 from _stbt.types import KeyT, RegionT
 
+TEST_PACK_ROOT: "str|None" = None
```
wmanley, line 162:
```suggestion
TEST_PACK_ROOT: str
```

drothlis: But then I'd still have to do `getattr(stbt_core, "TEST_PACK_ROOT", None)` everywhere.

PR #808, opened 2023-08-31
HdmiCecControl: Add "KEY_GUIDE" alias for "KEY_EPG"

drothlis:
We tend to use "KEY_GUIDE" as it's clearer than a cryptic acronym.


wmanley:
Hmm, `KEY_GUIDE` is not in https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h but `KEY_EPG` is. Oh well, we're using `KEY_GUIDE` pretty consistently now and it's not worth changing.

PR #809, opened 2023-09-27
Keyboard.add_grid: Merge identical adjacent keys

wmanley:
For example: this keyboard has a 3 square wide space bar and a two square wide OK button on the bottom row:

    KEYBOARD = stbt.Keyboard()
    KEYBOARD.add_grid(
        stbt.Grid(
            stbt.Region(377, 476, 527, 228),
            data=[
                "qwertyuiop",
                "asdfghjkl,",
                ["CAPS"] + list("zxcvbnm.") + ["BACKSPACE"],
                ["?123"] + list("◀▶   -_") + ["OK"] * 2,
            ]
        ),
        mode="lower", merge=True
    )

Without this change you'd have to separately add space and ok as separate keys and explicitly list the transitions up, left and right.  This is much less verbose and more readable.  See the changes to the tests to see.

### DIFF
--- a/_stbt/keyboard.py
+++ b/_stbt/keyboard.py
@@ -318,7 +318,7 @@ class Keyboard():
                 _join_with_commas([str(x) for x in sorted(keys)],
                                   last_one=" or ")))
 
-    def _find_keys(self, query, mode=None):
+    def _find_keys(self, query, mode=None) -> "list[Key]":
         """Like the public `find_keys`, but takes a "query" (see _find_key)."""
         if isinstance(query, Keyboard.Key):
             if mode is not None and query.mode != mode:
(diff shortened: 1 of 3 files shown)


_stbt/keyboard.py
```diff
@@ -770,6 +805,86 @@ def wait_for_transition_to_end(
...
+                    continue
+                if a.x == b.x and a.right == b.right:
+                    a.bottom = b.bottom
+                    del out[m]
+                    modified = True
+                    break
+                # Only support rectangles for now:
+                if b.x < a.right and a.x < b.right:
+                    # Overlapping, but not equal
+                    raise NotImplementedError(
```
drothlis, line 850: ValueError perhaps?

drothlis (approved):

_stbt/keyboard.py
wmanley: I see `ValueError` like an HTTP 4xx error - there's something wrong with the value that the user has passed.  This is more like a TODO: the user asked for something reasonable, but we just haven't got round to implementing it yet.  Maybe `NotImplementedError` isn't appropriate for that, but that's the reasoning anyway.

Maybe we need a `ToDoError` that points users at an issue in our GitHub so they can voice that they actually need this functionality.

wmanley: IOW the meaning in this case is "not implemented yet" error rather than "not implementable" error.

Also implied: I would consider a future change such that this does not raise `NotImplementedError` to be backwards compatible.

PR #819, opened 2023-11-27
Add OcrApprox: Approximate string equality for text read with `ocr`

drothlis:
`ocr` sometimes mistakes some characters, such as "O" instead of "0", especially when reading short fragments of text without enough context. Wrapping the text in an `OcrApprox` object will treat such characters as equal to each other. It also ignores spaces. For example:

    >>> OcrApprox("hello") == "hel 10"
    True


wmanley:
Inspired by [pytest.approx](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-approx) which is used for comparing floats.

wmanley (commented):
:+1: in principle, implementation could be tighter using dataclasses

drothlis:
I'm going to ditch this API because of the issues with hashability and transitivity. I've raised #820 with a simpler API (`ocr_eq` function that takes 2 strings and returns True / False).

PR #820, opened 2023-11-28
Add ocr_eq: Approximate string equality for text read with `ocr`

drothlis:


wmanley (approved):
Yes, a much safer interface

_stbt/ocr.py
```diff
@@ -452,6 +452,43 @@ def set_global_ocr_corrections(corrections: CorrectionsT):
...
+def ocr_eq(a: str, b: str) -> bool:
+    """Compare two strings for equality, ignoring common OCR errors.
+
+    `ocr` sometimes mistakes some characters, such as "O" instead of "0",
+    especially when reading short fragments of text without enough context.
+    ``ocr_eq`` wil treat such characters as equal to each other. It also
+    ignores spaces. For example:
+
+    >>> ocr_eq("hello", "hel 10")
+    True
```
wmanley, line 464: Missing "added in v34"

drothlis: Done in 2e60e54.

PR #821, opened 2023-11-30
ocr_eq docstring: Document ocr_eq.replacements and ocr_eq.normalize

drothlis:


drothlis:
The generated sphinx documentation looks like this:

image

drothlis:
@wmanley what do you think about documenting these?

wmanley (commented):
I'm happy for this to be public as long as VSCode understands it.  If not then it would be better as `ocr_eq_replacements`.

drothlis:
> I'm happy for this to be public as long as VSCode understands it. If not then it would be better as `ocr_eq_replacements`.

VSCode doesn't understand it — doesn't autocomplete, doesn't catch typos — but at least it doesn't mark code like this as an error:

    stbt.ocr_eq.replacements["x"] = "X"

It thinks it's a function of type `Any`:

image

wmanley:
https://github.com/python/mypy/issues/2087

wmanley:
I've just realised that there's a subtlety here when adding additional replacements.  You could do:

```py
ocr_eq.replacements["İ"] = "I"
```

But then this would fail:

```py
ocr_eq("İ", "I")
```

Because that would now be equivalent to `"I" == "l"`

drothlis:
Or we could run through the replacements twice?

drothlis:
> I've just realised that there's a subtlety here when adding additional replacements.

Fixed in #823.

PR #822, opened 2023-12-16
Keyboard: Suport Page Object with `focus` property instead of `selection`

drothlis:
"focus" is going to be our official name going forward; "selection" is ambiguous because you can have "selected" something and then move the "focus" somewhere else without changing the active "selection".

I'm adding this to Keyboard for consistency with [navigate_1d].

For backwards compatibility we use the `selection` property, if it exists (even if `focus` also exists, to avoid changing behaviour for existing users, in case they have both a `focus` and a `selection`, even though it seems unlikely... but doing it this way means we can release this in a v34 minor release).

[navigate_1d]: https://stb-tester.com/manual/python-api#stbt.navigate_1d

### DIFF
--- a/_stbt/keyboard.py
+++ b/_stbt/keyboard.py
@@ -31,17 +31,17 @@ class Keyboard():
     by specifying two things:
 
     * A `Directed Graph`_ that specifies the navigation between every key on
-      the keyboard. For example: When *A* is selected, pressing *KEY_RIGHT* on
+      the keyboard. For example: When *A* is focused, pressing *KEY_RIGHT* on
       the remote control goes to *B*, and so on.
 
-    * A `Page Object`_ that tells you which key is currently selected on the
+    * A `Page Object`_ that tells you which key is currently focused on the
       screen. See the ``page`` parameter to `enter_text` and `navigate_to`.
 
     The constructor takes the following parameters:
 
     :param str|numpy.ndarray|Mask|Region mask:
         A mask to use when calling `stbt.press_and_wait` to determine when the
-        current selection has finished moving. If the search page has a
+        current focus has finished moving. If the search page has a
         blinking cursor you need to mask out the region where the cursor can
         appear, as well as any other regions with dynamic content (such as a
         picture-in-picture with live TV). See `stbt.press_and_wait` for more
@@ -109,18 +109,18 @@ class Keyboard():
                 ...
 
             @property
-            def selection(self):
-                """Returns the selected key.
+            def focus(self):
+                """Returns the focused key.
 
                 Used by `Keyboard.enter_text` and `Keyboard.navigate_to`.
 
-                Note: The reference image (selection.png) is carefully cropped
+                Note: The reference image (focus.png) is carefully cropped
                 so that it will match the normal keys as well as the larger
                 "SPACE", "DELETE" and "CLEAR" keys. The middle of the image
                 (where the key's label appears) is transparent so that it will
                 match any key.
                 """
-                m = stbt.match("selection.png", frame=self._frame)
+                m = stbt.match("focus.png", frame=self._frame)
                 if m:
                     return kb.find_key(region=m.region)
                 else:
@@ -154,14 +154,14 @@ class Keyboard():
       object, and then use `add_key`, `add_transition`, `add_edgelist`, and
       `add_grid` to build the model of the keyboard.
     * Removed the ``stbt.Keyboard.Selection`` type. Instead, your Page Object's
-      ``selection`` property should return a `Key` value obtained from
+      ``focus`` property should return a `Key` value obtained from
       `find_key`.
 
     Changed in v33:
 
     * Added class `stbt.Keyboard.Key` (the type returned from `find_key`). This
       used to be a private API, but now it is public so that you can use it in
-      type annotations for your Page Object's ``selection`` property.
+      type annotations for your Page Object's ``focus`` property.
     * Tries to recover from missed or double keypresses. To disable this
       behaviour specify ``retries=0`` when calling `enter_text` or
       `navigate_to`.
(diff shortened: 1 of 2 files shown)


_stbt/keyboard.py
```diff
@@ -804,6 +808,20 @@ def wait_for_transition_to_end(
                                                timeout_secs=timeout_secs,
                                                stable_secs=stable_secs)
 
+    def _get_focus(self, page):
+        # `focus` is the official name we support; `selection` for backward
+        # compatibility.
+        try:
+            return page.selection
```
wmanley, line 815:
```suggestion
            return "selection", page.selection
```

_stbt/keyboard.py
```diff
@@ -735,10 +738,11 @@ def navigate_to(
             self.G_ = _strip_shift_transitions(self.G)
 
         assert page, "%s page isn't visible" % type(page).__name__
-        assert page.selection in self.G_, \
-            "page.selection (%r) isn't in the keyboard" % (page.selection,)
+        current = self._get_focus(page)
+        assert current in self.G_, \
+            "page.%s (%r) isn't in the keyboard" % (
+                self._get_focus_property_name(page), current)
```
wmanley, line 744:
```suggestion
        prop_name, current = self._get_focus(page)
        assert current in self.G_, \
            "page.%s (%r) isn't in the keyboard" % (prop_name, current)
```

_stbt/keyboard.py
```diff
@@ -804,6 +808,20 @@ def wait_for_transition_to_end(
...
+        try:
+            return page.selection
+        except AttributeError:
+            pass
+        return page.focus
+
+    def _get_focus_property_name(self, page):
+        if hasattr(page, "selection"):
+            return "selection"
+        return "focus"
```
wmanley, line 823:
```suggestion
        return "focus", page.focus
```

_stbt/keyboard.py
```diff
@@ -753,21 +757,21 @@ def navigate_to(
             for key, immediate_targets in keys:
                 transition = self.press_and_wait(key, stable_secs=0.5)
                 assert transition.status != TransitionStatus.STABLE_TIMEOUT, \
-                    "Selection didn't stabilise after pressing %s" % (key,)
+                    "Focus didn't stabilise after pressing %s" % (key,)
```
wmanley, line 760:
```suggestion
                    "%s didn't stabilise after pressing %s" % (prop_name.capitalize(), key,)
```

_stbt/keyboard.py
```diff
@@ -753,21 +757,21 @@ def navigate_to(
...
                     transition = self.wait_for_transition_to_end(
                         initial_frame=page._frame, stable_secs=2)
                     assert transition.status != \
                         TransitionStatus.STABLE_TIMEOUT, \
-                        "Selection didn't stabilise after pressing %s" % (key,)
+                        "Focus didn't stabilise after pressing %s" % (key,)
                     page = page.refresh(frame=transition.frame)
                     assert page, "%s page isn't visible" % type(page).__name__
-                    current = page.selection
+                    current = self._get_focus(page)
```
wmanley, line 774:
```suggestion
                    prop_name, current = self._get_focus(page)
```

_stbt/keyboard.py
```diff
@@ -753,21 +757,21 @@ def navigate_to(
...
                 if (current not in immediate_targets and
                         not verify_every_keypress):
-                    # Wait a bit longer for selection to reach the target
+                    # Wait a bit longer for focus to reach the target
                     transition = self.wait_for_transition_to_end(
                         initial_frame=page._frame, stable_secs=2)
                     assert transition.status != \
                         TransitionStatus.STABLE_TIMEOUT, \
-                        "Selection didn't stabilise after pressing %s" % (key,)
+                        "Focus didn't stabilise after pressing %s" % (key,)
```
wmanley, line 771:
```suggestion
                        "%s didn't stabilise after pressing %s" % (prop_name.capitalize(), key)
```

_stbt/keyboard.py
```diff
@@ -753,21 +757,21 @@ def navigate_to(
             for key, immediate_targets in keys:
                 transition = self.press_and_wait(key, stable_secs=0.5)
                 assert transition.status != TransitionStatus.STABLE_TIMEOUT, \
-                    "Selection didn't stabilise after pressing %s" % (key,)
+                    "Focus didn't stabilise after pressing %s" % (key,)
                 page = page.refresh(frame=transition.frame)
                 assert page, "%s page isn't visible" % type(page).__name__
-                current = page.selection
+                current = self._get_focus(page)
```
wmanley, line 763:
```suggestion
                prop_name, current = self._get_focus(page)
```

_stbt/keyboard.py
```diff
@@ -690,8 +694,8 @@ def enter_text(
                                     page, verify_every_keypress, retries)
             self.press_and_wait("KEY_OK", stable_secs=0.5, timeout_secs=1)  # pylint:disable=stbt-unused-return-value
             page = page.refresh()
-            log.debug("Entered %r; the selection is now on %r",
-                      letter, page.selection)
+            log.debug("Entered %r; the focus is now on %r",
+                      letter, self._get_focus(page))
```
wmanley, line 698:
```suggestion
            prop_name, current = self._get_focus(page)
            log.debug("Entered %r; the %s is now on %r",
                      letter, prop_name, current)
```

wmanley (commented):
My only backwards compatibility concern is for customers parsing their logs.  It's minor, but easy to work around nayway (see suggestions).

Overall I far prefer the name "focus" to "selection" for the reasons you've described.

_stbt/keyboard.py
drothlis: Done.

PR #823, opened 2024-01-04
ocr_eq.replacements: Custom __setitem__

drothlis:
There's a subtlety when adding your own replacements, if the replaced value is itself already a key in the replacements. Previously, you would have to do this:

    ocr_eq.replacements["İ"] = "I"                    # incorrect!
    ocr_eq.replacements["İ"] = ocr_eq.normalise("I")  # correct

Now, `ocr_eq.replacements["İ"] = "I"` is correct.

If a user replaces `ocr_eq.replacements` wholesale with a new dict, it'll lose this behaviour; but then that's their problem. Either replace it once with a single dict that you don't modify, or add to our one.

Note: I subclass [collections.UserDict] instead of subclassing `dict` directly, so that `copy()` preserves my subclass type (needed for my unit tests).

[collections.UserDict]: https://docs.python.org/3/library/collections.html#collections.UserDict

### DIFF
--- a/tests/test_ocr.py
+++ b/tests/test_ocr.py
@@ -11,7 +11,7 @@ import _stbt.config
 import stbt_core as stbt
 from _stbt import imgproc_cache
 from _stbt.imgutils import load_image
-from _stbt.ocr import _tesseract_version
+from _stbt.ocr import _tesseract_version, Replacements
 from _stbt.utils import named_temporary_directory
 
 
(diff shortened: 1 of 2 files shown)


wmanley:
I think there's another subtlety (not tested).

```py
ocr_eq.replacements["a"] = "b"
ocr_eq.replacements["b"] = "c"
assert ocr_eq("a", "b")
# is the same as:
assert "b" == "c"
```

wmanley:
Maybe we need a list of sets of characters to consider equivalent? But this would preclude normalise.

drothlis:
> I think there's another subtlety (not tested):
> 
> ```python
> ocr_eq.replacements["a"] = "b"
> ocr_eq.replacements["b"] = "c"
> assert ocr_eq("a", "b")
> # is the same as:
> assert "b" == "c"
> ```

That's OK, isn't it? I mean, it's exactly what you asked for by specifying those replacements.

wmanley:
> That's OK, isn't it? I mean, it's exactly what you asked for by specifying those replacements.

The problem arises when adding the "a" replacement is in different code to the "b" replacement.  It means that when adding the "b" replacement you need to be aware of what's already in the dict.

So the motivation for this PR is:

> There's a subtlety when adding your own replacements, if the replaced value is itself already a key in the replacements

but this additional issue is 

>There's a subtlety when adding your own replacements, if the replaced **key** is itself already a **value** in the replacements

drothlis:
I still don't see the problem. I had already thought of this scenario and had added a test just to make sure that it works like I expect. It's not like you're going to specify replacements in different parts of the code. It'll all be in a single place, probably in your `tests/__init__.py`.

Adding replacements *anywhere* in the code (as long as it gets executed) is always going to cause an effect in other code elsewhere, which might surprise you if you didn't know about the original replacement. The important thing is that the order in which you add these replacements doesn't change the behaviour of `ocr_eq`.

tests/test_ocr.py
```diff
@@ -484,21 +484,40 @@ def test_ocr_eq_shouldnt_match(a, b):
     assert not stbt.ocr_eq(b, a)  # pylint:disable=arguments-out-of-order
 
 
+@contextmanager
+def temporary_ocr_eq_replacements():
+    orig = stbt.ocr_eq.replacements.copy()
+    try:
+        yield
+    finally:
+        stbt.ocr_eq.replacements = orig.copy()
```
wmanley, line 493: Don't need two copies here

```suggestion
        stbt.ocr_eq.replacements = orig
```

wmanley (approved):

wmanley:
Sounds like it's not a problem then if the test passes.

PR #824, opened 2024-01-04
keyboard: Prefer deterministic path over shorter non-deterministic path to a different (but suitable) key

drothlis:
This is an edge case that we found when navigating the Channel 4 app on Apple TV. There are 3 keyboard modes, and you press KEY_PLAY to switch between them. The exact key you land on can vary based on previous state, so to simplify our model we just said that pressing KEY_PLAY can land you on *any* key of the next mode. This resulted in buggy behaviour: `Keyboard.navigate_to("BACKSPACE")` would press `KEY_PLAY` (over and over, until `navigate_timeout`).

Here's what was happening: BACKSPACE is available in all modes, so `navigate_to` was calling `_keys_to_press` with 3 possible targets. We do take into account the weights when calculating the shortest path to any one of those targets, but we weren't taking into account the weights when choosing which of those 3 paths was the shortest. Because of our lazy hack when modelling the keyboard, it thought that a single `KEY_PLAY` might get us to BACKSPACE in the next mode.

The fix is: It should always prefer a deterministic path over a non-deterministic path† no matter how short the latter is.

† That's what we call a path in our model where pressing a key might land you on any one of several states.

Note: I have committed the screenshots so you can see what the keyboard looks like, but I don't actually use the screenshots in the test.


drothlis:
Question: Is this worth merging? It's arguably a bug in the model rather than `Keyboard`.

If we do want to merge it I'd want to save it until v35 as I'm nervous about changing behaviour in a minor release. Or put it behind some kind of configuration.

drothlis:
Definitely worth merging: It doesn't make sense to apply the weighting in some cases but not others.

wmanley:
> Question: Is this worth merging? ~It's arguably a bug in the model rather than `Keyboard`.~

Yes, definitely.

> If we do want to merge it I'd want to save it until v35 as I'm nervous about changing behaviour in a minor release. Or put it behind some kind of configuration.

I'd be reluctant to hold off until 35, unless we're willing to make a release now ish. I've been quite happy with our recent approach of being strict with backwards compatibility so we can roll out improvements to customers as we develop them. So it looks like some configuration would be worthwhile here - unless with sufficient understanding of this issue we could confirm that this change would have no effect on existing customers' test-packs.

A brute force approach would be to test this change against all the `stbt.Keyboard`s we can find, trying to navigate from all keys to all other keys - checking that no behaviour changes.  Even better would be something we could grep for.

PR #825, opened 2024-01-09
Grid: type annotations

drothlis:
`grid.py` itself is now free of pylance warnings.

TODO:

- [ ] Check callers in the unit tests.
- [ ] Check how "T" renders in the generated documentation.


drothlis:
Note: `Grid.Cell` is now a `dataclass` instead of a `namedtuple` — could this be considered an API change?

drothlis:
> Unfortunately it now complains about the type of Grid.Cell.data:
> 
> > Type variable "T" has no meaning in this context

Copilot chat suggests:

> One possible workaround is to define Cell as a separate class outside of Grid, and pass the type variable T as a parameter to both classes. Here's how you can do it:
>
> ```python
> @dataclasses.dataclass
> class Cell(Generic[T]):
>     index: int
>     position: Position
>     region: Region
>     data: Optional[T]
> 
> class Grid(Generic[T]):
>     # ...
>     Cell: Type[Cell[T]]
>     # ...
> 
>     def __init__(
>         self,
>         region: Region,
>         cols: Optional[int] = None,
>         rows: Optional[int] = None,
>         data: Optional[Sequence[Sequence[T]]] = None,
>     ):
>         # ...
>         self.Cell = Cell[T]
> ```
>
> In this code, Cell is a generic class that takes a type variable T. The type of Cell.data is Optional[T], which means it can be either None or an instance of type T. When you create a Grid with a specific type, the Cell instances in that grid will also have data of that type.

Who knows if that'll work, but I won't bother.

PR #827, opened 2024-01-22
get_config: Better type annotations

wmanley:
This way VSCode knows that `ip_address is not None` in this case:

    ip_address = stbt.get_config("device_under_test", "ip_address")

### DIFF
--- a/_stbt/config.py
+++ b/_stbt/config.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import configparser
 import enum
 import os
+import typing
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


drothlis (approved):
:+1: 

I suppose here it couldn't work out what `DefaultT` is, when `default=NoDefault`:

```
def get_config(
    section: str,
    key: str,
    default: DefaultT | Type[NoDefault] = NoDefault,
    type_: Callable[[str], T] = str,
) -> "T | DefaultT":
```

I wonder why we have DefaultT instead of `default: T | Type[NoDefault]`. Maybe to support `default=None`?

PR #828, opened 2024-02-07
wait_until type annotations

drothlis:


wmanley (approved):
This will be handy. :+1:.  The two overloads overlap, but if it works it works.

PR #831, opened 2024-02-20
ocr: Make `upsample`'s default value configurable

drothlis:
In some GUIs it might be more accurate to disable the upsampling.

See also: #146 (from 2014).


wmanley (commented):
I’m slightly skeptical that it makes sense to adjust this globally, rather than on a frameobject property by property basis, but that’s not an objection to merging this.

We’ll need logging to say what value was actually used, as it will now be harder to tell from inspection. I think we already log regions, so it could be naturally added there.

Backwards compatibility note: passing None previously disabled upsampling, whereas now it will enable it. I doubt this will affect any real customers, but it might be worth a cursory glance.

In principle: 👍

drothlis:
> I’m slightly skeptical that it makes sense to adjust this globally, rather than on a frameobject property by property basis, but that’s not an objection to merging this.

Probably true, but at least it'll allow experimentation without invasive changes to the test-pack Python code.

> We’ll need logging to say what value was actually used, as it will now be harder to tell from inspection. I think we already log regions, so it could be naturally added there.

The logging is a bigger task that needs more thought. We don't currently log lang, text_color_threshold, or engine (which all come from stbt.conf).

PR #837, opened 2024-06-24
Internal: Add `logging.draw_source_region`

wmanley:
So we can show which regions are being ocred in the object repo.


drothlis (approved):
We already record `ImageLogger.data["region"]` which is the same thing (the input region aka "Region Of Interest"). Maybe we could change `ImageLogger` to log / draw this even when the rest of the ImageLogger debug is disabled? Nah, not worth the complexity. `ImageLogger` needs a total overhaul anyway.

PR #838, opened 2024-07-01
stbt.Keyboard: Allow disjoint modes

drothlis:
If there are two "A"s in the keyboard, it should succeed if it can find a path to one of them.

For convenience I'm modelling a Keyboard that has 2 different appearances (on different STBs) as a single Keyboard with 2 different modes (and a single Page Object that handles both appearances). Both modes have all the letters, and there's no way to get from one mode to the other. This way I don't have to make my Page Object remember which `Keyboard` instance to use — there's only one `Keyboard` instance.

### DIFF
--- a/_stbt/keyboard.py
+++ b/_stbt/keyboard.py
@@ -907,12 +907,22 @@ def _minimal_query(query):
 
 
 def _keys_to_press(G, source, targets):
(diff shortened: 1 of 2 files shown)


wmanley (approved):
Yep. Backwards compatible too.

PR #839, opened 2024-07-22
imgproc_cache: Allow `Color` argument types to cached functions

drothlis:


drothlis:
I considered making `memoize` and `memoize_iterator` catch `JsonEncodeError` in addition to `NotCachable`, but then we would never notice this type of bug. I decided that: Since it's only used internally by stb-tester, better to fail visibly so that we can notice and fix it.

wmanley (approved):

PR #841, opened 2024-08-22
LircRemote: Close temporary connection to lircd promptly on error

wmanley:
Untested


drothlis (approved):
I suppose the exception traceback could be keeping the connection socket open.

PR #847, opened 2025-02-12
AdbDevice.get_frame: Handle PNG data not corrupted by adb shell

drothlis:
Older versions of `adb shell` would convert every LF in the PNG data to CRLF. Apparently newer versions of adb don't do this, so our "fix" actually corrupts the PNG and we get "AdbError: Failed to capture screenshot from android device".

https://stackoverflow.com/a/13593914:

> It's worth noting that the PNG file format is explicitly designed to catch exactly this (and related) problems. The magic number begins with 0x89 to catch anything that strips high bits, followed by "PNG" so you can easily tell what's in the file, followed by CR LF to catch various ASCII line converters, then 0x1a to trap old MS-DOS programs that used Ctrl-Z as a special end-of-file marker, and then a lone LF. By looking at the first few bytes of the file you can tell exactly what was done to it.

### DIFF
--- a/_stbt/android.py
+++ b/_stbt/android.py
@@ -258,6 +258,7 @@ class AdbDevice():
                             timeout=60, capture_output=True) \
                        .stdout
             assert isinstance(data, bytes)
+            header = data[:10]
             if data.startswith(b"\x89PNG\r\r\n"):
                 # Older versions of `adb shell` convert LF to CRLF.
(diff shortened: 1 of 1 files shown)


drothlis:
I have reproduced the issue, and tested this fix, with an NVidia Shield, v9.1.1.

wmanley (approved):
Clever header.

PR #849, opened 2025-02-26
Region.bounding_box: typing overload when we know it'll return a Region

drothlis:


drothlis:
Before the fix:
https://github.com/stb-tester/stb-tester/actions/runs/13545312465/job/37855361120?pr=849

```
Run pyright --version && make check-pyright

added 1 package, and audited 2 packages in 2s

found 0 vulnerabilities
pyright 1.1.389
WARNING: there is a new pyright version available (v1.1.389 -> v1.1.394).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`

virtual-stb support disabled
PYTHONPATH=$PWD pyright
/__w/stb-tester/stb-tester/tests/test_pyright.py
  /__w/stb-tester/stb-tester/tests/test_pyright.py:15:21 - error: "width" is not a known attribute of "None" (reportOptionalMemberAccess)
1 error, 0 warnings, 0 informations 
WARNING: there is a new pyright version available (v1.1.389 -> v1.1.394).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`

make: *** [Makefile:242: check-pyright] Error 1
```

_stbt/types.py
```diff
@@ -72,6 +72,10 @@ def intersect(cls, *args: RegionT) -> RegionT:
                 return None
         return Region.from_extents(*out)
 
+    @typing.overload
+    def bounding_box(cls, *args: Region) -> Region:
+        ...
```
wmanley, line 77: I think you'll need a second overload. I don't think the type checker will take the real function into account if an overload is present.

drothlis: Fixed.

PR #851, opened 2025-03-25
ocr & get_config changes for Object Repository

drothlis:
This allows overriding the default language for a particular device type. This will be particularly useful for OCR properties in the Object Repository.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -156,7 +156,7 @@ def temporary_config(contents, prefix="stbt-test-config"):
             f.write(dedent(contents))
         _config_init(force=True)
         try:
-            yield
+            yield filename
         finally:
             os.environ["STBT_CONFIG_FILE"] = original_env
             _config_init(force=True)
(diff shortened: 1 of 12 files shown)


tests/test_config.py
```diff
@@ -193,3 +193,30 @@ def test_get_config_with_default_value():
...
+    # https://docs.python.org/3.10/library/configparser.html#configparser.ExtendedInterpolation
+    with temporary_config(CONFIG_WITH_INTERPOLATION):
+        assert get_config("ocr", "lang") == "deu"
+        # Basic interpolation syntax is treated literally:
+        assert get_config("ocr", "basic") == "%(lang)s"
+
+
+def test_that_set_config_preserves_interpolation():
+    with temporary_config(CONFIG_WITH_INTERPOLATION) as filename:
+        set_config("device_type", "lang", "eng")
```
wmanley, line 220: I know this isn't a public API, but out of interest: What does `set_config` do in this case if you pass in a string with $?

wmanley (approved):

tests/test_config.py
drothlis: It doesn't allow unescaped $:

```
  File "tests/test_config.py", line 220, in test_that_set_config_preserves_interpolation
    set_config("device_type", "lang", "$eng")
  File "_stbt/config.py", line 145, in set_config
    config.set(section, option, value)
  File "/usr/lib/python3.10/configparser.py", line 1206, in set
    super().set(section, option, value)
  File "/usr/lib/python3.10/configparser.py", line 896, in set
    value = self._interpolation.before_set(self, section, option,
  File "/usr/lib/python3.10/configparser.py", line 464, in before_set
    raise ValueError("invalid interpolation syntax in %r at "
ValueError: invalid interpolation syntax in '$eng' at position 0
```

wmanley: Right: so when using `set_config` the caller is responsible for encoding, but at least it won't fail silently in the most obvious non-encoded case. In cases where the string happens to contain a valid interpolation sequence then the data won't round-trip through `set_config`/`get_config`.

PR #853, opened 2025-05-28
control: RedRat BT-Server API for BlueRat hardware

drothlis:


drothlis:
Tested with:

    ./stbt_control.py --control bluerat:localhost:4254:25415:98-F7-81-91-8A-AB down

PR #863, opened 2026-04-23
Enable pyright "standard" mode

drothlis:
And fix the new warnings it reports:

* _stbt.config:

  > Type "Enum | IntEnum" is not assignable to return type "T@get_config | DefaultT@get_config"

  It seems to me that it *should* match the type. :shrug: 

* _stbt.control:

  > "error_msg" is possibly unbound

  Pyright doesn't know that `response.raise_for_status()` will have definitely raised.

* _stbt.imgutils:

  > Method "\_\_eq__" overrides class "object" in an incompatible manner
  > Parameter 2 type mismatch: base parameter is type "object", override parameter is type "Color"

* The rest are "possibly unbound" warnings, mostly because pyright doesn't know that the loop was guaranteed to run at least once.

### DIFF
--- a/_stbt/control.py
+++ b/_stbt/control.py
@@ -560,6 +560,7 @@ class RedRatHttpControl(RemoteControl):
                 error_msg = response.json()["error"]
             except Exception:  # pylint:disable=broad-except
                 response.raise_for_status()
+            assert error_msg  # pyright:ignore[reportPossiblyUnboundVariable]
             if re.match("Command '.*' is not known.", error_msg):
                 raise UnknownKeyError(error_msg)
             else:
--- a/_stbt/match.py
+++ b/_stbt/match.py
@@ -488,6 +488,7 @@ def wait_for_match(
     last_pos = Position(0, 0)
     image = load_image(image)
     debug("Searching for " + (image.relative_filename or "<Image>"))
+    res = None
     for frame in frames:
         res = match(image, match_parameters=match_parameters,
                     region=region, frame=frame)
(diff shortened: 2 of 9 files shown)


drothlis:
In commit 751868f50 I had said:

> I have configured it to "basic" mode; the default is "standard", which enables some extra checks such as "reportPossiblyUnboundVariable". We can enable these when we fix the issues. "standard" would report only 22 additional issues (on the files not currently ignored) so I would recommend fixing those first and enabling "standard" mode, before we start fixing other files.

_stbt/match.py
```diff
@@ -611,6 +615,9 @@ def _find_candidate_matches(image, template, match_parameters, imglog):
     image_pyramid = _build_pyramid(image, len(template_pyramid))
     roi_mask = None  # Initial region of interest: The whole image.
 
+    (best_match_position, certainty, heatmap, heatmap_scale, level, matched,
+     threshold) = (None, None, None, None, None, None, None)
```
wmanley, line 619: It's a shame that this is required. https://github.com/microsoft/pyright/issues/844 . It would be nice if we could mark loops as executing at least once.

I'm surprised that this didn't cause issues with "`best_match_position` is possibly None" errors later.

_stbt/control.py
```diff
@@ -560,6 +560,7 @@ def press(self, key):
                 error_msg = response.json()["error"]
             except Exception:  # pylint:disable=broad-except
                 response.raise_for_status()
```
wmanley, line 562: I'd put `assert False` here in the except block to express that `raise_for_status` will definitely have raised.

wmanley (approved):

PR #864, opened 2026-04-23
detect_motion, wait_for_motion: Add threshold parameter

drothlis:
...and deprecate `noise_threshold`. We can remove it in v35.

`noise_threshold` is mis-named, and inconsistent with other APIs.

Note: Branched from #863.

### DIFF
--- a/_stbt/motion.py
+++ b/_stbt/motion.py
@@ -68,10 +68,15 @@ def detect_motion(
     `Region` to ``mask`` instead. You can't specify ``mask`` and ``region``
     at the same time.
 
-    Changed in v34: The difference-detection algorithm takes color into account.
-    Renamed the ``noise_threshold`` parameter to ``threshold``, and changed its
-    range (from 0.0-1.0 to 0-255), sense (from "bigger is stricter" to "smaller
-    is stricter"), and default value (from 0.84 to 25).
+    Changed in v34:
+
+    * The difference-detection algorithm takes color into account.
+    * Renamed the ``noise_threshold`` parameter to ``threshold``.  The old name
+      is still accepted as a keyword argument but is deprecated.
+    * Changed the way `threshold` is interpreted:
+        * Range is now 0-255, was 0.0-1.0
+        * It's now "smaller is stricter", was "bigger is stricter"
+        * Default value: was 0.84 now 25.
     """
     if frames is None:
         import stbt_core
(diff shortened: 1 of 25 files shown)


_stbt/motion.py
```diff
@@ -14,10 +14,12 @@
 
 def detect_motion(
     timeout_secs: float = 10,
-    noise_threshold: Optional[int] = None,
+    threshold: Optional[int] = None,
```
wmanley, line 17: That's quite clever - so positional arguments will resolve to the new one, and kwargs will to the old.  The only issue would be if someone is calling `inspect` or something on this function.

This should probably be `Optional[float]` - we're only using it for comparison, and with an alternative differ the numbers here might be smaller.

_stbt/motion.py
```diff
@@ -39,14 +41,13 @@ def detect_motion(
...
-    :param int noise_threshold:
+    :param int threshold:
         The difference in pixel intensity to ignore. Valid values range from 0
         (any difference is considered motion) to 255 (which would never report
         motion).
 
         This defaults to 25. You can override the global default value by
-        setting ``noise_threshold`` in the ``[motion]`` section of
-        :ref:`.stbt.conf`.
+        setting ``threshold`` in the ``[motion]`` section of :ref:`.stbt.conf`.
```
wmanley, line 50: Commit message says:

> Note that I haven't added `motion.threshold` to `_stbt/stbt.conf` because that would screw up the backward compatibility logic.

Oh I see - we read the config option, but just don't include it in the default stbt.conf.

_stbt/motion.py
```diff
@@ -84,8 +88,22 @@ def detect_motion(
...
-    if noise_threshold is None:
-        noise_threshold = get_config('motion', 'noise_threshold', type_=int)
+    if noise_threshold is not None:
+        if threshold is not None:
+            raise ValueError("Cannot specify 'noise_threshold' and 'threshold' "
+                             "at the same time")
+        warnings.warn(
+            "stbt.detect_motion: The 'noise_threshold' parameter is "
+            "deprecated; use 'threshold' instead",
+            DeprecationWarning, stacklevel=2)
```
wmanley: Don't add this yet. If we deploy this then customers will start seeing new warnings unrelated to any change that they've made. We don't want to cause them busy-work for an aesthetic change.

I think the warning should only be added in v35, or at least after we've had a chance to raise PRs for our customers updating their code.

_stbt/motion.py
```diff
@@ -68,10 +68,10 @@ def detect_motion(
...
     at the same time.
 
-    Changed in v34: The difference-detection algorithm takes color into
-    account. The ``noise_threshold`` parameter changed range (from 0.0-1.0 to
-    0-255), sense (from "bigger is stricter" to "smaller is stricter"), and
-    default value (from 0.84 to 25).
+    Changed in v34: The difference-detection algorithm takes color into account.
+    Renamed the ``noise_threshold`` parameter to ``threshold``, and changed its
+    range (from 0.0-1.0 to 0-255), sense (from "bigger is stricter" to "smaller
+    is stricter"), and default value (from 0.84 to 25).
```
wmanley: Taking `noise_threshold` as a keyword parameter is retained for compatibility. I know it's mentioned above too, but I think it's important in a changelog to emphasise that we don't break customer code.

wmanley (approved):
I've made a few changes.  Please review.

drothlis:
:+1: thanks.

PR #866, opened 2026-04-23
press_and_wait: Add threshold & differ parameters

drothlis:
TODO:

- [x] Tests.


wmanley (approved):
I have some reservations about greater visibility of `Differ` and conflicting reservations about threading through `min_size`, `threshold`, etc - but I think it exposes important functionality, and is better than anything else I can think of.

PR #867, opened 2026-04-23
Lewis/use tmp path in unittests

LewisHaley:
I'm looking at contributing some changes to
1. unify the various config files into the modern pyproject.toml file
2. update the packaging of the library pushed to PyPi - hoping to avoid the build of libstbt.so within the Makefile
3. add support for <=python3.13

In order to do all that, I'd benefit from being able to run CI locally, and thus this PR contains a few changes that ultimately let me run pytest against the whole repo with everything passing.

Obviously if you're not interested in the above, please let me know.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,8 @@
 # pytest configuration
 
 import os
+from pathlib import Path
+from typing import Generator
 
 import pytest
 
@@ -18,3 +20,14 @@ def test_pack_root():
         yield
     finally:
         stbt_core.TEST_PACK_ROOT = None
+
+
+@pytest.fixture(name="cwd_is_tmp_path")
+def _cwd_is_tmp_path(tmp_path: Path) -> Generator[None, None, None]:
+    """Change the CWD to a temporary directory for the duration of a test."""
+    orig_cwd = os.getcwd()
+    os.chdir(tmp_path)
+    try:
+        yield
+    finally:
+        os.chdir(orig_cwd)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,7 @@ from typing import Generator
 
 import pytest
 
+import _stbt.config
 import _stbt.logging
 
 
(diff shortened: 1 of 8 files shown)


tests/test_config.py
```diff
@@ -108,88 +112,100 @@ class MyIntEnum(enum.IntEnum):
...
+
+    The callable has the following signature:
+    * contents: str - the contents of the config file
+    * filename: str = "stbt-test.cfg" - the name of the config file to
+      create
+    """
+    def factory(contents, filename="stbt-test.cfg"):
+        filename = tmp_path / filename
+        original_env = os.environ.get("STBT_CONFIG_FILE", "")
+        monkeypatch.setenv(
```
drothlis, line 165: I see, `monkeypatch` is a pytest fixture. Neat!

tests/conftest.py
```diff
@@ -18,3 +22,23 @@ def test_pack_root():
...
+    # So that the output directory name doesn't depend on how many tests
+    # were run before this one.
+    ImageLogger._frame_number = itertools.count(1)
+    orig_root_dir = ImageLogger._root_dir
+    tmp_stbt_debug_dir = tmp_path / "stbt-debug"
+    ImageLogger._root_dir = str(tmp_stbt_debug_dir)
+    try:
+        yield tmp_stbt_debug_dir
+    finally:
+        ImageLogger._root_dir = orig_root_dir
```
drothlis: I'd prefer a fixture that does `chdir` into a temporary directory. Then you won't need to modify `ImageLogger`. I think the tests will just work, because `stbt.match` etc look up the files relative to the caller Python file.

If this fixture could (optionally) not delete the directory afterwards, say if you've set KEEP_TMPDIR=1, and print the location, that would be useful for debugging when the test fails.

drothlis (changes requested):
Thanks @LewisHaley, looks like some much needed cleanup. Just one change to the stbt-debug tests please:

drothlis:
> I'm looking at contributing some changes to
> 
>     1. unify the various config files into the modern pyproject.toml file
> 
>     2. update the packaging of the library pushed to PyPi - hoping to avoid the build of libstbt.so within the Makefile
> 
>     3. add support for <=python3.13

FYI we don't have the bandwidth to review significant changes, but these all seem small & reasonable tidyups. (Assuming for pyproject.toml you only mean the dev-related config files like pylint, not stbt.conf.)

tests/conftest.py
LewisHaley: Please see https://github.com/stb-tester/stb-tester/pull/867/commits/19693e23618e4dbab2433a8614843b8645fde575.
In the event the test fails, pytest keeps the directory anyway. On Linux, this is would be something like `/tmp/pytest-of-lewis/pytest-current/test_ocr_debugcurrent/stbt-debug/`.

drothlis (approved):
Thanks Lewis. Please squash the fixup commit and I'll merge.

PR #869, opened 2026-05-07
imgproc_cache: Ignore EIO errors

wmanley:
We don't want these errors polluting object repository results.  I've seen:

      File ..., line ... in ...
        if stbt.match(image, frame= self._frame):
    lmdb.Error: mdb_txn_commit: Input/output error

In the object repo.

### DIFF
--- a/_stbt/imgproc_cache.py
+++ b/_stbt/imgproc_cache.py
@@ -14,6 +14,7 @@ import inspect
 import itertools
 import json
 import os
+import warnings
 import sys
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


drothlis (approved):
You could use [warnings.warn] which will only print the warning once.

[warnings.warn]: https://docs.python.org/3/library/warnings.html

wmanley:
> You could use [warnings.warn](https://docs.python.org/3/library/warnings.html) which will only print the warning once.

Done, that's much nicer.

PR #870, opened 2026-05-07
Lewis/remove virtual stb entirely

LewisHaley:
some more clean-up whilst working towards better packaging/python3.13 support.


drothlis:
Thanks @LewisHaley.

PR #872, opened 2026-05-14
stbt debug improvements targetted at LLMs

wmanley:
Here we provide more information in the debug HTML about the meaning and interpretation of each written image. This helps LLMs which are much better at text than images. It's also incidentally better for humans too.

In particular there are significant improvements to OCR debug (the cause of most customer support requests) by leveraging tesesract's TSV output to get regions and confidence percentages.

We also include a traceback of where the stbt functions are called from, which can be helpful navigating abstractions in unfamiliar test-packs.


_stbt/logging.py
```diff
@@ -196,15 +212,27 @@ def imwrite(
             return
         if name in self.images:
             raise ValueError("Image for name '%s' already logged" % name)
-        if image is None:
-            return
```
drothlis, line 200: Are you sure this isn't necessary? It was introduced in 4b8c73de210398a2c156622bd12a82dd12fb3b94 and it looks like mask can be None here:

https://github.com/stb-tester/stb-tester/blob/cc0d9ae9127f305515dd584dcd26df97bf678600/_stbt/match.py#L837-L843

_stbt/logging.py
```diff
@@ -311,21 +311,26 @@ def _draw(region, source_size, css_class, title=None):
...
+            <div class="region {{css_class}}{{ ' has-tooltip' if title_html else '' }}"
                  style="left: {{region.x / image.width * 100}}%;
                         top: {{region.y / image.height * 100}}%;
                         width: {{region.width / image.width * 100}}%;
                         height: {{region.height / image.height * 100}}%"
-                 {% if title %}
-                 title="{{ title | escape }}"
-                 {% endif %}
-                 ></div>
+                 >{% if title_html %}<div class="tooltip">{{ title_html }}</div>{% endif %}</div>
```
drothlis, line 326: We're still using the title attribute on images (``) in is_screen_black, match, and motion. Not a blocker.

_stbt/ocr.py
```diff
@@ -890,21 +897,76 @@ def _tesseract_subprocess(
 
             imglog.set(tsv_regions=tsv_regions)
 
+            tsv_words = [x for x in tsv_regions if x[1] == "tsv-word"]
+            if not tsv_words:
+                info_msg = "Note: No words were detected by Tesseract."
+            else:
+                info_msg = (
+                    "This image should have at least a 1px white border "
+                    "around every edge."
```
drothlis, line 906: I see. This is stating expectations of what the image should look like, but it isn't stating whether or not the image actually meets those expectations.

drothlis:
The OCR annotations look really useful.

_stbt/ocr.py
wmanley: Yeah, I think it helps to put the region annotations in context.

_stbt/logging.py
wmanley: You mean because we're not using `annotated_image` there? Yeah.

_stbt/logging.py
wmanley: It's not necessary because there's already a `if image is None: return` in this function.

_stbt/logging.py
wmanley: Also note: This is the title of the region, not the image.  The image still has a title.

wmanley:
I've lost `class=thumb` here for some images...

PR #874, opened 2026-05-19
Lewis/ci config in pyproject

LewisHaley:
For reference:
https://peps.python.org/pep-0621/
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/

For explicitness, this is what the Pylint output looks like now without the output-captured-into-grep in the pylint.sh script:
image


drothlis (approved):

drothlis:
Thanks @LewisHaley!

PR #875, opened 2026-05-19
stbt debug fixes

wmanley:
Follow up fixes from #872

### DIFF
--- a/_stbt/logging.py
+++ b/_stbt/logging.py
@@ -335,7 +335,13 @@ class ImageLogger():
                     title_html=title_html,
                     )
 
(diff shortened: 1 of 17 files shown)


drothlis (approved):
> tesseract does no quoting of TSV.

I guess it never generates a tab as part of the actual OCRed text?

wmanley:
> I guess it never generates a tab as part of the actual OCRed text?

That's my assumption. The text is only available on "word" rows, and as whitespace I'd expect tabs not to appear. After working on this I've been wondering whether hocr would have been a better idea, but that can change later if we bump into limitations with this approach.

PR #876, opened 2026-05-19
android: Don't run adb server in current directory

drothlis:
On the Stb-tester Node each test is run in a separate directory. We don't want the adb server to hold the current test-run directory open, so that it won't be killed by the result upload process.

### DIFF
--- a/_stbt/android.py
+++ b/_stbt/android.py
@@ -180,14 +180,17 @@ class AdbDevice():
             "android", "coordinate_system", default=CoordinateSystem.HDMI_720P,
             type_=CoordinateSystem)
 
-        if not (self.adb_server or
-                os.environ.get("ANDROID_ADB_SERVER_ADDRESS") or
-                os.environ.get("ADB_SERVER_SOCKET")):
+        if not self._is_remote_server():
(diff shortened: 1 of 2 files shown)


drothlis:
`AdbDevice` is often constructed at module import time, e.g. to use `AdbDevice().logcat` as a function decorator.

Starting adb server takes 3 seconds (measured on my laptop). In the no-op case (when adb server is already running) `adb start-server` takes 5ms.

If `lazy_connect` is False (the default) and `tcpip` is True (commonly true), we pay the startup time anyway, because we call `_connect` from the constructor.

Otherwise this change will slow down import time by a noticeable amount.

*Edit: Addressed by commit 095ba940.*

drothlis:
Previously:

- #857
- #859 (reverted the above's change and made it configurable instead)

wmanley (approved):
adb server will be started when `AdbDevice` is constructed when in TCP mode, but not in either USB mode or "no adb device" mode.

It would be nice to be able to call start-server in the constructor if the user is going to actually use it, but we can't know that in advance, and many customers may construct an AdbDevice at module import time.

Well thought through.

PR #877, opened 2026-06-16
Lewis/better python packaging

LewisHaley:
As previously discussed, this is done in aid of adding support for Python3.13, as the current supported version (Python3.10) goes EOL in October this year.

I'm fully understanding that this is a large changeset, but I think it's all justified and there are minimal changes to the unit-tests which hopefully proves that what work previously still works now.

Some notes:
* 2c77fe3cf18121f3eb3bd95c706d61bdb0a7fac9 - as per the CM this is done to enable me to test changes to the CI Dockerfile. I'm not expecting this to be a permanent change (although I do find it a curious approach to have static Docker image for CI, as it means that making changes to it is... difficult?)
* I've tried my best to group dependencies intelligently and give the "extras" sensible names - I defer to you to change anything as you wish.
* I'm pretty confident that the changes around `libstbt` and `xxhash` are covered by the call to `install-for-test` as part of the packaging CI, and I think that none of my changes negatively affect the building of your Debian package. However I'm not familiar enough with this to say for sure.


LewisHaley:
I've trying to dig into the Debian package tooling. I can't get it to work even on `main`:
```
$ make -B deb
rsvg-convert tests/buttons.svg >tests/buttons-on-blue-background.png
rsvg-convert <(sed 's/#0000ff/#ffffff/' tests/buttons.svg) >tests/buttons.png
make: *** No rule to make target 'tests/images/appletv/BBC', needed by 'stb-tester-34-283-g78b67901-dirty.tar.gz'. Stop.
```
I can't even determine how it's got a target of `tests/images/appletv/BBC`.

LewisHaley:
Hi folks, I appreciate it's summer and things might be busy, but I'm looking to get even a cursory idea of whether this PR might be accepted. I can't really move forward with Python3.13 support until we have better structure to the Python dependencies, as provided by this branch.

Makefile
```diff
@@ -309,7 +308,7 @@ publish-ci-docker-images: $(CI_DOCKER_IMAGES:%=.github/workflows/.%.built)
 
 pypi-publish:
 	rm -rf dist/
-	python3 setup.py sdist
+	python3 -m build --sdist
```
drothlis, line 309: Where does this package come from? I get:

```
❯ python3 -m build --sdist
/usr/bin/python3: No module named build.__main__; 'build' is a package and cannot be directly executed
```

Makefile
```diff
@@ -266,8 +265,8 @@ sq = $(subst ','\'',$(1)) # function to escape single quotes (')
 TAGS:
 	etags stbt_core/**.py _stbt/**.py
 
-_stbt/libstbt.$(platform).so : _stbt/sqdiff.c
-	$(CC) -shared -fPIC -O3 -o $@ _stbt/sqdiff.c $(CFLAGS)
+_stbt/_libstbt$(shell python3-config --extension-suffix) : _stbt/sqdiff.c
+	$(CC) $(shell python3-config --includes) -shared -fPIC -O3 -o $@ _stbt/sqdiff.c $(CFLAGS)
```
drothlis, line 268: Before: libstbt.x86_64.so
After: libstbt.cpython-310-x86_64-linux-gnu.so

i.e. it's built against a specific Python version.

The difference in the compiler command line is `-I/usr/include/python3.10`

Not too bad, shouldn't be much work to update our build system for the Stb-tester Node.

pyproject.toml
```diff
@@ -17,7 +17,7 @@ requires-python = ">=3.10"
 dependencies = [
     "astroid==2.11.6",
     "attrs==21.2.0",
-    "opencv-python~=4.5",
+    "opencv-python-headless~=4.5",
```
drothlis: Will need to bump the package version (to `34.13.0` I guess).

*Edit: This is a note to myself, not asking you to do it.*

.github/workflows/ubuntu2204.dockerfile
```diff
@@ -45,6 +44,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
         python3-requests \
         python3-responses \
         python3-serial \
+        python3-xxhash \
```
drothlis: `python3-xxhash` isn't available in Ubuntu 22.04, which we still support (but not for much longer?).

24.04 and 26.04 both have python3-xxhash 3.2.0.

https://github.com/ifduyue/python-xxhash tag v3.2.0 has submodule `deps/xxhash` (https://github.com/Cyan4973/xxHash) at 35b0373c697b5f160d3db26b1cbb45a0d5ba788c (v0.8.1, 2021).
The `stb-tester/vendor/xxhash` submodule is at fcabf31 (v0.5.1, 2016).

extra/debian/control
```diff
@@ -54,7 +53,6 @@ Depends: ${shlibs:Depends},
          gstreamer1.0-plugins-good,
          gstreamer1.0-tools,
          gstreamer1.0-x,
-         libxxhash0,
```
drothlis: Need to add python3-xxhash

pyproject.toml
```diff
@@ -15,24 +15,60 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
-    "astroid==2.11.6",
-    "attrs==21.2.0",
-    "opencv-python-headless~=4.5",
-    "pylint==2.14.3",
```
drothlis: We are unlikely to merge this commit (`pyproject: Declare all dependencies (runtime and dev)`).

It's good work, and makes sense if we were using this pip package to install Stb-tester in production — but we aren't. This package's purpose is so that our customers can install `stbt_core` for lint & intellisense in their IDEs. So all the dependencies need to be exactly pinned to what is installed in the Stb-tester Node's environment. Particularly pylint & astroid as we want the pylint results in the user's IDE to match the [stbt-lint][1] GitHub checks.

[1]: https://stb-tester.com/manual/lint-checks#lint-job

And the `python-package` CI job needs to be (fairly) representative of such an environment, which definitely won't have anything GStreamer/GI installed.

drothlis (commented):
Hi Lewis, sorry for the delay and thanks for the reminder.

It all looks reasonable, but I don't think we'll be able to merge e9d353890 (see comment below). Would that be a blocker for your subsequent pull requests? Are you using the python packaging scripts for your deployments, or are your changes here only for CI? 

Are you using the debian packaging in `extra/debian/`? We aren't, so we're happy to drop them if that makes things easier for you.

.github/workflows/ubuntu2204.dockerfile
drothlis: It's unclear to me from the [xxhash changelog](https://github.com/Cyan4973/xxHash/blob/dev/CHANGELOG) whether there are any compatibility breaks between these 2 versions, i.e. will the hash change for the same input?

In any case there are enough dependency changes in this pull request that we'd want to release a new major version of stbt_core — and now's not a bad time to do it, with Ubuntu 26.04 LTS just released.

drothlis:
> although I do find it a curious approach to have static Docker image for CI, as it means that making changes to it is... difficult

I think when I implemented this GitHub wasn't caching the result so it was faster than building the image each time. It was a long time ago though.

Makefile
LewisHaley: It's the exemplar, no-frills python front-end implementation, as documented here: https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives. The error message you're seeing makes me wonder whether you have a local, unrelated `build` directory, e.g.:
```
$ mkdir build
$ touch build/__init__.py
$ python3.14 -m build
/usr/bin/python3.14: No module named build.__main__; 'build' is a package and cannot be directly executed
```

LewisHaley:
> Are you using the debian packaging in `extra/debian/`? We aren't, so we're happy to drop them if that makes things easier for you.

No, we have up until now been modifying `setup.py` on a fork to include a more comprehensive dependency list and creating a python wheel from it. If you're happy to do away with it, it's one less thing to have to worry about.

pyproject.toml
LewisHaley: I think this is workable. I think we can revert the core dependencies list to what was in setup.py thus not dragging in any of the "difficult" packages by default. We can then have a dependency group for the packages required to interface with the Stb-tester Node and a further one which adds the "difficult" packages that makes the application work at runtime.
As such, your users can install something like `pip install stb-tester[ide]` and we can do `pip install stb-tester[runtime]`.

Is the `stbt-lint` you're referencing the same/an extension of the `stbt_lint.py`/`stbt lint` CLI tool? As in, that can be used as a litmus test for whether the correct packages at the correct pins are installed?

LewisHaley:
I've raised https://github.com/stb-tester/stb-tester/pull/878 to remove the Debian packaging: https://github.com/stb-tester/stb-tester/pull/878

pyproject.toml
LewisHaley: Please see https://github.com/stb-tester/stb-tester/pull/877/commits/9f801f8f20cd739e1d0384c3d7bc92f99be6a830. I think it meets your requirements.

LewisHaley:
Do you have any thoughts on my commit in response to https://github.com/stb-tester/stb-tester/pull/877#discussion_r3630642723 
https://github.com/stb-tester/stb-tester/pull/877/commits/9f801f8f20cd739e1d0384c3d7bc92f99be6a830
Please and thanks.

pyproject.toml
```diff
@@ -1,6 +1,118 @@
...
+    "markupsafe"
+]
+keyboard = [
+    "networkx",
+]
+roku = [
+    "requests",
+]
+stb_tester_node = [
+    # These are precise pins needed to be 100% compatible with the stb-tester.com Node.
```
drothlis:
```suggestion
    # Customers of stb-tester install this variant for lint, type-checking, and
    # autocompletion support in their IDE. These are precise pins to match the
    # run-time environment on the Stb-tester Node.
```

pyproject.toml
```diff
@@ -1,6 +1,118 @@
...
+    "astroid==2.11.6",
+    "attrs==21.2.0",
+    "opencv-python-headless~=4.5",
+    "pylint==2.14.3",
+    "xxhash",
```
drothlis: Before it was:

```python
    extras_require={
        "ocr": ["lxml==4.8.0"],
        "debug": ["Jinja2==3.0.3"],
        "keyboard": ["networkx==2.4"],
    },
    install_requires=[
        "astroid==2.11.6",
        "attrs==21.2.0",
        "opencv-python~=4.5",
        "pylint==2.14.3",
    ],
```

drothlis (changes requested):
Hi Lewis

Thanks for the reminder, I recently got back from holiday.

First, about the naming: I'd rather the variant was called "ide" instead of "stb-tester-node". The purpose of this variant is to allow our customers to install lint, type-checking and autocompletion support in their IDE.

Would it be possible to avoid this "ide" variant, and have those dependencies as mandatory dependencies, and then have your variant pull in its own additional dependencies? I assume this is not possible, because you don't want those exact versions, or they clash somehow.

The constraint for us is backward compatibility for our customers: They shouldn't have to change the command they use to install or upgrade stb-tester. Currently this is: `pip install stb-tester`. That package in turn has a dependency on stbt_core. So I think your change is OK: We can easily change the definition of stb-tester to depend on `stbt_core[ide]` instead.

_stbt/imgproc_cache.py
```diff
@@ -20,10 +20,10 @@
 from itertools import zip_longest
 
 import numpy
+import xxhash
```
drothlis: This should continue to be an optional dependency (i.e. inside the try block below). The rest of the file already copes if this dependency isn't available. This is for easier deployment to platforms where the PyPI package doesn't provide a pre-built library.

pyproject.toml
drothlis: ...and remove `xxhash` from the dependencies here.

pyproject.toml
```diff
@@ -1,6 +1,118 @@
...
+
+[tool.setuptools.package-data]
+_stbt = [
+    "stbt.conf",
+]
+
+[tool.setuptools]
+ext-modules = [
+    { name = "_stbt._libstbt", sources = ["_stbt/sqdiff.c"] },
+]
```
drothlis, line 93: Is it possible to only build the shared library for your `runtime` variant? This would keep the `ide` variant a pure-python package (which is the current behaviour).

LewisHaley:
> Would it be possible to avoid this "ide" variant, and have those dependencies as mandatory dependencies, and then have your variant pull in its own additional dependencies? I assume this is not possible, because you don't want those exact versions, or they clash somehow.
> 
> The constraint for us is backward compatibility for our customers: They shouldn't have to change the command they use to install or upgrade stb-tester. Currently this is: `pip install stb-tester`. That package in turn has a dependency on stbt_core. So I think your change is OK: We can easily change the definition of stb-tester to depend on `stbt_core[ide]` instead.

It would result in clashes in various places. Ultimately, once a package's metadata says "I have a primary dependency on foobar==1.2.3", nothing can be further added to circumvent or relax that constraint. Equally, if a package were then to declare an optional dependency on "foobar==7.8.9", then the dependencies can not be resolved. This is the issue we have with the existing setup.py, because it (among other things) pins pylint to a specific patch version. Therefore our downstream test script repo, which relies on stb-tester, cannot deviate from that pin. If we tried to pin pylint to something else, the Python Package resolver would say "this is impossible". This is why it is typically not advised to have exact pins in your project's dependencies - because it is inflexible for downstream users.

Another option we could have, which would also solve the [pure-python](https://github.com/stb-tester/stb-tester/pull/877#discussion_r3812528823) part, would be to have 2 pyproject.toml files, with one being setup exactly for the `ide` and the other being a more typical one with unpinned dependencies. You only need to publish the sdist for the `ide` version as you currently do - I can continue to build my own wheel. I've show what this looks like here https://github.com/stb-tester/stb-tester/pull/877/commits/9b72d4fd3fe9de3a7e3002d6c5f59210f59fa662. It's a bit messy. You can't have multiple pyproject.toml files in the same directory, and pyproject.toml must be in the root of the project, so I've symlinked the relevant bits into the subdir.

.github/workflows/ubuntu2204.dockerfile
```diff
@@ -20,31 +20,15 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
...
-        python3-numpy \
-        python3-opencv \
+        python3-dev \
         python3-pip \
-        python3-pysnmp4 \
-        python3-pytest \
-        python3-requests \
-        python3-responses \
-        python3-serial \
-        python3-yaml \
```
drothlis, line 47: With the pain that is Python packaging, and maintainers' cavalier attitude to backward compatibility, I think that moving away from stable Ubuntu packages is a mistake. We're basically saying these are the versions that we support, and no more. We'll be moving to 26.04 soon, and maybe that'll fix the bulk of your problems?

drothlis:
I had this comment written 2 weeks ago, it looks like I hadn't submitted it:

Thanks for the example. I feel uncomfortable having the root pyproject.toml not match what we publish to PyPI. And sadly we don't have the bandwidth to support a "full" open-source packaging.

Could you symlink `_stbt` and `stbt_core` into your downstream project and package it all there? That way you're not depending on us.

I'm happy to merge the other commits, just not the packaging changes. CI doesn't rely on the pip packaging, apart from the `python-package` job.

.github/workflows/ubuntu2204.dockerfile
LewisHaley: I'm not expecting to change your mind with this reply, but for what it's worth, I don't share your opinion that Python packaging is a pain. I think there have been large steps forward in the ecosystem in the last 5-10 years and that most major packages follow the guidelines quite well. Not all of them, but generally.

> We're basically saying these are the versions that we support, and no more.

This could be achieved with package constraints in the project metadata, and it would be clearer because the actual version numbers would be there in version control.

As I said, I'm not trying to convince you, and I appreciate the nature of this project, who is using it and how, and the time available to work on it. I just wanted to share my 2 cents on the matter.

LewisHaley:
We'll figure out how to handle this our side. I'll try to reduce this down to not include https://github.com/stb-tester/stb-tester/pull/877/commits/5f7006813e287deffb311049fd5f72d543c15c4a.

.github/workflows/ubuntu2204.dockerfile
LewisHaley: What do you want me to do here, given we're dropping the full dependency list in pyproject? As you highlighted, the `python3-xxhash` package is not present in the version of ubuntu you're targeting.
Doing the ubuntu bump seems out of scope, but the only other option is to install `xxhash` as a pip package which you are against.

Edit: I'm just going to drop the commit.

.github/workflows/ubuntu2204.dockerfile
drothlis: You're probably right. What I meant by pain is that you can't trust that a package on PyPI won't break backward compatibility even in a minor update, whereas for a given Ubuntu version the python packages in the apt repos will only get security updates and no breaking changes.

.github/workflows/ubuntu2204.dockerfile
drothlis: > Edit: I'm just going to drop the commit.

Yep, and I'll pick it up again once we update stbt-core to target Ubuntu 26.04 (hopefully soon after we return from IBC).

